PR #5, opened 2020-06-16
Random helpers

kbond:
Enables the following helpers:

```php
repository(Post::class)->random(); // random Post
repository(Post::class)->randomSet(3); // 3 random Post's
repository(Post::class)->randomSet(0, 5); // 0-5 random Post's

PostFactory::random(); // random Post
PostFactory::randomSet(3); // 3 random Post's
PostFactory::randomSet(0, 5); // 0-5 random Post's
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -563,6 +563,9 @@ $repository->assertNotExists(['title' => 'My Title']);
 $repository->getCount(); // number of rows in the database table
 $repository->first(); // get the first object (wrapped in a object proxy)
 $repository->truncate(); // delete all rows in the database table
+$repository->random(); // get a random object
+$repository->randomSet(5); // get 5 random objects
+$repository->randomSet(0, 5); // get 0-5 random objects
 
 // instance of ObjectRepository (all returned objects are proxied)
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -591,6 +594,8 @@ use Zenstruck\Foundry\Proxy;
 
 /**
  * @method static Post|Proxy findOrCreate(array $attributes)
+ * @method static Post|Proxy random()
+ * @method static Post|Proxy randomSet(int $min, ?int $max = null)
```
weaverryan, line 598: documenting this is tricky because `randomSet(5)` means "give me 5", so it's not a `$min` in that situation. I'm not sure exactly what to call it though - that method really wears two hats. I'm also not sure if we split it into 2 separate methods, what each would be called.

Also, this would return `Post[]|Proxy[]`

kbond: > I'm also not sure if we split it into 2 separate methods, what each would be called.

Yeah... `randomSet()` and `randomRange()`?

> Also, this would return Post[]|Proxy[]

Ahh, good catch!

PR #6, opened 2020-06-16
Add support for OneToMany/ManyToOne/ManyToMany relationships

kbond:
```php
use function Zenstruck\Foundry\create;
use function Zenstruck\Foundry\factory;

// ManyToOne
create(Post::class, [
    'category' => $category, // $category is instance of Category
]);
create(Post::class, [
    // Proxy objects are converted to object before calling Post::setCategory()
    'category' => create(Category::class, ['name' => 'My Category']),
]);
create(Post::class, [
    // Factory objects are persisted before calling Post::setCategory()
    'category' => factory(Category::class, ['name' => 'My Category']),
]);

// OneToMany
create(Category::class, [
    'posts' => [
        $post, // $post is instance of Post, Category::addPost($post) will be called during instantiation

        // Proxy objects are converted to object before calling Category::addPost()
        create(Post::class, ['title' => 'Post B', 'body' => 'body']),

        // Factory objects are persisted before calling Category::addPost()
        factory(Post::class, ['title' => 'Post A', 'body' => 'body']),
    ],
]);

// ManyToMany
create(Post::class, [
    'tags' => [
        $tag, // $tag is instance of Tag, Post::addTag($tag) will be called during instantiation

        // Proxy objects are converted to object before calling Post::addTag()
        create(Tag::class, ['name' => 'My Tag']),

        // Factory objects are persisted before calling Post::addTag()
        factory(Tag::class, ['name' => 'My Tag']),
    ],
]);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,11 +62,12 @@ public function test_can_post_a_comment(): void
     1. [Instantiate](#instantiate)
     2. [Persist](#persist)
     3. [Attributes](#attributes)
-    4. [Events](#events)
-    5. [Instantiator](#instantiator)
-    6. [Immutable](#immutable)
-    7. [Object Proxy](#object-proxy)
-    8. [Repository Proxy](#repository-proxy)
+    4. [Doctrine Relationships](#doctrine-relationships)
+    5. [Events](#events)
+    6. [Instantiator](#instantiator)
+    7. [Immutable](#immutable)
+    8. [Object Proxy](#object-proxy)
+    9. [Repository Proxy](#repository-proxy)
 6. [Model Factories](#model-factories)
     1. [Generate](#generate)
     2. [Usage](#usage)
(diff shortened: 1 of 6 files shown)


weaverryan:
Would the model factory usage look like (this includes #5) this?

```php
// previously use CategoryFactory to create some categories

PostFactory::create([
    'categories' => CategoryFactory::randomSet(3)
]);
```

And if so, this wouldn't work *before* this PR because it would have tried setting `Proxy` objects? And this PR unwraps them?

kbond:
> Would the model factory usage look like (this includes #5) this?

Correct, and this:

```php
PostFactory::createMany(4, [
    'categories' => CategoryFactory::randomSet(3)
]);
```

Would create 4 Posts with the same 3 categories. You would have to wrap the attributes in a closure to get a different set of categories for each of the 4 posts:

```php
PostFactory::createMany(4, function () {
    return [
        'categories' => CategoryFactory::randomSet(3)
    ];
});
```

>  this wouldn't work before this PR because it would have tried setting Proxy objects?

Correct, only `ManyToOne` worked before this PR.

PR #8, opened 2020-06-19
[POC] Refactor static state

kbond:
Attempt to make foundry configurable with the bundle. Still allows foundry to be used without the bundle.

TODO:

- [x] Add bundle services (`Manager`, `StoryManager`, `Faker`, `Instantiator`)
- [x] Bundle config:
    ```yaml
    zenstruck_foundry:
        faker:
            service: ~ # for complete control
            locale: ~ # to configure a different locale (only applicable if `service` above is not configured)
        instantiator:
            service: ~ # for complete control
            without_constructor: false # only applicable if `service` above is not configured
            allow_extra_attributes: false # only applicable if `service` above is not configured
            always_force_properties: false # only applicable if `service` above is not configured
    ```
- [x] Story service auto-configuration
- [x] Boot factory in `ZenstruckFoundryBundle::boot()`? This will allow using factories/stories in your doctrine fixtures with *zero* config.
- [x] allow for configuring if not using bundle (in `tests/bootstrap.php`). I'm thinking of a `Configuration` static class with `setFaker()`/`setInstantiator()` methods. When booting, if this has been configured, it takes precedence over bundle config. (ensure works with `GlobalState`)
- [x] Bundle service/config tests
- [x] Documentation


src/Bundle/Resources/config/services.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services
+        https://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="Zenstruck\Foundry\Manager" public="true">
```
weaverryan, line 8: Maybe FoundryInitializer ? 

Trying to avoid **Manager, especially a class that’s literally called Manager

src/Factory.php
```diff
@@ -137,21 +135,28 @@ public function withoutPersisting(): self
...
 
-    final public static function registerFaker(Faker\Generator $faker): void
+    /**
+     * @internal
+     */
+    final public static function manager(): Manager
     {
-        self::$faker = $faker;
+        if (!self::$manager) {
+            throw new \RuntimeException('Factory not yet booted.'); // todo
```
weaverryan, line 151: Good opportunity here for an better error message - probably your todo relates to this

src/Test/Configuration.php
```diff
@@ -0,0 +1,50 @@
...
+        self::$faker = $faker;
+    }
+
+    public static function bootFactory(ContainerInterface $container): Manager
+    {
+        try {
+            $manager = $container->get(Manager::class);
+        } catch (NotFoundExceptionInterface $e) {
+            // bundle not enabled
+            Factory::boot($manager = new Manager($container->get('doctrine'), new StoryManager([])));
```
weaverryan, line 37: Possible wtf here if you intend to use the bundle but it’s not enabled or misconfigured. I’m not sure if that’s a real problem... as most users won’t have much config on the Manager service anyways. But, you could force non-bundle users to opt into on-bundle usage with some Configuration::withoutBundle(). But I’m not sure this is needed

src/Bundle/Resources/config/services.xml
kbond: If I rename to `FoundryInitializer` what should I rename `Factory::manager()` to `Factory::initializer()` - seems a bit weird.

What about `FoundryConfiguration` (or just `Configuration`), "you boot with some configuration and then have access to that configuration"

Alternatively `FoundryContainer` (or just `Container`).

src/Factory.php
kbond: Absolutely!

src/Test/Configuration.php
kbond: Yeah, good call - I'll re-evaluate.

kbond:
I renamed `Test/Configuration` to `Test/TestState` (better name suggestion welcome) and using without the bundle must be enabled specifically with `TestState::withoutBundle()`.

I also merged `GlobalState` into `TestState` as they both related to *test-only* static configuration.

src/Factory.php
kbond: I'm going to improve all exception messages.

kbond:
I renamed `Manager` to `Configuration`

kbond:
One thing that kind of sucks is if you want your dev and test config to match, you have to duplicate it in `config/packages/dev/zenstruck_foundry.yaml` and `config/packages/test/zenstruck_foundry.yaml`

PR #9, opened 2020-06-23
[minor] Default makers to generate in src/

kbond:
- Add `--test` option to generate in `tests/`

**Questions:** 

1. Should `--test` be interactive?
2. Should the folders be plural (current) or singular? `src/Stories` or `src/Story`, `src/Factories` or `src/Factory`. Mostly, it seems the standard is singular (with the exception of `Migrations`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -640,15 +640,15 @@ Create a model factory for one of your entities with the maker command:
 
 **NOTES**:
 
-1. Creates `PostFactory.php` in `src/Factories`, add `--test` flag to create in `tests/Factories`.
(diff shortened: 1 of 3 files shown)


weaverryan:
> Should --test be interactive?

I don't mind interactive questions. But in this case, in either direction, it could get kinda annoying once you've generated your 5th thing. When `--test` is NOT passed, we could print a message on top:

> // pass --test if you want to generate files into your tests/ directory

Longer-term, you could envision a config option to default which one you want

> Should the folders be plural (current) or singular? src/Stories or src/Story, src/Factories or src/Factory.

Probably singular??? :)

PR #19, opened 2020-07-10
Disable proxy auto-refresh by default but add ability to enable globally

### LINKED ISSUE #18: Refactor Proxy auto-refresh feature
Ref: https://github.com/zenstruck/foundry/pull/13#discussion_r451683857 and slack discussions with Ryan

1. Disable by default
2. Update example doc to use `->refresh()`
3. Global config option
4. In docs:
    1. First show how `->refresh()` works (no mention of auto-refreshing)
    2. Then show how to enable/disable auto-refreshing on your Proxy
    3. Then show global config option to auto-refresh by default
5. `Factory::withAutoRefresh()/withoutAutoRefresh()` (?)

kbond:
closes #18

### DIFF
--- a/README.md
+++ b/README.md
@@ -858,7 +858,7 @@ $post->getTitle(); // "New Title" (equivalent to $post->refresh()->getTitle())
 
 Without auto-refreshing enabled, the above call to `$post->getTitle()` would return "Original Title".
 
-**NOTE**: A situation you need to be aware of when using auto-refresh is all methods refresh the object first. If
+**NOTE**: A situation you need to be aware of when using auto-refresh is that all methods refresh the object first. If
 changing the object's state via multiple methods (or multiple force-sets), the previous changes will be lost:
 
 ```php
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -790,77 +792,138 @@ FOUNDRY_RESET_CONNECTIONS=connection1,connection2
...
+```php
+use App\Factory\PostFactory;
+
+$post = PostFactory::new(['title' => 'Original Title', 'body' => 'Original Body'])
+    ->create()
+    ->enableAutoRefresh()
+;
+
+$post->setTitle('New Title'); // or using ->forceSet('title', 'New Title')
+$post->setBody('New Body'); // or using ->forceSet('body', 'New Body')
```
weaverryan, line 873: Instead of mentioning `forceSet` here (which you already mention above), I would remove the comment from the first line. And on the second line, say:

> // this causes the object to be refresh, which means the "New Title" title was replaced by the database contents

PR #38, opened 2020-09-27
Add FactoryCollection

kbond:
Adds a `Factory::many()` method that returns a `FactoryCollection` object to help with *lazily* creating factories with *-many relationships. This really helps the DX of one-to-many relationships.

#### Examples:

```php
// create 10 Post's
PostFactory::new()->createMany(10, [
    'comments' => CommentFactory::new()->many(3), // each post will have 3 Comment's
]);

// create 10 Post's
PostFactory::new()->createMany(10, [
    'comments' => CommentFactory::new()->many(0, 5), // each post will have between 0 and 5 Comment's
]);
```

#### TODO:

- [x] update docs

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -92,18 +92,21 @@ class Factory
     }
 
     /**
+     * @see FactoryCollection::__construct()
+     */
+    final public function many(int $min, ?int $max = null): FactoryCollection
+    {
+        return new FactoryCollection($this, $min, $max);
+    }
+
+    /**
      * @param array|callable $attributes
      *
      * @return Proxy[]|object[]
      */
     final public function createMany(int $number, $attributes = []): array
     {
-        return \array_map(
-            function() use ($attributes) {
-                return $this->create($attributes);
-            },
-            \array_fill(0, $number, null)
-        );
+        return $this->many($number)->create($attributes);
     }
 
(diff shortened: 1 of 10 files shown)


src/Factory.php
```diff
@@ -229,6 +229,14 @@ private function normalizeAttribute($value)
             return $value->object();
         }
 
+        if ($value instanceof FactoryCollection) {
+            $this->afterPersist[] = function(Proxy $proxy) use ($value) {
```
kbond, line 233: Simplify to `[$value, 'createFor']` for callback?

Got the following error: 

`TypeError: ReflectionFunction::__construct() expects parameter 1 to be string, array given`

From `Proxy::executeCallback()`

src/FactoryCollection.php
```diff
@@ -0,0 +1,46 @@
...
+     * @return Proxy[]|object[]
+     */
+    public function create($attributes = []): array
+    {
+        return $this->factory->createMany($this->count, $attributes);
+    }
+
+    public function createFor(Proxy $proxy): void
+    {
+        $attribute = \lcfirst((new \ReflectionClass($proxy->object()))->getShortName());
```
kbond, line 38: This should probably be configurable...

PR #49, opened 2020-10-08
SQLite test suite fails with FoundryBundle & DAMABundle enabled

kbond:
The tests run successfully when run separate. Seems to be caused by a test that does not have the `ResetDatabase` trait and boots the kernel (in this case `MakeFactoryTest`). The next test that runs with the `ResetDatabase` trait fails...


nikophil:
if this issues gets resolved some day, please don't forget to update make target `phpunit-fast` with `USE_FOUNDRY_BUNDLE=1`

see https://github.com/zenstruck/foundry/pull/306

nikophil:

PR #53, opened 2020-10-09
Allow factories to be defined as services

### LINKED ISSUE #32: DI support for ModelFactory class
There is no Dependency Injection support when initializing a new ModelFactory with the new() method. I came upon this problem very early on when creating a User factory because I was unable to inject the PasswordEncoderInterface service into the factory. This meant I couldn't contain all logic involved in creating a new User inside the factory.

Any plans to address this?

### LINKED ISSUE #36: Document using services in ModelFactory's
- ~not possible but you can use stories/tests/orm fixtures to access the services~
- `UserFactory` with encoded password is a common use-case (pre-encode known password)

(ref #32)

kbond:
This feature enables your ModelFactories to be defined as and use services from the DI container. A very common use-case would be to inject the password encoder into your `UserFactory`:

```php
final class UserFactory extends ModelFactory
{
    private $passwordEncoder;

    public function __construct(UserPasswordEncoderInterface $passwordEncoder)
    {
        parent::__construct();

        $this->passwordEncoder = $passwordEncoder;
    }

    {
        return [
            'email' => self::faker()->unique()->safeEmail,
            'password' => '1234',
        ];
    }

    {
        return $this
            ->afterInstantiate(function(User $user) {
                $user->setPassword($this->passwordEncoder->encodePassword($user, $user->getPassword()));
            })
        ;
    }

    {
        return User::class;
    }
}
```

This factory can be used as normal and the password will be encoded:

```php
UserFactory::new()->create(['password' => 'mypass'])->getPassword(); // "mypass" encoded
UserFactory::new()->create()->getPassword(); // "1234" encoded (because "1234" is set as the default password)
```

All factories created in a standard flex app (in `src/Factory` via the `make:factory` command) that have a public constructor are registered as services. Factories without a public constructor are not services.

**Some outstanding questions:**
1. ~I made `ModelFactory::__construct()` protected so concrete ModelFactory's can extend and make public to inject services. The reason I didn't make this method public is an attempt to restrict users from creating non-service ModelFactory's with `new PostFactory()` which causes problems. I was unable to find a way to throw a helpful error if ModelFactory's are created this way and not the proper way with `PostFactory::new()`.~ *I just made the constructor public with a warning in the docs to not instantiate factories with it.*
2. ~For the `make:factory` command: I think the vast majority of factory's will not require services. I'm thinking a `--service` flag to the command, that when added, stubs out a public constructor.~ *Opted not to do this and simply added a constructor stub.*

Fixes #32, Fixes #36

**TODO**
- [x] Add docs
- [x] Update `make:factory` command

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,9 @@ to load fixtures or inside your tests, [where it has even more features](#using-
     8. [Instantiation](#instantiation)
     9. [Immutable](#immutable)
     10. [Doctrine Relationships](#doctrine-relationships)
-    11. [Anonymous Factories](#anonymous-factories)
-    12. [Without Persisting](#without-persisting)
+    11. [Factories as Services](#factories-as-services)
+    12. [Anonymous Factories](#anonymous-factories)
+    13. [Without Persisting](#without-persisting)
 4. [Using with DoctrineFixturesBundle](#using-with-doctrinefixturesbundle)
 5. [Using in your Tests](#using-in-your-tests)
     1. [Enable Foundry in your TestCase](#enable-foundry-in-your-testcase)
@@ -199,6 +200,13 @@ use Zenstruck\Foundry\Proxy;
  */
 final class PostFactory extends ModelFactory
 {
+    public function __construct()
+    {
+        parent::__construct();
+
+        // TODO inject services if required (https://github.com/zenstruck/foundry#factories-as-services)
+    }
+
     protected function getDefaults(): array
     {
         return [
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -277,6 +285,9 @@ $posts = PostFactory::randomSet(4); // array containing 4 "Post|Proxy" objects
 $posts = PostFactory::randomRange(0, 5); // array containing 0-5 "Post|Proxy" objects
 ```
 
+**WARNING**: Never instantiate your `ModelFactory` with the constructor (ie `new PostFactory()`). This will
+cause the factory not be be instantiated properly. Always instantiate with `PostFactory::new()`.
```
weaverryan, line 289:
```suggestion
cause the factory to not be instantiated properly. Always instantiate with `PostFactory::new()`.
```

PR #59, opened 2020-10-16
Testing improvements

kbond:


README.md
```diff
@@ -1201,6 +1204,34 @@ Zenstruck\Foundry\Test\TestState::addGlobalState(function () {
...
+
+public static function postDataProvider(): iterable
+{
+    yield [PostFactory::new()];
+    yield [PostFactory::new()->published()];
+}
+```
+
+**NOTE**: Due to data providers being computed early in the phpunit process, it is not possible to use
+[Factory Services](#factories-as-services) in them.
```
weaverryan, line 1233: I think if I'm reading this, I'm thinking:

> Ok. My service *is* a factory... so what should I do?

And... what *would* happen? Does it fallback to just trying to instantiate manually and that would work if you had optional args?

It might also be important to point out that you should be returning *factory* instances from your data providers (exactly like you've shown here) and not actually calling `->create()` until the real method, because `->create()` in the data provider wouldn't work because it's too early (would it error? Or just create but not persist?)

README.md
```diff
@@ -1269,20 +1300,46 @@ these tests to be unnecessarily slow. You can improve the speed by reducing the
...
+    {
+        $post = PostFactory::new()->create();
+
+        // $post is not persisted to the database
+    }
+}
+```
+
+**NOTE**: [Factories as Services](#factories-as-services) and [Stories as Services](#stories-as-services) are not
+usable in non-Kernel tests.
```
weaverryan, line 1332: Same here: if my factory IS a service, what happens and what are my options?

src/ModelFactory.php
```diff
@@ -23,7 +23,12 @@ public function __construct()
             $defaultAttributes = [];
         }
 
-        $factory = self::configuration()->factories()->create(static::class);
+        try {
+            $factory = self::isBooted() ? self::configuration()->factories()->create(static::class) : new static();
+        } catch (\ArgumentCountError $e) {
+            throw new \RuntimeException('Model Factories with dependencies (Model Factory services) cannot be created before foundry is booted.', 0, $e);
```
weaverryan, line 29: I might even link to the docs here :p.

But also, one part of the solution is to make your args optional. But again, that might be best discussed by pointing to the docs. Heck, it might even make sense to have a FAQ/Errors section where you specifically guide users for common errors.

weaverryan (commented):
Took a quick look at the docs and the high-level touch points. Looks good - just a few spots where we can guide users better :)

src/ModelFactory.php
kbond: #61

PR #60, opened 2020-10-20
Adding a link to the SymfonyCasts Foundry Videos

weaverryan:
Hi!

We have a few videos about Foundry on SymfonyCasts, which might help teach it or introduce people. Both chapters in the tutorial we link to are free :).

Cheers and thanks for Foundry!


kbond:
Thanks!

PR #63, opened 2020-10-21
[RepositoryProxy] Add criteria to findOneBy()

### LINKED ISSUE #62: FindOneBy with orderBy do not return the expected entity/proxy
When using `***Factory::repository()->findOneBy([], ['id' => 'DESC'])`, the returned Proxy/Entity doesn't match.

As an example, I've something like:

```php
$entity= EntityFactory::new()->create();

self::ensureKernelShutdown();
$client = static::createClient();
$this->logIn($client, 'ROLE_USER');
$client->request('GET', '/my-entity/'.$offer->getId().'/duplicate');

$this->assertResponseRedirects();

$duplicatedOffer = EntityFactory::repository()->findOneBy([], ['id' => 'DESC']);

$this->assertNotSame($offer->getId(), $duplicatedOffer->getId());
```

But this code fail, if I replace `EntityFactory::repository()->findOneBy([], ['id' => 'DESC'])` by `self::getEntityManager()->getRepository(CustomerOffer::class)->findOneBy([], ['id' => 'DESC'])` it works.

I can order by what I want: id, date; and set the order as ASC or DESC, I always have the bad.

mpiot:
Close #62

### DIFF
--- a/src/RepositoryProxy.php
+++ b/src/RepositoryProxy.php
@@ -219,9 +219,9 @@ final class RepositoryProxy implements ObjectRepository
     /**
      * @return Proxy|object|null
      */
-    public function findOneBy(array $criteria): ?Proxy
+    public function findOneBy(array $criteria, ?array $orderBy = null): ?Proxy
     {
-        return $this->proxyResult($this->repository->findOneBy(self::normalizeCriteria($criteria)));
+        return $this->proxyResult($this->repository->findOneBy(self::normalizeCriteria($criteria), $orderBy));
     }
 
     public function getClassName(): string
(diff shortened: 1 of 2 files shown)


kbond:
Thanks @mpiot! `Doctrine\Persistence\ObjectRepository` doesn't define this param but `Doctrine\ORM\EntityRepository` does.

I did notice your code sample in #62 could benefit from a `RepositoryProxy::last()` convenience method to get the last entity added. What do you think?

mpiot:
Yes, I think it's a good idea, better experience than findOneBy with the DESC order.
I'll do a PR for that.

PR #64, opened 2020-10-21
[RepositoryProxy] Add a last() method

mpiot:
As mentioned in PR #63 about the sample of issue #62, this PR add a `->last()` method and update the `->first()` method. I'm not sure about the update, but maybe it's better to be sure of the order.

Sometimes with postgres, it's not really ordered: https://www.postgresql.org/docs/9.1/queries-order.html

> After a query has produced an output table (after the select list has been processed) it can optionally be sorted. If sorting is not chosen, the rows will be returned **in an unspecified order**. The actual order in that case will depend on the scan and join plan types and the order on disk, but it must not be relied on. A particular output ordering can only be guaranteed if the sort step is explicitly chosen.

### DIFF
--- a/src/RepositoryProxy.php
+++ b/src/RepositoryProxy.php
@@ -99,9 +99,17 @@ final class RepositoryProxy implements ObjectRepository
     /**
      * @return Proxy|object|null
      */
-    public function first(): ?Proxy
+    public function first(string $sortedField = 'id'): ?Proxy
     {
-        return $this->findOneBy([]);
+        return $this->findBy([], [$sortedField => 'ASC'], 1)[0] ?? null;
+    }
+
+    /**
+     * @return Proxy|object|null
+     */
+    public function last(string $sortedField = 'id'): ?Proxy
+    {
+        return $this->findBy([], [$sortedField => 'DESC'], 1)[0] ?? null;
     }
 
     /**
--- a/README.md
+++ b/README.md
@@ -1135,7 +1135,10 @@ $repository = repository(Post::class);
 
 // helpful methods - all returned object(s) are proxied
 $repository->getCount(); // number of rows in the database table
-$repository->first(); // get the first object (wrapped in a object proxy)
+$repository->first(); // get the first object (assumes an auto-incremented "id" column)
+$repository->first('createdAt'); // assuming "createdAt" is a datetime column, this will return latest object
+$repository->last(); // get the last object (assumes an auto-incremented "id" column)
+$repository->last('createdAt'); // assuming "createdAt" is a datetime column, this will return oldest object
 $repository->truncate(); // delete all rows in the database table
 $repository->random(); // get a random object
 $repository->randomSet(5); // get 5 random objects
--- a/tests/Functional/RepositoryProxyTest.php
+++ b/tests/Functional/RepositoryProxyTest.php
@@ -55,10 +55,6 @@ final class RepositoryProxyTest extends KernelTestCase
 
         CategoryFactory::new()->createMany(2);
 
-        $object = $repository->first();
-
-        $this->assertInstanceOf(Proxy::class, $object);
-
         $objects = $repository->findAll();
 
         $this->assertCount(2, $objects);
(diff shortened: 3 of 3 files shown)


src/RepositoryProxy.php
```diff
@@ -101,7 +101,15 @@ public function assertNotExists($criteria, string $message = ''): self
      */
     public function first(): ?Proxy
```
kbond, line 102:
```suggestion
    public function first(string $sort = 'id'): ?Proxy
```

The user could be using a different convention? (ie maybe UUIDs for the id want to use `createdAt` for the sort)

src/RepositoryProxy.php
```diff
@@ -101,7 +101,15 @@ public function assertNotExists($criteria, string $message = ''): self
...
     {
-        return $this->findOneBy([]);
+        return $this->findOneBy([], ['id' => 'ASC']);
```
kbond, line 104:
```suggestion
        return $this->findBy([], [$sort => 'ASC'], 1)[0] ?? null;
```

The second parameter of `findOneBy` isn't necessarily available as it is not defined on the `ObjectRepository` interface.

src/RepositoryProxy.php
```diff
@@ -101,7 +101,15 @@ public function assertNotExists($criteria, string $message = ''): self
...
+    }
+
+    /**
+     * @return Proxy|object|null
+     */
+    public function last(): ?Proxy
```
kbond, line 110:
```suggestion
    public function last(string $sort = 'id'): ?Proxy
```

src/RepositoryProxy.php
```diff
@@ -101,7 +101,15 @@ public function assertNotExists($criteria, string $message = ''): self
...
+    {
+        return $this->findOneBy([], ['id' => 'DESC']);
```
kbond, line 112:
```suggestion
        $this->findBy([], [$sort => 'DESC'], 1)[0] ?? null;
```

kbond (changes requested):
I had the exact same thought about ordering. Just a few minor changes to make this methods more robust.

src/RepositoryProxy.php
mpiot: The project isn't limited to the EntityManager ? I ask it because this interface work with the Entity and Document Manager I think that's the reason about the missmatch.

But if we really want to deal with the interface (not only the EntityRepository), I think we must create an other `findOneBy` with an other name (that don't exist in the interface) to be sure to not break it.
Actually, if someone do not use the EntityManager, the `findOneBy` will be broken since the previous merge (it would try to add $orderBy in the arguments of the not EntityManager and... crash).

kbond: Yes, this lib technically only requires `doctrine/persistence` which provides `ObjectManager::findOneBy(array $criteria)` (no `$orderBy`). It "should" be compatible with any manager/repository that implements this interface (ORM/ODM/etc...).

I have yet to see any user use something other than the ORM but I'd like to try and keep compatible as much as possible.

The ORM's `EntityRepository` adds `$orderBy` as an optional parameter. The MongoDB-ODM's `DocumentRepository` does not.

You're right, the `$orderBy` will potentially not be available with every `ObjectRepository` instance but I don't think it will crash, it will just ignore the extra parameter. This isn't great either. Probably should use reflection to see if this parameter exists - I've created an issue (#65).

For this PR, I'd like to use the method I described to make compatible with all instances of `ObjectRepository`'s.

mpiot: Yep, I've done that.

https://github.com/doctrine/orm/blob/2.7/lib/Doctrine/ORM/EntityRepository.php#L189

https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Repository/DocumentRepository.php#L166
I think that it will have a problem if we pass it an extra argument.

kbond: Yeah, it won't raise an exception but it will silently ignore the extra passed parameter (https://3v4l.org/RR1YS)

kbond (approved):

kbond:
Thanks again @mpiot, I've added some tests and documented.

PR #73, opened 2020-10-29
Run postgres test suite with DAMADoctrineTestBundle enabled

kbond:
Followup for #51.

The test suite does not pass for posgres with DAMADoctrineTestBundle enabled.


nikophil:

PR #77, opened 2020-11-06
Remove attribute prefixes in findOrCreate()

### LINKED ISSUE #75: Remove attribute prefix when doing findOrCreate() ?
I'm not totally sure whether we're wrongly using this library or whether I have stumbled on a bug (I'm having this bug in so many places that I almost believe it's the first). That's why I decided to open an issue first, I'm happy to provide a PR to fix this issue.

---

In our code, some properties are not set by the constructor or by property accessible setters. E.g. a `User#subscribeTo($subscription)` method. However, I want to create fixtures with subscriptions. From what I understand, the `force:` attribute prefix is the way forward: `['force:subscription' => $subscription]`.

However, this does not work with `findOrCreate()`, as doctrine will try to use a field called `force:subscription` instead of the expected `subscription`. I'm often using `findOrCreate()` in my tests, as I want to avoid creating new test fixtures if a matching is already found.

The solution would be to strip attribute prefixes before passing them into `repository()->find($attributes)`: https://github.com/zenstruck/foundry/blob/ea8430f1ad295c88818e056ca565e5db2c528e0b/src/ModelFactory.php#L55-L59

wouterj:
Fixes #75 

Btw, do you have an easy way to run the tests locally? I couldn't really figure out how to do it, so I rely quite heavily on the CI.

### DIFF
--- a/src/Instantiator.php
+++ b/src/Instantiator.php
@@ -115,6 +115,21 @@ final class Instantiator
         return self::accessibleProperty($object, $property)->getValue($object);
     }
 
+    public static function getFindCriteriaFromAttributes(array $attributes): array
+    {
+        $criteria = [];
+        foreach ($attributes as $key => $value) {
+            if (0 === \mb_strpos($key, 'optional:')) {
+                continue;
+            }
+
+            $normalizedKey = false !== \mb_strpos($key, ':') ? \mb_substr($key, \mb_strpos($key, ':') + 1) : $key;
+            $criteria[$normalizedKey] = $value;
+        }
+
+        return $criteria;
+    }
+
     private static function propertyAccessor(): PropertyAccessor
     {
         return self::$propertyAccessor ?: self::$propertyAccessor = PropertyAccess::createPropertyAccessor();
--- a/src/ModelFactory.php
+++ b/src/ModelFactory.php
@@ -54,7 +54,7 @@ abstract class ModelFactory extends Factory
      */
     final public static function findOrCreate(array $attributes): Proxy
     {
-        if ($found = static::repository()->find($attributes)) {
+        if ($found = static::repository()->find(Instantiator::getFindCriteriaFromAttributes($attributes))) {
             return $found;
         }
 
--- a/tests/Functional/ModelFactoryTest.php
+++ b/tests/Functional/ModelFactoryTest.php
@@ -36,6 +36,17 @@ final class ModelFactoryTest extends KernelTestCase
     /**
      * @test
      */
+    public function can_find_or_create_using_prefixed_fields(): void
+    {
+        CategoryFactory::findOrCreate(['force:name' => 'php', 'optional:version' => 7.4]);
+        CategoryFactory::repository()->assertCount(1);
+        CategoryFactory::findOrCreate(['force:name' => 'php', 'optional:version' => 7.4]);
+        CategoryFactory::repository()->assertCount(1);
+    }
+
+    /**
+     * @test
+     */
     public function can_override_initialize(): void
     {
         $this->assertFalse(PostFactory::new()->create()->isPublished());


kbond:
> Btw, do you have an easy way to run the tests locally?

Yes, see #78 (`DATABASE_URL=<dsn> ./run-tests`).

src/ModelFactory.php
```diff
@@ -54,7 +54,14 @@ public function __construct()
      */
     final public static function findOrCreate(array $attributes): Proxy
     {
-        if ($found = static::repository()->find($attributes)) {
+        // remove attribute prefixes (like "force:" and "optional:")
```
kbond, line 57: I'm thinking this logic should live inside `Zenstruck\Foundry\Instantiator` as this is where the attribute prefix logic lives. Probably a static `normalizeAttributes(array $attributes): array` method. WDYT?

kbond (changes requested):

src/ModelFactory.php
```diff
@@ -54,7 +54,14 @@ public function __construct()
...
+        $normalizedAttributes = [];
+        foreach ($attributes as $key => $value) {
+            $normalizedKey = false !== \mb_strpos($key, ':') ? \mb_substr($key, \mb_strpos($key, ':') + 1) : $key;
+            $normalizedAttributes[$normalizedKey] = $value;
```
kbond, line 61: Also, `optional:` attributes be removed from the array, no?

wouterj: Oh, good catch. Included that one as well

wouterj:
> Yes, see #78 (`DATABASE_URL=<dsn> ./run-tests`).

Thanks, that works! It is very slow on my poor HDD laptop though. It takes about 30 sec to run the bootstrap on my PC. So for now, I'll stick to only run the dama doctrine bundle based tests (otherwise it's 30 sec per test). I'm not sure if that's expected, I'll see if I can find something to improve (in a few weeks).

src/Instantiator.php
```diff
@@ -115,6 +115,21 @@ public static function forceGet(object $object, string $property)
...
 
+    public static function getFindCriteriaFromAttributes(array $attributes): array
+    {
+        $criteria = [];
+        foreach ($attributes as $key => $value) {
+            if (0 === \mb_strpos($key, 'optional:')) {
+                continue;
+            }
+
+            $normalizedKey = false !== \mb_strpos($key, ':') ? \mb_substr($key, \mb_strpos($key, ':') + 1) : $key;
```
wouterj, line 126: I'm not sure whether we should simply remove all colon prefixes, or if we should explicitly only remove the `force:` prefix. I think it's safe to say that no column name will ever have a colon in it, but I'm not sure about your idea on this.

kbond:
> I'll stick to only run the dama doctrine bundle based tests

Yeah, that's what I do as well :)

src/Instantiator.php
kbond: I think it's fine as is.

tests/Functional/ModelFactoryTest.php
```diff
@@ -33,6 +33,17 @@ public function can_find_or_create(): void
         CategoryFactory::repository()->assertCount(1);
     }
 
+    /**
+     * @test
+     */
+    public function can_find_or_create_using_prefixed_fields(): void
+    {
+        CategoryFactory::findOrCreate(['force:name' => 'php', 'optional:version' => 7.4]);
```
kbond, line 41: Was just about to suggest adding an optional value here.

kbond (approved):

kbond:
Looks good, let me know when you're done and I'll merge.

wouterj:
I think this is done (I indeed forgot to add a test for the optional attribute in my previous push).

wouterj:
However, I was just thinking about something else related to these prefixes.

If in this case the `CategoryFactory` has a default attribute named `force:name`, we need to use `CategoryFactory::new()->create(['force:name' => 'php'])' in order to override the default value. Isn't this leaking implementation details about the factory to the outside world? (e.g. I wouldn't be able to refactor my factory to use a custom instantiator without updating all usages to remove the `force:` prefix)

A possible solution is to always check `force:name` when a `name` attribute is passed.
Or maybe the `force:` prefix should be deprecated in favor of something like `->forceSetAttributes(['name'])` in the `Instantiator`?

kbond:
Closing in favour of #81.

PR #79, opened 2020-11-08
Add static return annotations

micheh:
Currently the factories and the Story are using the `self` native return type to enable the fluid interface, which can confuse static analyzers and PHPStorm.

This pull request adds the `@return static` annotation for those methods, to further specify that those methods return the actual class in the inheritance hierarchy and not the abstract class itself.

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -109,6 +109,9 @@ class Factory
         return $this->many($number)->create($attributes);
     }
 
+    /**
+     * @return static
+     */
     public function withoutPersisting(): self
     {
         $cloned = clone $this;
@@ -119,6 +122,8 @@ class Factory
 
     /**
      * @param array|callable $attributes
+     *
+     * @return static
      */
     final public function withAttributes($attributes = []): self
     {
@@ -130,6 +135,8 @@ class Factory
 
     /**
      * @param callable $callback (array $attributes): array
+     *
+     * @return static
      */
     final public function beforeInstantiate(callable $callback): self
     {
@@ -141,6 +148,8 @@ class Factory
 
     /**
      * @param callable $callback (object $object, array $attributes): void
+     *
+     * @return static
      */
     final public function afterInstantiate(callable $callback): self
     {
@@ -152,6 +161,8 @@ class Factory
 
     /**
      * @param callable $callback (object|Proxy $object, array $attributes): void
+     *
+     * @return static
      */
     final public function afterPersist(callable $callback): self
     {
@@ -163,6 +174,8 @@ class Factory
 
     /**
      * @param callable $instantiator (array $attributes, string $class): object
+     *
+     * @return static
      */
     final public function instantiateWith(callable $instantiator): self
     {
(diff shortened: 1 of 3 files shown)


kbond:
Thank you @micheh

micheh:
Thank you Kevin. Would you accept a PR with Psalm specific annotations (`@psalm-...`) as well (e.g. to tell Psalm that the `Instantiator` will return an object of the provided `$class` argument)?

kbond:
@micheh, can you join me in the symfony-devs slack?

wouterj:
(please note that PHPstan also supports the psalm templated annotations, so annotating the code with either of them will help the most common static analysers to better understand what's going on)

kbond:
I've never used either, but very interested in learning more about them:

1. Why one over the other?
2. Anywhere else in this lib to utilize the annotations?
3. Should I add config and add a check to the CI process?

wouterj:
> Why one over the other?

I don't think there is much difference in functionality. It's mostly how to configure (enable/disable) things. At a basic level, Psalm uses XML while PHPstan uses Neon. As far as I know, PHPstan only has "rule levels", while Psalm has "rule levels" and the ability to explicitly disable/enable any rule you like.
Also, the communities are different (PHPstan for instance has a "pro" version).
There are also project out there that use both tools, as sometimes one catches something the other didn't.
Personally, we've choosen Psalm a couple years ago as PHPstan didn't yet have the "baseline" feature (which is very useful when introducing static analysis to a big project).

> Anywhere else in this lib to utilize the annotations?

Anything that is not-internal and is not understood when reading the PHP signatures will benefit from annotations. PHPstan/Psalm also parse the normal PHPdoc, so not much changes.

The special annotations are especially useful in places with lots of magic (e.g. the `Proxy` class and to replace the big PHPdoc that is generated for each factory).

> Should I add config and add a check to the CI process?

That's up to you :) I think every project will benefit from some static analysis (PHP is quite forgiving, it's good to be a bit stricter and even the loose "rule levels" can catch things that your tests don't (yet) catch). If you introduce the annotations, it might be a good idea to also run static analysis for this library (as that also automatically "tests" the annotations).

kbond:
Thanks for the details @wouterj! I'm thinking of using psalm and going with this [config from doctrine](https://github.com/doctrine/collections/blob/1.6.x/psalm.xml.dist). WDYT? If you're willing, I'd be open to a PR to add this. Otherwise, I'll try my hand at it in the future.

@micheh, I'd be fine with a PR to add the Psalm specific annotations.

wouterj:
Alright. I'll prepare a poc PR tomorrow.

About the config, the doctrine config is a bit "old" compared to how Psalm currently works. What I usually do is run Psalm for all error levels and then see what number of errors comes up. I'll take the one that seems the most reasonable.
E.g. in my WouterJEloquentBundle (and also this library), there is quite a significant use of magic methods and more magic PHP features. That often causes static analyzers to be a bit more upset, so enforcing a level 3 might be a bit too strict.

kbond:
Ok, cool, I look forward to the PR!

Issue #15, opened 2020-07-08
make:factory --all


kbond:
`--all` flag to `make:factory` that generates factories that don't exist for all your entities. When using interactively, when listing registered entities, "All Missing" should be the final option.

nikophil:
this is fixed in https://github.com/zenstruck/foundry/pull/247

PR #80, opened 2020-11-09
[make:factory] Sort available entities

wouterj:
I'm a bit lost using the unsorted list in an application with 40 entities (even more ideal would be to exclude the entities that already have a factory, but I couldn't easily figure out how to do that).


kbond:
Thanks @wouterj - yeah, that would be a nice feature (#82) (in addition to #15).

PR #81, opened 2020-11-09
Instantiator improvements

### LINKED ISSUE #75: Remove attribute prefix when doing findOrCreate() ?

kbond:
Fixes #75, Alternate to #77.

1. This deprecates the usage of the "force:" and "extra:" attribute prefixes. There were several problems (ref: #75 & https://github.com/zenstruck/foundry/pull/77#issuecomment-723344210).

    The new way of using is to define them on the `Instantiator` itself:

    ```php
    $factory->instantiateWith(
        (new Instantiator())->allowExtraAttributes(['extra1', 'extra2'])->alwaysForceProperties(['title', 'body'])
    );
    ```
2. This also deprecates using snake/kebab-case for attribute names for the same reasons as above.

### DIFF
--- a/src/Instantiator.php
+++ b/src/Instantiator.php
@@ -69,6 +69,7 @@ final class Instantiator
                 // see if attribute was snake/kebab cased
                 try {
                     self::propertyAccessor()->setValue($object, self::camel($attribute), $value);
+                    trigger_deprecation('zenstruck\foundry', '1.5.0', 'Using a differently cased attribute is deprecated, use the same case as the object property instead.');
                 } catch (NoSuchPropertyException $e) {
                     if (!$this->allowExtraAttributes) {
                         throw new \InvalidArgumentException(\sprintf('Cannot set attribute "%s" for object "%s" (not public and no setter).', $attribute, $class), 0, $e);
@@ -150,8 +151,10 @@ final class Instantiator
         $class = new \ReflectionClass($object);
 
         // try fetching first by exact name, if not found, try camel-case
-        if (!$property = self::reflectionProperty($class, $name)) {
-            $property = self::reflectionProperty($class, self::camel($name));
+        $property = self::reflectionProperty($class, $name);
+
+        if (!$property && $property = self::reflectionProperty($class, self::camel($name))) {
+            trigger_deprecation('zenstruck\foundry', '1.5.0', 'Using a differently cased attribute is deprecated, use the same case as the object property instead.');
         }
 
         if (!$property) {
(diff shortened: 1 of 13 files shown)


kbond:
Anyone know what's causing https://github.com/zenstruck/foundry/pull/81/checks?check_run_id=1374983511#step:9:22 ? Only with --prefer-lowest (maker-bundle 1.5.0). Ping @weaverryan

src/Instantiator.php
```diff
@@ -79,20 +92,32 @@ public function withoutConstructor(): self
...
         return $this;
     }
 
     /**
      * Always force properties, never use setters (still uses constructor unless disabled).
+     *
+     * @param string ...$properties The properties you'd like the instantiator to "force set" (if empty, force set all)
      */
-    public function alwaysForceProperties(): self
+    public function alwaysForceProperties(string ...$properties): self
```
wouterj, line 114: Variadic arguments are difficult to manage in a BC way (it's no longer possible to add any optional argument after `$properties` in the future). I would recommend to go for a simple `array $properties = []` argument.

kbond: Good call.

src/Instantiator.php
```diff
@@ -150,8 +151,11 @@ private static function accessibleProperty(object $object, string $name): \Refle
         $class = new \ReflectionClass($object);
 
         // try fetching first by exact name, if not found, try camel-case
-        if (!$property = self::reflectionProperty($class, $name)) {
-            $property = self::reflectionProperty($class, self::camel($name));
+        $property = self::reflectionProperty($class, $name);
+
+
+        if (!$property && $property = self::reflectionProperty($class, self::camel($name))) {
+            trigger_deprecation('zenstruck\foundry', '1.5.0', 'Using snake/kebab-case attributes is deprecated, use the same case as the object property instead.');
```
wouterj, line 158: This message is slightly incorrect, as it assumes the property name is camelCase. I would remove explicit case names, e.g: "Using a different case attribute is deprecated, [...]."

kbond: Changed to: "'Using a differently cased attribute is deprecated, use the same case as the object property instead.'"

weaverryan (approved):
seems reasonable :)

PR #84, opened 2020-11-09
Add `Factory::delayFlush()`

### LINKED ISSUE #37: Single flush per "user-land" create
Currently, the following code calls `$em->flush()` twice. First, when creating the nested `CategoryFactory` and a second time after persisting the `PostFactory`.

```php
PostFactory::new()->create([
    'category' => CategoryFactory::new(),
]);
```

It would be better if the nested `CategoryFactory` was just persisted, then after the `PostFactory` is persisted, a single flush occurs capturing all changes.

### LINKED ISSUE #265: How to use Factory::createMany() with multiple predefined datasets
I would like to insert multiple fixture records based on an array of predefined info. Currently I simply loop over an array and do multiple `Factory::createOne()` statements, but this is slow as it flushes each one separately. This is a simplified example but imagine there are a lot more records etc.

```php
final class ContactRoleStory extends Story
{
    /**
     * @var array<array<string, string>>
     */
    final public const CONTACT_ROLES = [
        [
            'slug' => 'prospect',
            'name' => 'Prospective Guest'
        ],
        [
            'slug' => 'guest',
            'name' => 'Guest'
        ],
        [
            'slug' => 'marketing',
            'name' => 'Marketing Rep'
        ],
        [
            'name' => 'Potential Staff'
        ],
        [
            'name' => 'Supplier'
        ],
        [
            'slug' => 'staff',
            'name' => 'Staff'
        ],
    ];

    public function build(): void
    {
        foreach (self::CONTACT_ROLES as $role)
        {
            ContactRoleFactory::createOne([
                'role' => $role['name'],
                'slug' => $role['slug'] ?? ''
            ]);
        }
    }
}
```

Is it possible to do this and only flush after all the records are added? I'm guessing using `Factory::new()->many()` with a callback or something like that?

Thanks

kbond:
Closes #37. Closes #265.

Using this in a data fixture to create 1000 objects that each creates a related object has a significant performance improvement (~20x).

```php
Factory::delayFlush(function() {
    PostFactory::createMany(10000);

    // etc...
});
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -865,6 +865,22 @@ Foundry can be used to create factories for entities that you don't have model f
     $entity = create(Post::class, ['field' => 'value']);
     $entities = create_many(Post::class, 5, ['field' => 'value']);
 
+Delay Flush
+~~~~~~~~~~~
+
+When creating/persisting many factories at once, it can be improve performance
+to instantiate them all without saving to the database, then flush them all at
+once. To do this, wrap the operations in a ``Factory::delayFlush()`` callback:
+
+.. code-block:: php
+
+    use Zenstruck\Foundry\Factory;
+
+    Factory::delayFlush(function() {
+        CategoryFactory::createMany(100); // instantiated/persisted but not flushed
+        TagFactory::createMany(200); // instantiated/persisted but not flushed
+    }); // single flush
+
 .. _without-persisting:
 
 Without Persisting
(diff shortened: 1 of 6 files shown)


kbond:
@benr77, can you give a final confirmation that this should solve #265?

benr77:
This is looking good. My fixture loading for a massive set of global data was 1m 26s and having applied this fix to just a few of the worst offenders it is already down to 55s. Nice one!

PR #85, opened 2020-11-10
Add psalm static analysis

wouterj:
As promised in https://github.com/zenstruck/foundry/pull/79#issuecomment-724091511 , this is a PR introducing Psalm analysis to this package.

I've split this into multiple commits:

* de928ff adds (mostly) [templated annotations](https://psalm.dev/docs/annotating_code/templated_annotations/) to make Psalm understand the magic in `Proxy` and `RepositoryProxy`. E.g. `@psalm-var Proxy<User> $userProxy` is now understood by Psalm as a User proxy (so all methods of `User` are now allowed in `$userProxy`).
* e9f1fe6 and 11457c7 fixes issues found by Psalm (level 6 and 5) by changing PHP code.

I tried running level 4, but at that level all errors were "invalid" and I wasn't able to explain the logic to Psalm using annotations. I've asked about this and if I get a solution, we can continue to go down the error levels. In total there are 61 more "errors" found by Psalm at this point (which are ignored because of the set level).

> Semi related notes:
> - This PR is still using Psalm 3.x, as Psalm 4 is not yet supported by the Doctrine plugin: https://github.com/weirdan/doctrine-psalm-plugin/pull/78
> - I have some outstanding questions about Psalm in Slack related to this package. I've suppressed these issues for now, hopefully it's either a bug that can be fixed or something that I did wrong.

### DIFF
--- a/src/RepositoryProxy.php
+++ b/src/RepositoryProxy.php
@@ -8,13 +8,14 @@ use Doctrine\Persistence\ObjectRepository;
 use PHPUnit\Framework\Assert;
 
 /**
- * @mixin EntityRepository
+ * @mixin EntityRepository<TProxiedObject>
+ * @template TProxiedObject of object
  *
  * @author Kevin Bond <kevinbond@gmail.com>
  */
 final class RepositoryProxy implements ObjectRepository, \IteratorAggregate, \Countable
 {
-    /** @var ObjectRepository */
+    /** @var ObjectRepository<TProxiedObject> */
     private $repository;
 
     public function __construct(ObjectRepository $repository)
@@ -122,6 +123,8 @@ final class RepositoryProxy implements ObjectRepository, \IteratorAggregate, \Co
 
     /**
      * @return Proxy|object|null
+     *
+     * @psalm-return Proxy<TProxiedObject>|null
      */
     public function first(string $sortedField = 'id'): ?Proxy
     {
@@ -130,6 +133,8 @@ final class RepositoryProxy implements ObjectRepository, \IteratorAggregate, \Co
 
     /**
      * @return Proxy|object|null
+     *
+     * @psalm-return Proxy<TProxiedObject>|null
      */
     public function last(string $sortedField = 'id'): ?Proxy
     {
@@ -162,6 +167,8 @@ final class RepositoryProxy implements ObjectRepository, \IteratorAggregate, \Co
      * @return Proxy|object
      *
      * @throws \RuntimeException if no objects are persisted
+     *
+     * @psalm-return Proxy<TProxiedObject>
      */
     public function random(): Proxy
     {
@@ -224,6 +231,9 @@ final class RepositoryProxy implements ObjectRepository, \IteratorAggregate, \Co
      * @param object|array|mixed $criteria
      *
      * @return Proxy|object|null
+     *
+     * @psalm-param Proxy<TProxiedObject>|array|mixed $criteria
+     * @psalm-return Proxy<TProxiedObject>|list<Proxy<TProxiedObject>>|null
      */
     public function find($criteria): ?Proxy
     {
@@ -260,6 +270,8 @@ final class RepositoryProxy implements ObjectRepository, \IteratorAggregate, \Co
      * @return Proxy|object|null
      *
      * @throws \RuntimeException if the wrapped ObjectRepository does not have the $orderBy parameter
+     *
+     * @psalm-return Proxy<TProxiedObject>|null
      */
     public function findOneBy(array $criteria, ?array $orderBy = null): ?Proxy
     {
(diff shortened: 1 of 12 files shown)


src/RepositoryProxy.php
```diff
@@ -8,13 +8,15 @@
 use PHPUnit\Framework\Assert;
 
 /**
- * @mixin EntityRepository
+ * todo: this mixin already existed, but with another syntax. Will this break something?
+ * @mixin EntityRepository<TProxiedObject>
```
wouterj, line 12: Psalm doesn't supported a prefixed `psalm-mixin` annotation. I wonder if we break something that was working before by using this syntax?

src/RepositoryProxy.php
```diff
@@ -43,6 +45,8 @@ public function count(): int
 
     public function getIterator(): \Traversable
     {
+        // TODO: $this->repository is set to ObjectRepository, which is not
+        //       iterable. Can this every be another RepositoryProxy?
         if (\is_iterable($this->repository)) {
             return yield from $this->repository;
         }
```
wouterj, line 52: Psalm says "`$this->repository` is never iterable", because it's typehinted as `ObjectRepository`. Is this indeed a redundant if statement, or are there situations where `$this->repository` is already a `RepositoryProxy`? (which case we need to update its PHPdoc to `ObjectRepository<TProxiedObject>|RepositoryProxy<TProxiedObject>`)

src/RepositoryProxy.php
kbond: I only added as a helper (for PhpStorm) because most people's repository's would be an instance of `EntityRepository` but this may not be the case. For instance, in my apps, my repositories are instances of `ObjectRepository` but not `EntityRepository`.

Does PhpStorm still autocomplete entity repository with this change? That would be my only concern.

src/RepositoryProxy.php
kbond: There isn't a situation where this `$this->repository` could already be a `RepositoryProxy` but I think it's valid that the `ObjectRepository` could be iterable. In my apps, all my repository's are instances of `ObjectRepository`, iterable and countable.

kbond:
Awesome @wouterj! This will be a great resource to help me dive into static analysis in PHP.

What's the `--output-format=github` do? I can't seem to find docs on this.

BTW, not sure what's up with the test failures. They all passed in 8364424c7d2420e7aeea7263664ecf9643f00d7a...

wouterj:
> What's the --output-format=github do? I can't seem to find docs on this.

Instead of outputting the errors in the console, it'll use the GitHub Actions message commands. See e.g. the reporting in https://github.com/vimeo/psalm/pull/4505/files
image

src/RepositoryProxy.php
wouterj: Just tested it by trying to autocomplete `UserFactory::repository()->createQueryBuilder()` (and removing the `UserRepository|RepositoryProxy repository()` PHPdoc on the `UserFactory`). This works fine, so I think PHPstorm ignores the `<...>` part.

src/RepositoryProxy.php
kbond: Could we update the PHPdoc to `ObjectRepository<TProxiedObject>|iterable<TProxiedObject>|\Countable`?

kbond:
That's really cool! I think I once saw a phpunit extension that does the same for failed tests. I've been meaning to investigate that at some point.

src/RepositoryProxy.php
wouterj: I believe the [union type](https://psalm.dev/docs/annotating_code/type_syntax/union_types/) is considered exclusive. So doing `ObjectRepository|iterable|\Countable` means that this property is either `ObjectRepository` or `iterable` or `\Countable`. This means that all code using `$this->repository` as if it's always `ObjectRepository` will be considered by Psalm.

Psalm uses the [intersection type (e.g. `ObjectRepository&iterable`)](https://psalm.dev/docs/annotating_code/type_syntax/intersection_types/) as an inclusive type. However this means that any code setting the property to an instance of `ObjectRepository` that isn't `iterable` will produce an error.

However, in this case the issue is ignored by the current Psalm level. So let's just ignore this for now. I think the best solution - if this becomes a real error - is to add `@psalm-suppress ...` (this is like saying "I know this strictly is wrong, but it's here for a reason").

wouterj:
As for the static analysis in this PR, I think this is now finished up to a state is that mergable. Beyond level 4, Psalm starts to be a lot more strict on "possible null" values, which requires lots of extra checks (e.g. for any service fetched in `DatabaseResetter`, we need to check whether the service actually exists). While this might make sense, let's not broaden the scope of this PR any more :)

I'm unsure how I managed to break all tests :confused:

kbond:
@wouterj, maybe a rebase capturing #86 will fix the tests?

wouterj:
I just discovered I wrongly fixed a merge conflict before, which resulted in `phpunit/phpunit` to still be a dev dependency of this package. Removing that has fixed the tests :tada:

kbond:
Awesome, once you move this out of draft I'll merge (so I know when your ready).

kbond:
Thanks @wouterj - I'm sure I'll be referring to this PR a lot :)

wouterj:
You're welcome :) 

The amount of templated annotations required to get static analysis working in this application is a bit rare btw. The packages I normally work on only needed one or two templated annotations.

PR #91, opened 2020-11-25
Added docs for ModelFactory::new()

Nyholm:
When I first started to use this library, I had a hard time understanding why there were two very similar ways doing the same thing: 

```php
PostFactory::new()->create(['title' => 'My Title']); // A
PostFactory::new(['title' => 'My Title'])->create(); // B
```

Of course I put my bets on the wrong horse and used B all over.. That is not what one should do.. 
I've added some config that will make it more clear that there is a difference and that one (most of the time) should use A. 

-----

Is this clear enough? Should I expand on my example by doing `$postFactory = PostFactory::new([...]);`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,15 +267,11 @@ $title = $post->getTitle(); // getTitle() can be autocompleted by your IDE!
 
 // if you need the actual Post object, use ->object()
 $realPost = $post->object();
-
-PostFactory::new()->create(['title' => 'My Title']); // override defaults 
-PostFactory::new(['title' => 'My Title'])->create(); // alternative to above
+PostFactory::new()->create(['title' => 'My Title']);
 
 // create/persist 5 Posts with random data from getDefaults()
 PostFactory::new()->createMany(5); // returns Post[]|Proxy[]
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -290,9 +286,21 @@ $posts = PostFactory::randomSet(4); // array containing 4 "Post|Proxy" objects
...
 cause the factory to not be instantiated properly. Always instantiate with `PostFactory::new()`.
 
+The first argument to `PostFactory::new()` will allow you to overwrite the default
+values that are defined in the `PostFactory::getDefaults()`.
+
+```php
+use App\Factory\PostFactory;
+
+PostFactory::new(['title' => 'My Title'])->create();
+PostFactory::new(['title' => 'My Title'])->createMany(5);
```
kbond, line 301: > Is this clear enough? Should I expand on my example by doing $postFactory = PostFactory::new([...]);?

Here you mean? Yeah, I think that might in understanding that all factories created will have this new default.

Nyholm: Yes. Exactly.

Nyholm: Updated!

kbond:
Thanks!

Nyholm:
Thank you for merging

PR #95, opened 2020-12-08
Re-enable Psalm and fixed annotations

wouterj:
Not entirely sure why I needed a function-scope template for `proxyResult()`, but it did the trick. It makes a some sense, as nothing forbids you to pass a non-`TProxiedObject` object as argument.

Meanwhile, I removed the weirdan doctrine plugin as it's no longer needed (Doctrine added Psalm annotations in doctrine/orm#8289 ).


kbond:
Thanks @wouterj!

PR #96, opened 2020-12-08
Streamline GitHub CI by using ramsey/composer-install

wouterj:
This action gains some traction in PHP OSS projects. It automatically caches dependencies and runs composer with the correct flags, making the CI configuration a bit more readable.


kbond:
Thanks @wouterj and good find! Just the other day I was trying to figure out why caching didn't appear to be working. Glad to see this action should take care of it for me.

PR #99, opened 2020-12-16
psalm-suppress InternalMethod

kbond:
dama/doctrine-test-bundle 6.5 marked StaticDriver as internal

Ping @wouterj.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,6 +151,7 @@ jobs:
           php-version: 7.4
           extensions: pgsql
           coverage: xdebug
+          ini-values: xdebug.mode=coverage
 
       - name: Install dependencies
         uses: ramsey/composer-install@v1
--- a/psalm.xml
+++ b/psalm.xml
@@ -24,6 +24,13 @@
                 <referencedClass name="PHPUnit\Framework\Assert"/>
             </errorLevel>
         </UndefinedClass>
+        <InternalMethod>
+            <errorLevel type="suppress">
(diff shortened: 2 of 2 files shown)


wouterj:
I think it's better to discuss how to fix this in the dama/doctrine-test-bundle. The issue here is that the debug class loader will trigger deprecations if you're using an internal class outside the package it was defined in. So I believe all PHPunit bridge test runs using Foundry will now show an unfixable deprecation.

kbond:
I'm not seeing a direct deprecation warning regarding this with phpunit bridge but it could be the way this [class is implemented in dama/doctrine-test-bundle](https://github.com/dmaicher/doctrine-test-bundle/blob/master/src/DAMA/DoctrineTestBundle/Doctrine/DBAL/StaticDriver.php).

wouterj:
Ah, that must be it.

I just got the report while testing out Collision, so it definitely is reporting it:

```
  Tests:  47 passed
  Time:   15.87s

Other deprecation notices (42)

    ...

  1x: The "NunoMaduro\Collision\Adapters\Phpunit\Printer" class implements "PHPUnit\Framework\TestListener" that is deprecated Use the `TestHook` interfaces instead.

  1x: The "PHPUnit\Util\Printer" class is considered internal This class is not covered by the backward compatibility promise for PHPUnit. It may change without further notice. You should not use it from "NunoMaduro\Collision\Adapters\Phpunit\Printer".
```

wouterj (approved):

PR #111, opened 2021-01-16
[RFC] Remove "visual duplication" of ModelFactory::new()->create()

wouterj:
This is an approach on removing "visual duplication" when creating new models:

```php
// before
$post = PostFactory::new()->create();
$post = PostFactory::new()->published()->create();
$posts = PostFactory::new()->createMany(3);
$posts = PostFactory::new()->published()->createMany(3);

// after
$post = PostFactory::createOne();
$post = PostFactory::new()->published()->create();
$posts = PostFactory::createMany(3);
$posts = PostFactory::new()->published()->many(3)->create();
```

We can't use `PostFactory::create()` as it conflicts with the `Factory#create()` instance method. <s>I used the magic method name to clearly indicate that we're not creating a new factory, but we are creating a new entity. An alternative name that doesn't need the magic `__callStatic()` method can be `createSingle()` (which seems consistent with `createMany()`).</s>

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ use Zenstruck\Foundry\ModelFactory;
 use Zenstruck\Foundry\Proxy;
 
 /**
+ * @method static Post|Proxy createOne(array $attributes = [])
  * @method static Post|Proxy findOrCreate(array $attributes)
  * @method static Post|Proxy random()
  * @method static Post[]|Proxy[] randomSet(int $number)
@@ -257,17 +258,20 @@ protected function getDefaults(): array
 ```php
 use App\Factory\PostFactory;
 
-PostFactory::new()->create(); // create/persist Post with random data from `getDefaults()`
+// create/persist Post with random data from `getDefaults()`
+PostFactory::createOne();
 
-// create() returns the persisted Post object wrapped in a Proxy object
-$post = PostFactory::new()->create();
+// or provide values for some properties (others will be random)
+PostFactory::createOne(['title' => 'My Title']);
+
+// createOne() returns the persisted Post object wrapped in a Proxy object
+$post = PostFactory::createOne();
 
 // the "Proxy" magically calls the underlying Post methods and is type-hinted to "Post"
 $title = $post->getTitle(); // getTitle() can be autocompleted by your IDE!
 
 // if you need the actual Post object, use ->object()
 $realPost = $post->object();
-PostFactory::new()->create(['title' => 'My Title']);
 
 // create/persist 5 Posts with random data from getDefaults()
 PostFactory::new()->createMany(5); // returns Post[]|Proxy[]
@@ -284,22 +288,6 @@ $posts = PostFactory::randomSet(4); // array containing 4 "Post|Proxy" objects
 
 // random range of persisted objects
 $posts = PostFactory::randomRange(0, 5); // array containing 0-5 "Post|Proxy" objects
-```
-
-### Instantiate your `ModelFactory`
-
-One should never instantiate your `ModelFactory` with the constructor (ie `new PostFactory()`). This will
-cause the factory to not be instantiated properly. Always instantiate with `PostFactory::new()`.
-
-The first argument to `PostFactory::new()` will allow you to overwrite the default
-values that are defined in the `PostFactory::getDefaults()`.
-
-```php
-use App\Factory\PostFactory;
-
-$factory = PostFactory::new(['title' => 'My Title']);
-$factory->create();
-$factory->createMany(5);
 ```
 
 ### Reusable Model Factory "States"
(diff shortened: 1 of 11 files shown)


kbond:
I think I like `createSingle()` as it fully removes the "visual duplication".

Now that `createMany()` is just an alias for `->many()->create()` what about making it static so:

```php
PostFactory::createSingle(): Post
PostFactory::createMany(): Post[]
```

Could this be handled w/o a bc break with some `__call()/`__callStatic()` trickery?

kbond:
Or what about `createOne()`? Would sort of match `findOneBy` in doctrine...

wouterj:
I've just updated everything to `createSingle()`, but I'm +1 for `createOne()` :) 

> Now that createMany() is just an alias for ->many()->create() what about making it static so [...]

I've included this change in the PR and deprecated the `createMany()` instance method. Let me know what you think.

As you see, the build breaks because of the `create_many()` helper function. I'm not sure what you want to do here, deprecate the function? (I can't see how to implement this function without a `createMany()` instance method)

src/Bundle/Resources/skeleton/Factory.tpl.php
```diff
@@ -10,6 +10,7 @@
 use Zenstruck\Foundry\Proxy;
 
 /**
+ * @method static <?= $entity->getShortName() ?>|Proxy createSingle(array $attributes = [])
```
kbond, line 13: `createMany` below should be marked as `static` now, right?

kbond: Or will that typehint interfere with the non-static createMany() `@method` on `Factory`?

kbond:
> build breaks because of the create_many() helper function. I'm not sure what you want to do here

What about:

```php
function create_many(int $number, string $class, $attributes = []): array
{
    return factory($class)->many($number)->create($attributes);
}
```

src/Bundle/Resources/skeleton/Factory.tpl.php
wouterj: I've now made it static, didn't try autocompletion yet (I don't normally use an IDE when working on small projects like this)

kbond: Regarding auto-completion, I think the'll be some issues mixing static and non-static `@method`'s of the same name. At least with PhpStorm.

wouterj: As the instance method is deprecated, should we only document the static method in the PHPdoc?

wouterj: I've updated the PR with my latest suggestion

src/ModelFactory.php
```diff
@@ -6,6 +6,9 @@
  * @template TModel of object
  * @template-extends Factory<TModel>
  *
+ * @method static Proxy[]|object[] createMany(int $number, array|callable $attributes = []): array
```
kbond, line 9: I think the trailing `: array` can be removed.

src/Bundle/Resources/skeleton/Factory.tpl.php
kbond: Not ideal but I think this is best.

kbond:
I love how the tests read now (after replacing `::new()->create()` with `::createOne()`)!

weaverryan (approved):
I love this.

no more new() call for most cases (where you’re just creating one or many and maybe passing in some attributes). Then, the longer ->new() only when you need extra config/setup for your factory.

wouterj:
Updated the PR to also update the README with the new API. I didn't find any CHANGELOG file in the repository, I guess you're doing that manually while releasing?

The actions fail because of 2 uncovered lines of code (the bad method call exceptions). I don't think these should be tested, but if you want I can add a test.

kbond:
Awesome, looks great!

> I didn't find any CHANGELOG file in the repository, I guess you're doing that manually while releasing?

Yeah, I use the Github releases as a changelog right now.

> The actions fail because of 2 uncovered lines of code

No worries - I only use these as a loose metric - I'm not strict about it. I've been meaning to increase the "passable" threshold from 0%

README.md
```diff
@@ -286,22 +290,6 @@ $posts = PostFactory::randomSet(4); // array containing 4 "Post|Proxy" objects
 $posts = PostFactory::randomRange(0, 5); // array containing 0-5 "Post|Proxy" objects
 ```
 
-### Instantiate your `ModelFactory`
```
Nyholm, line 289: Can I run `new PostFactory()` now? Or why was this removed?

Nyholm (approved):

README.md
wouterj: No, you still have to use `::new()`. However, you'll only use `new()` in complex cases, otherwise you use `Factory::createOne(...)` or `Factory::createMany(...)`. So I removed the main section about this and instead moved this into the complex code examples (line 331 and 361)

kbond:
Thanks @wouterj!

Nyholm (commented):
(submitting my comment I wrote yestetday)

PR #123, opened 2021-02-19
Add static ModelFactory helpers

### LINKED ISSUE #105: [Suggestion] Allow to get the object from assertExists()
```php
UserFactory::repository()->assertExists([
    'fullName' => $this->userData['full-name'],
    'username' => $this->userData['username'],
]);

$user = UserFactory::repository()->find(['fullName' => $this->userData['full-name'], 'username' => $this->userData['username']])->object();
$this->assertTrue($container->get('security.password_encoder')->isPasswordValid($user, $this->userData['password']));
```

It would be nice if `$user` is returned from `assertExists()`, to avoid duplication. However, this will break the current fluent interface, so I'm not sure if (or how) we should fix this.

### LINKED ISSUE #124: Add repository methods to ModelFactory as static methods
#123 made me think that more `RepositoryProxy` methods should be available on the `ModelFactory` as static methods:

- [x] `PostFactory::count(): int`
- [x] `PostFactory::assert()->empty(): void`
- [x] `PostFactory::assert()->count(): void`
- [x] `PostFactory::assert()->countGreaterThan(): void`
- [x] `PostFactory::assert()->countGreaterThanOrEqual(): void`
- [x] `PostFactory::assert()->countLessThan(): void`
- [x] `PostFactory::assert()->countLessThanOrEqual(): void`
- [x] `PostFactory::assert()->exists(): void` ~possible solution to #105~
- [x] `PostFactory::assert()->notExists(): void`
- [x] `PostFactory::truncate(): void`
- [x] `PostFactory::find(): Proxy` possible solution to #105
- [x] ~`PostFactory::findAll(): Proxy[]`~ `PostFactory::all(): Proxy[]`
- [x] `PostFactory::findBy(): Proxy[]`
- [x] ~`PostFactory::findOneBy(): Proxy`~ merge with `find()` above
- [x] `PostFactory::first(): Proxy`
- [x] `PostFactory::last(): Proxy`

@weaverryan, @wouterj WDYT? I'm wondering if we should use `__callStatic` here to accomplish?

This could possibly make `RepositoryProxy` just an implementation detail and not something you'd ever really need to use in user-land.

kbond:
I find in a lot of my tests, I need to access the only object in the db. I can do `MyFactory::repository()->first()` but it is a little verbose and does not have auto-completion. With this PR, I can do `MyFactory::first()`.

I added `MyFactory::last()` and `$sortField` options to match the repository proxy. The methods on the repository proxy allow null to be returned, I think it should not but kept for BC. These new methods throw an exception if no entities exist.

See #124 for reference.

(Closes #124, Closes #105)


wouterj:
I'm always using `::random()` in these situations (I don't think order should matter here). However, no strong opinion here (as the method already exists).

kbond:
Yeah, `random()` would work too but I think it's a little more descriptive to say `first()` when reading your tests later. I'll modify this PR to handle #124.

wouterj:
Yeah, my personal foundry best practice (which btw is still evolving with the week) is to either use `::random()` (to say "I just need an entity, don't care which one" e.g. for a blog post author) or `::findOrCreate()` (to say "I need an entity with some specific requirements") whenever I need to fetch an entity in my tests/other factories.

kbond:
@wouterj, I've marked this as a solution to your suggestion in #105. It isn't exactly what you proposed but I think `ModelFactory::find()` will do the same thing. Let me know if you disagree.

kbond:
I think this PR is good to go. @wouterj, when you have a chance, mind doing a review?

kbond:
> my personal foundry best practice (which btw is still evolving with the week) is to either use ::random() (to say "I just need an entity, don't care which one" e.g. for a blog post author)

I have a lot of tests that make some http request, then I need to find the entity that was created. I don't necessarily know any attributes of the created entity so `find()` doesn't work. `random()` would work but it implies there are several entities to choose from. To me, `first()`/`last()` better implies I'm getting the object created by the http request (actually, `last()` is better now that I think about it).

Issue #113, opened 2021-01-16
Auto-refresh proxy improvements


kbond:
With [auto-refreshing](https://github.com/zenstruck/foundry#auto-refresh), there is a bit of a problem:

```php
$post = PostFactory::new(['title' => 'Original Title', 'body' => 'Original Body'])
    ->create()
    ->enableAutoRefresh()
;

$post->setTitle('New Title');
$post->setBody('New Body'); // !!! this causes the object to be refreshed, which means the "New Title" title was replaced by the database contents
$post->save();

$post->getBody(); // "New Body"
$post->getTitle(); // "Original Title" !! because the subsequent call to ->setBody() "auto-refreshed" the object
```

This is the reason auto-refreshing is disabled by default.

A great improvement would be to only refresh when it actually needs to be refreshed (it's row in the db changed).

If this could be fixed, I think we could enable auto-refreshing by default.

kbond:
One potential solution (thanks @wouterj!)  is to call `$om->flush()` before returning any `Proxy` methods that potentially change the object: `__set()`, `__call()`, `__unset()`, `forceSetAll()`. This would be an _auto-save_ feature.

```php
$post = PostFactory::new(['title' => 'Original Title', 'body' => 'Original Body'])
    ->create()
    ->enableAutoRefresh()
;

$post->setTitle('New Title'); // refreshes from the db, call setTitle() on the object, auto-save
$post->setBody('New Body'); // since above auto-saved, when this refreshes the object, it's title is still "New Title"
$post->save(); // not required as above auto-saved

$post->getBody(); // "New Body"
$post->getTitle(); // "New Title"
```

Issue #110, opened 2021-01-16
2.0 Ideas


kbond:
@wouterj and I have been discussing some potential changes for 2.0. This is a list of potential ideas:

- [x] #37
- [ ] Rename `ProxyRepository` to `RepositoryDecorator` and mark internal ("proxy" is a bit of a misnomer - it is actually a "decorator")
- [ ] Remove `Proxy` and change `Factory::create()` to return the real doctrine entity. While the proxy idea is clever and sort of creates a "quasi" AR system for your entities, it can be confusing (to use and document)
- [ ] Move the AR helpers (save, refresh, delete) to the `ModelFactory`/`RepositoryDecorator` and some stand alone helper functions
- [ ] Investigate if `ocramius/proxy-manager` could be used for auto-refresh as this would be lost with removal of the `Proxy`
- [x] Method names: `$user = UserFactory::new()->create()->object()` is unfortunate. The `->object()` would be removed by removing the `Proxy` (above) but `$user = UserFactory::new()->create()` is still confusing... (solved in #111)
- [x] #113 
- [ ] `ModelFactory::repository()`/`AnonymousFactory::repository()` to return the "real" repository for the entity
- [x] #126 
- [x] Move reset env variables to config options (`FOUNDRY_DISABLE_DATABASE_RESET`, `FOUNDRY_RESET_CONNECTIONS`, `FOUNDRY_RESET_OBJECT_MANAGERS`)? https://github.com/zenstruck/foundry/pull/320
- [x] Remove _bundless_ usage. This is a vestigial feature that likely only I'm using (with no reason I can't switch to using the bundle). I believe only allowing bundle config would drastically reduce complexity. https://github.com/zenstruck/foundry/issues/319
- [ ] [BC BREAK] make `ModelFactory::getClass()` public and remove `ModelFactory::getEntityClass()`? See #162 for details (specifically https://github.com/zenstruck/foundry/pull/162#discussion_r632956940)
- [x] ODM Support (or at least make adding support for this easier)

wouterj:
> Method names: $user = UserFactory::new()->create()->object() is unfortunate. The ->object() would be removed by removing the Proxy (above) but $user = UserFactory::new()->create() is still confusing...

See #111 for a proposal to fix this

---

(btw, you referenced the wrong wouter, I'm @wouterj :wink: )

nikophil:
Hello @kbond 
_sorry if it's not the correct place to ask kind of things_

[...]

PR #127, opened 2021-02-28
Try out ProxyManager

wouterj:
I started trying out some of the ideas in #110 and #113.

The idea of this change is that instead of a "fake" `Proxy` object, we return a proxied entity/model. This allows removing `->object()` calls and removing some confusion when trying to pass the objects into typehinted functions. For this to work, the solution has to fulfill these promises:

* [x] The proxied entity/model has to support auto-refresh at least as good as the current `Proxy`
* [x] The AR helpers (`save()`, `remove()`, ...) have to be moved to `Factory`/`ModelFactory`
* [x] In 1.x, this new behavior should be opt-in per factory. This allows a smooth migration path, before removing the old factories in 2.0
* [ ] Reintroduce support for global auto refresh setting
* [ ] Document UPGRADE guide + deprecate the old Proxy

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ $post = PostFactory::new() // Create the factory for Post objects
 The factories can be used inside [DoctrineFixturesBundle](https://symfony.com/doc/master/bundles/DoctrineFixturesBundle/index.html)
 to load fixtures or inside your tests, [where it has even more features](#using-in-your-tests).
 
-Want to watch a screencast 🎥 about it? Check out https://symfonycasts.com/foundry
+Want to watch a screencast  about it? Check out https://symfonycasts.com/foundry
 
 ## Documentation
 
@@ -44,19 +44,19 @@ Want to watch a screencast 🎥 about it? Check out https://symfonycasts.com/fou
 5. [Using in your Tests](#using-in-your-tests)
     1. [Enable Foundry in your TestCase](#enable-foundry-in-your-testcase)
     2. [Database Reset](#database-reset)
-    3. [Object Proxy](#object-proxy)
-        1. [Force Setting](#force-setting)
-        2. [Auto-Refresh](#auto-refresh)
-    4. [Repository Proxy](#repository-proxy)
-    5. [Assertions](#assertions)
-    6. [Global State](#global-state)
-    7. [PHPUnit Data Providers](#phpunit-data-providers)
-    8. [Performance](#performance)
+    3. [Force Setting](#force-setting)
+    4. [Object Proxy](#object-proxy)
+        1. [Auto-Refresh](#auto-refresh)
+    5. [Repository Proxy](#repository-proxy)
+    6. [Assertions](#assertions)
+    7. [Global State](#global-state)
+    8. [PHPUnit Data Providers](#phpunit-data-providers)
+    9. [Performance](#performance)
         1. [DAMADoctrineTestBundle](#damadoctrinetestbundle)
         2. [Miscellaneous](#miscellaneous)
-    9. [Non-Kernel Test](#non-kernel-tests)
-    10. [Test-Only Configuration](#test-only-configuration)
-    11. [Using without the Bundle](#using-without-the-bundle)
+    10. [Non-Kernel Test](#non-kernel-tests)
+    11. [Test-Only Configuration](#test-only-configuration)
+    12. [Using without the Bundle](#using-without-the-bundle)
 6. [Stories](#stories)
     1. [Stories as Services](#stories-as-services)
     2. [Story State](#story-state)
@@ -193,23 +193,22 @@ use App\Entity\Post;
 use App\Repository\PostRepository;
 use Zenstruck\Foundry\RepositoryProxy;
 use Zenstruck\Foundry\ModelFactory;
-use Zenstruck\Foundry\Proxy;
 
 /**
- * @method static Post|Proxy createOne(array $attributes = [])
- * @method static Post[]|Proxy[] createMany(int $number, $attributes = [])
- * @method static Post|Proxy find($criteria)
- * @method static Post|Proxy findOrCreate(array $attributes)
- * @method static Post|Proxy first(string $sortedField = 'id')
- * @method static Post|Proxy last(string $sortedField = 'id')
- * @method static Post|Proxy random(array $attributes = [])
- * @method static Post|Proxy randomOrCreate(array $attributes = []))
- * @method static Post[]|Proxy[] all()
- * @method static Post[]|Proxy[] findBy(array $attributes)
- * @method static Post[]|Proxy[] randomSet(int $number, array $attributes = []))
- * @method static Post[]|Proxy[] randomRange(int $min, int $max, array $attributes = []))
+ * @method static Post createOne(array $attributes = [])
+ * @method static Post[] createMany(int $number, $attributes = [])
+ * @method static Post find($criteria)
+ * @method static Post findOrCreate(array $attributes)
+ * @method static Post first(string $sortedField = 'id')
+ * @method static Post last(string $sortedField = 'id')
+ * @method static Post random(array $attributes = [])
+ * @method static Post randomOrCreate(array $attributes = []))
+ * @method static Post[] all()
+ * @method static Post[] findBy(array $attributes)
+ * @method static Post[] randomSet(int $number, array $attributes = []))
+ * @method static Post[] randomRange(int $min, int $max, array $attributes = []))
  * @method static PostRepository|RepositoryProxy repository()
- * @method Post|Proxy create($attributes = [])
+ * @method Post create($attributes = [])
  */
 final class PostFactory extends ModelFactory
 {
@@ -273,21 +272,16 @@ PostFactory::createOne();
 // or provide values for some properties (others will be random)
 PostFactory::createOne(['title' => 'My Title']);
 
-// createOne() returns the persisted Post object wrapped in a Proxy object
+// createOne() returns the persisted Post object
 $post = PostFactory::createOne();
-
-// the "Proxy" magically calls the underlying Post methods and is type-hinted to "Post"
-$title = $post->getTitle(); // getTitle() can be autocompleted by your IDE!
-
-// if you need the actual Post object, use ->object()
-$realPost = $post->object();
+$title = $post->getTitle();
 
 // create/persist 5 Posts with random data from getDefaults()
-PostFactory::createMany(5); // returns Post[]|Proxy[]
+PostFactory::createMany(5); // returns an array of Post objects
 PostFactory::createMany(5, ['title' => 'My Title']);
 
 // find a persisted object for the given attributes, if not found, create with the attributes
-PostFactory::findOrCreate(['title' => 'My Title']); // returns Post|Proxy
+PostFactory::findOrCreate(['title' => 'My Title']); // returns a Post object
 
 PostFactory::first(); // get the first object (assumes an auto-incremented "id" column)
 PostFactory::first('createdAt'); // assuming "createdAt" is a datetime column, this will return latest object
@@ -298,15 +292,15 @@ PostFactory::truncate(); // empty the database table
 
 PostFactory::count(); // the number of persisted Posts
 
-PostFactory::all(); // Post[]|Proxy[] all the persisted Posts
+PostFactory::all(); // Post[] all the persisted Posts
 
-PostFactory::findBy(['author' => 'kevin']); // Post[]|Proxy[] matching the filter
+PostFactory::findBy(['author' => 'kevin']); // Post[] matching the filter
 
-$post = PostFactory::find(5); // Post|Proxy with the id of 5
-$post = PostFactory::find(['title' => 'My First Post']); // Post|Proxy matching the filter
+$post = PostFactory::find(5); // Post with the id of 5
+$post = PostFactory::find(['title' => 'My First Post']); // Post matching the filter
 
 // get a random object that has been persisted
-$post = PostFactory::random(); // returns Post|Proxy
+$post = PostFactory::random(); // returns Post
 $post = PostFactory::random(['author' => 'kevin']); // filter by the passed attributes
 
 // or automatically persist a new random object if none exists
@@ -314,13 +308,16 @@ $post = PostFactory::randomOrCreate();
 $post = PostFactory::randomOrCreate(['author' => 'kevin']); // filter by or create with the passed attributes
 
 // get a random set of objects that have been persisted
-$posts = PostFactory::randomSet(4); // array containing 4 "Post|Proxy" objects
+$posts = PostFactory::randomSet(4); // array containing 4 "Post" objects
 $posts = PostFactory::randomSet(4, ['author' => 'kevin']); // filter by the passed attributes
 
 // random range of persisted objects
-$posts = PostFactory::randomRange(0, 5); // array containing 0-5 "Post|Proxy" objects
+$posts = PostFactory::randomRange(0, 5); // array containing 0-5 "Post" objects
 $posts = PostFactory::randomRange(0, 5, ['author' => 'kevin']); // filter by the passed attributes
 ```
+
+Instead of the real Post object, all factory methods return a proxy object wrapping the Post object. This proxy behaves
+exactly like the real object (including type hints). Read more about this in [Proxy object](#proxy-object).
 
 ### Reusable Model Factory "States"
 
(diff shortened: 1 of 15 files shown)


tests/Unit/ProxyGeneratorTest.php
```diff
@@ -0,0 +1,57 @@
...
+        $proxyTag->setName('proxy');
+        $this->assertEquals('proxy', $proxyTag->getName());
+    }
+
+    /** @test */
+    public function can_only_auto_refresh_getters(): void
+    {
+        $tag = new Tag();
+        $this->managerRegistry->expects($this->any())->method('getManagerForClass')->with(Tag::class)->willReturn($this->objectManager);
+        $proxyTag = $this->generator->generate($tag, ['get*']);
```
wouterj, line 51: This is my try at "working around" the auto-refresh problems from #113. You can specify which methods should be proxied/use auto-refresh (allowing for `*` globs, like `get*` or `['get*', 'has*, 'is*']`). The idea is that you will be able to configure this in the model factory.

I'm not sure if this is the ideal solution, but I didn't really understand the `spl_object_id()` solution proposed in #113 😇

src/Instantiator.php
```diff
@@ -123,6 +126,16 @@ public function alwaysForceProperties(array $properties = []): self
         return $this;
     }
 
+    /**
+     * Instead of returning an instance of {@see Proxy}, use a generator to create a real proxy for the object.
+     */
+    public function withProxyGenerator(): self
```
wouterj, line 132: My original idea was using this to enable the proxy generator. However, I'm not sure if this should live in `Instantiator` or `Factory`. It's much more common in Foundry to modify `Instantiator` rather than the factory instance. On the other hand, `Factory` is currently creating the `Proxy` and the `ProxyGenerator` needs access to the factory's `Configuration` instance.

tests/Unit/ProxyGeneratorTest.php
kbond: I can't tell if this solves the issue in #113. Might be better to test on an entity with multiple fields like `Post`.

> but I didn't really understand the spl_object_id()

Long story short, I don't think this will work. Basically, I was trying to come up with a way to only refresh the object if it's been changed in the db.

src/Instantiator.php
kbond: Hmm, my first thought is `Factory` as it requires the `Configuration` as you say. Obviously given the static nature of everything it doesn't really matter but I've been trying to follow proper a dependency injection pattern where possible.

tests/Unit/ProxyGeneratorTest.php
wouterj: Basically, my idea was that a user can limit which methods will cause a refresh. E.g. you wouldn't necessarily refresh on setters, which automatically resolves the issue when calling multiple setters.

We can maybe also make this more "magic" by running the fresh post-method invocation and only run it for methods that return something (and then, after a refresh, call the method again on the entity from within the interceptor). However, not sure if there are edge-cases there (e.g. fluent setters).

kbond: I had thought about limiting to non-setters (and of course make configurable) but for models that have rich domains, this becomes a burden I think. I came to the conclusion that it's better to refresh everything and give the ability to disable auto-refreshing for certain circumstances which I believe are rare.

I'm relying on auto-refreshing (on by default) in 100's of tests and only have one instance where I have to disable it.

tests/Functional/ProxyGeneratorTest.php
```diff
@@ -0,0 +1,211 @@
...
+        $this->markTestSkipped('not supported');
+    }
+
+    /**
+     * @test
+     */
+    public function can_force_set_and_save(): void
+    {
+        $this->markTestSkipped('not supported');
+    }
```
wouterj, line 55: I did not move the `forceSet`/`forceGet` methods to the factory (yet). Is this a feature we really want to keep in Foundry?

tests/Functional/ProxyGeneratorTest.php
```diff
@@ -0,0 +1,211 @@
...
+{
+    protected static $POST_FACTORY = PostFactoryWithProxyGenerator::class;
+
+    /**
+     * @test
+     */
+    public function can_assert_persisted(): void
+    {
+        $this->markTestSkipped('not supported');
+    }
```
wouterj, line 23: TODO: we probably have to move the `assertPersisted()` to the assertions class

tests/Functional/ProxyGeneratorTest.php
```diff
@@ -0,0 +1,211 @@
...
+        $this->assertSame('old title', $post->getTitle());
+        $this->assertSame('old body', $post->getBody());
+
+        // reboot kernel
+        static::ensureKernelShutdown();
+        static::bootKernel();
+
+        $this->assertSame('old title', $post->getTitle());
+        $this->assertSame('old body', $post->getBody());
+    }
```
wouterj, line 100: TODO: I'm not sure if it's possible to replace the value holder instance. This is required for re-fetching of entities (e.g. after the entity manager is cleared).

wouterj (commented):
After a very long silence, I've made some progress with this feature today.

Auto refresh is now implemented exactly like in the current `Proxy` object and other helper functions are moved to the factory. The implementation is not yet feature-complete compared to the old `Proxy`. I've left some comments on the missing bits (some of which I'm not sure if it's possible to implement).

wouterj:
Btw, I can't reproduce the Psalm errors locally :confused:

tests/Functional/ProxyGeneratorTest.php
kbond: I do find them useful for feature-rich models with no public access to certain fields (ie verification or password reset token).

I'd be fine with leaving them on the `Instantiator` or adding some helper functions.

tests/Functional/ProxyGeneratorTest.php
kbond: I don't personally find the Proxy assertions useful, maybe we can remove in 2.0?

kbond (commented):
Awesome, thanks for working on this @wouterj! Couple questions/thoughts:
1. Am I correct in that we can't add methods to the generated proxy?
2. Are you planning a way to enable globally via config?

wouterj:
> Am I correct in that we can't add methods to the generated proxy?

Yes. Proxies, at least within the scope of the ProxyManager library, are meant to behave exactly like the real object (with the proxy thing being an implementation detail).
The proxy manager has a couple proxy types: value holder (used here, allows you to call code before and after each real object method call); and lazy loading proxies (those you know from Doctrine, where the entity is only fetched from the db when you call a method).

> Are you planning a way to enable globally via config?

I think it should be doable. I made the opt-in per factory at the moment, as that allows you to migrate one factory at the time. Also, it means that e.g. third party factories can already add support for it while your application's factories don't.

tests/Functional/ProxyGeneratorTest.php
wouterj: That sounds great :)

Honestly, I personally would like to remove as much custom methods from the proxy as possible. I much rather use the normal Doctrine API. But I know you're quite fond of these little helper methods. In short: you're completely at the lead as to which proxy methods to keep and which to drop in 2.0 :)

wouterj:
To summarize our Slack conversation:

* [ ] Move these proxy methods to normal functions: `force_set`, `force_get`, `disable_autorefresh`, `enable_autorefresh`, `without_autorefresh`?
* [ ] Move `save` and `remove` to methods on the `Factory`
* [ ] Handling auto refresh while the kernel was rebooted is a must have

wouterj:
I've updated the PR solving the biggest blocker: refreshing the model (value holder) of the proxy.

I'll try to find some time later this week/weekend to fix the 2 remaining (simple) tasks. After that, this should be ready :)

src/Proxy/ProxyGenerator.php
```diff
@@ -0,0 +1,91 @@
...
+
+                $objectManager->refresh($model);
+
+                return;
+            }
+
+            // refetch the model as it's no longer managed
+            $modelId = $modelClassMetadata->getIdentifierValues($model);
+            if (!$modelId) {
+                // TODO: throw exception?
```
wouterj, line 56: The current `Proxy` class sets the value holder to `null`. Is that what we want? Or should we throw an exception? Or leave the value holder as-is?

kbond: If I understand correctly, currently, if the object cannot be "refreshed", we [throw an exception](https://github.com/zenstruck/foundry/blob/d2cd4c7b4c6821b96d8fda0c1d2febad345ffa35/src/Proxy.php#L167).

kbond:
> Btw, I can't reproduce the Psalm errors locally

https://github.com/laminas/laminas-code/issues/67 looks similar.

src/Proxy/ProxyGenerator.php
wouterj: Oh, I misread the code. Exception it is :)

> By "value holder", do you mean this? 

Yes, in ProxyManager terms, "value holder" is the object that is being proxied.

README.md
```diff
@@ -20,7 +20,7 @@ $post = PostFactory::new() // Create the factory for Post objects
 The factories can be used inside [DoctrineFixturesBundle](https://symfony.com/doc/master/bundles/DoctrineFixturesBundle/index.html)
 to load fixtures or inside your tests, [where it has even more features](#using-in-your-tests).
 
-Want to watch a screencast 🎥 about it? Check out https://symfonycasts.com/foundry
+Want to watch a screencast  about it? Check out https://symfonycasts.com/foundry
```
wouterj, line 23: This should probably be reverted while merging (my terminal has a hard crash when it tries to display emojis currently, making it impossible to edit this file)

wouterj:
PR updated once again with the helper functions and an updated README. As the proxy object now really behaves like the real object, I've removed most of the mentions of Proxy (it's mostly an implementation detail now).

While documenting, I discovered 2 remaining tasks that I've added to the PR description.

README.md
kbond: I'll leave it out. If it causes an issue for you it will for others as well.

kbond:
Awesome, thanks for continuing to work on this!  Have you thought anymore a out how we're going to migrate users to this? It'd be nice if new installs auto used this and existing installs have to choose one way or another (like what we did in #131). The maker should be updated as well.

wouterj:
> Have you thought anymore a out how we're going to migrate users to this?

Currently, you would opt-in for the new proxy manager in each factory:

```php
class PostFactory extends ModelFactory
{
    protected function initialize()
    {
        return $this->withProxyGenerator();

        // or, if you want to limit proxy to only some methods (supporting globs)
        //return $this->withProxyGenerator(['get*']);
    }
}
```

I suggest adding a deprecation notice in `Factory#create()` if auto refresh is `true` and the proxy generator was not enabled (i.e. the old proxies are in use). The `make:factory` maker should indeed be updated to use `->withProxyGenerator()` by default.

Optionally, we can add a global option as you suggested. The advantage would be an easier migration when we remove the `withProxyGenerator()` method (probably in Foundry 3, if you stick to Symfony's BC policy). The disadvantage is that this disallows updating one factory each time (when updating the docs, I realized it's quite some things you have to change). I'm happy to hear if you would like such a global option :)

kbond:
I'm on board 100% with regards to `ModelFactory`'s and giving the ability to upgrade on a factory-by-factory basis. With `AnonymousFactory`'s it will be tricky without a global option. We certainly don't want users to have to call `withProxyGenerator()` for these.

Following Symfony's BC policy, we can't remove the current Proxy system entirely in 2.0?

kbond:
What about hooking into the current `auto_refresh_proxies` config and making it an enum: `new`, `old`, `disable` (need help with better names :thinking:) and make using a boolean deprecated?

wouterj:
> Following Symfony's BC policy, we can't remove the current Proxy system entirely in 2.0?

Introducing a new system is a 3 step progress in Symfony BC:

1. In a *minor version* (`x.y`): Introduce the new system, add an option/method to opt in to the new system, and deprecate not doing the opt-in (i.e. trigger deprecation if `withProxyManager()` isn't called)
2. On the next *major version* (`(x+1).0`): switch the opt-in to opt-out and throw an exception if the new system isn't opt-in (i.e. `withProxyManager()` isn't called). You can safely remove the old system now.
3. On the next *major version* (`(n+2).0`): remove the opt-in.

If you remove the method directly in step (2), you require people to update their code to remove the opt-in. This is also why I think it's very strict and OK for smaller libraries to not do this (the method is a useless bit of code hanging around for the entirety of Foundry 2).


> What about hooking into the current auto_refresh_proxies config and making it an enum: new, old, disable (need help with better names thinking) and make using a boolean deprecated?

Hmm, this is an interesting idea. This would allow you to change the "default" and change it on a factory basis using `withProxyManager(bool $enabled = true)`. This is the best of 2 worlds: an app that is fully migrated wouldn't have `withProxyManager()` anywhere and everyone is free to update each factory individually.

I think adding a new boolean option would make the most sense ("new" doesn't mean much in Foundry 2). What about `auto_refresh: true/false`?

kbond:
> If you remove the method directly in step (2), you require people to update their code to remove the opt-in. This is also why I think it's very strict and OK for smaller libraries to not do this (the method is a useless bit of code hanging around for the entirety of Foundry 2).

Yeah, this does seem a bit strict. I'd like to remove `Proxy` (and related code) completely in 2.0. Do you agree? If so, `withProxyManager` wouldn't exist in 2.0, correct? I'd hate to add this to the maker, then deprecated and remove in 2.0. Or maybe I'm not fully understanding?

> I think adding a new boolean option would make the most sense ("new" doesn't mean much in Foundry 2). What about auto_refresh: true/false?

So what would happen if `auto_refresh: true` and `auto_refresh_proxies: true`? Error?

kbond:
Something that just occurred to me: public properties cannot be auto-refreshed. I don't know how common it is to use public properties on entities. I imagine with static analysis becoming more popular and read-only properties in 8.1 it will become more common.

kbond:
I've been playing with an alternate implementation using Symfony 6.2's `ProxyHelper` (and a bit of hacking). Here's my POC: https://github.com/kbond/symfony-reproducer/tree/foundry-real-proxy 

Here's the relevant [commit](https://github.com/kbond/symfony-reproducer/commit/abf7842c61c8409b0728914efb7b2dab3f45030a) and files:
- https://github.com/kbond/symfony-reproducer/blob/foundry-real-proxy/src/ProxyFactory.php (this is what foundry's Factory could look like)
- https://github.com/kbond/symfony-reproducer/blob/foundry-real-proxy/src/Proxy.php (the generated proxy implements this interface)
- https://github.com/kbond/symfony-reproducer/blob/foundry-real-proxy/src/IsProxy.php (the generated proxy uses this trait which implements the above interface)
- https://github.com/kbond/symfony-reproducer/blob/foundry-real-proxy/tests/RealProxyTest.php (the test that proves this works - everything is fully auto-completable in phpstorm)

Auto-refreshing doesn't work - maybe this is possible with some more tweaks. I'm also ok deprecating auto-refreshing and removing in 2.0. It is complex and has caused some confusion...

/cc @nikophil

wouterj:
> Auto-refreshing doesn't work - maybe this is possible with some more tweaks. I'm also ok deprecating auto-refreshing and removing in 2.0. It is complex and has caused some confusion...

In my honest opinion, if you plan on deprecating auto-refreshing I would advise to completely deprecate the proxy stuff. It's too complex and maintenance heavy to introduce a concept (AR) to entities that Symfony developers aren't really used to use.

I would then propose to just apply the changes of this PR related to moving the AR helpers to `Factory`: https://github.com/zenstruck/foundry/pull/127/files#diff-f1158fbec0bb446d87dbd5346fba683c58ccec1a98b1787511ec7786585e8c31

kbond:
> In my honest opinion, if you plan on deprecating auto-refreshing I would advise to completely deprecate the proxy stuff. It's too complex and maintenance heavy to introduce a concept (AR) to entities that Symfony developers aren't really used to use.

That's a fair point. Although, current users of foundry _are_ used to the AR concept (assuming they're using these methods on the current `Proxy` object).

I'm not sure yet about deprecating auto-refreshing... When it works, it's quite nice.

kbond:
My [POC](https://github.com/kbond/symfony-reproducer/commit/d72cc234acfaef1357dab3994d83e080ab272ba4) now includes auto-refreshing methods (not public properties):

```php
$post->_enableAutoRefresh();

$post->getTitle(); // auto-refreshes before returning value
$post->body; // does not auto-refresh...
```

PR #129, opened 2021-03-04
Fix exception message

jdreesen:
The exception message was: `Uncaught BadMethodCallException: Call to undefined static method "FooFactory::FooFactory::__callStatic".`

Now it is: `Uncaught BadMethodCallException: Call to undefined static method "FooFactory::bar".`


kbond:
Thanks @jdreesen!

PR #131, opened 2021-03-04
Throw exception during auto-refresh if unsaved changes

### LINKED ISSUE #113: Auto-refresh proxy improvements

kbond:
When auto-refreshing a proxy, if there are unsaved changes detected, a helpful exception is thrown. Also, trigger deprecation if not specifically globally configured.

Closes #113.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1133,7 +1133,8 @@ $post->getTitle(); // "New Title" (equivalent to $post->refresh()->getTitle())
 Without auto-refreshing enabled, the above call to `$post->getTitle()` would return "Original Title".
 
 **NOTE**: A situation you need to be aware of when using auto-refresh is that all methods refresh the object first. If
-changing the object's state via multiple methods (or multiple force-sets), the previous changes will be lost:
+changing the object's state via multiple methods (or multiple force-sets), an "unsaved changes" exception will be
+thrown:
 
 ```php
 use App\Factory\PostFactory;
@@ -1144,11 +1145,7 @@ $post = PostFactory::new(['title' => 'Original Title', 'body' => 'Original Body'
 ;
 
 $post->setTitle('New Title');
-$post->setBody('New Body'); // this causes the object to be refreshed, which means the "New Title" title was replaced by the database contents
-$post->save();
-
-$post->getBody(); // "New Body"
-$post->getTitle(); // "Original Title" !! because the subsequent call to ->setBody() "auto-refreshed" the object
+$post->setBody('New Body'); // exception thrown because of "unsaved changes" to $post from above
 ```
 
 To overcome this, you need to first disable auto-refreshing, then re-enable after making/saving the changes:
--- a/README.md
+++ b/README.md
@@ -1184,14 +1184,13 @@ $post->forceSetAll([
 $post->save();
 ```
 
-**NOTE**: You can optionally enable auto-refreshing globally to have every proxy auto-refreshable by default. With this
-enabled, you will have to *opt-out* of auto-refreshing as opposed to the default, which is *opt-in*. Be aware of the
-above situation before enabling.
+**NOTE**: You can enable/disable auto-refreshing globally to have every proxy auto-refreshable by default or not. When
+enabled, you will have to *opt-out* of auto-refreshing.
 
 ```yaml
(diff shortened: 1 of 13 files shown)


src/Proxy.php
```diff
@@ -108,10 +109,27 @@ public function isPersisted(): bool
      */
     public function object(): object
     {
-        if ($this->autoRefresh && $this->persisted) {
-            $this->refresh();
+        if (!$this->autoRefresh || !$this->persisted) {
+            return $this->object;
         }
 
+        // todo: what if object manager isn't entity manager?
```
weaverryan, line 116: Do we support that in Foundry?

kbond: Technically, yes but I've never tested with or even used an object manager other than the orm.

kbond: I think I'll just check that it's an instance of `EntityManagerInterface` in the conditional below.

src/Proxy.php
```diff
@@ -108,10 +109,24 @@ public function isPersisted(): bool
...
 
+        $om = $this->objectManager();
+
+        // only check for changes if the object is managed in the current om
+        if ($om instanceof EntityManagerInterface && $om->contains($this->object)) {
+            // cannot use UOW::recomputeSingleEntityChangeSet() here as it wrongly computes embedded objects as changed
+            $om->getUnitOfWork()->computeChangeSet($om->getClassMetadata($this->class), $this->object);
+
+            if (!empty($om->getUnitOfWork()->getEntityChangeSet($this->object))) {
+                throw new \RuntimeException(\sprintf('Cannot auto refresh "%s" as there are unsaved changes. Be sure to call ->save() or disable auto refreshing (see https://github.com/zenstruck/foundry#auto-refresh for details).', $this->class));
```
kbond, line 124: How's this message?

README.md
```diff
@@ -1580,7 +1579,7 @@ imports:
 zenstruck_foundry:
 
     # Whether to auto-refresh proxies by default (https://github.com/zenstruck/foundry#auto-refresh)
-    auto_refresh_proxies: false
+    auto_refresh_proxies: null # (defaults to false but will default to true in 2.0)
```
kbond, line 1582: How else should we document that this "needs" to be configured?

src/Configuration.php
```diff
@@ -62,6 +62,12 @@ public function instantiator(): callable
 
     public function defaultProxyAutoRefresh(): bool
     {
+        if (null === $this->defaultProxyAutoRefresh) {
+            trigger_deprecation('zenstruck\foundry', '1.9', 'Not explicitly configuring default proxy auto-refresh ("zenstruck_foundry.auto_refresh_proxies" in config or TestState::enableDefaultProxyAutoRefresh()/disableDefaultProxyAutoRefresh()) is deprecated. It will default to true in 2.0.');
```
kbond, line 66: How's this deprecation message?

README.md
wouterj: I would just say `auto_refresh_proxies: true`. The rest is implementation detail to me, and the code/deprecation message do a good enough job to help the user

src/Configuration.php
wouterj: What about:

Not explicitly configuring default proxy auto-refresh is deprecated and will default to true in 2.0. Use "zenstruck_foundry.auto_refresh_proxies" in the bundle config or TestState::enableDefaultProxyAutoRefresh()/disableDefaultProxyAutoRefresh().

PR #143, opened 2021-03-19
Fix false positive for auto-refresh deprecation

### LINKED ISSUE #141: False positive for auto_refresh_proxies deprecation
In a PHPUnit 9.5 test w/ Symfony 5.2, the following edge triggers 

```
Remaining direct deprecation notices (1)

1x: Since zenstruck\foundry 1.9: Not explicitly configuring the default proxy auto-refresh is deprecated and will default to "true" in 2.0. Use "zenstruck_foundry.auto_refresh_proxies" in the bundle config or TestState::enableDefaultProxyAutoRefresh()/disableDefaultProxyAutoRefresh().
```

```php
// class ATest extends TestCase

use Factories;

public function testUnitTest(): void
{
    // ... create a few mocks

    $object = AFactory::new()->withoutPersisting()->create();

    ...
}
```

I've `--force`'d the recipe and have `zenstruck_foundry.auto_refresh_proxies` defined/set in `test` & `dev` configs. This unit test does not user the Symfony Kernel nor interact with the database. 

Declaring `TestState::disableDefaultProxyAutoRefresh();` before xor after calling the factory has no affect on the deprecation.

kbond:
Fixes #141. When using foundry in unit tests (no db), the deprecation was being thrown even though no db is being used.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -62,6 +62,10 @@ final class Configuration
 
     public function defaultProxyAutoRefresh(): bool
     {
+        if (!$this->hasManagerRegistry()) {
+            return false;
+        }
+
         if (null === $this->defaultProxyAutoRefresh) {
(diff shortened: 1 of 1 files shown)


jrushlow:
Tested this patch locally and it solves the issue.

jrushlow (approved):

kbond:
[1.9.1](https://github.com/zenstruck/foundry/releases/tag/v1.9.1) includes this fix.

PR #145, opened 2021-03-23
Add option to use doctrine migrations to reset database

### LINKED ISSUE #144: How to migrate database during test
Hi, I'm trying to use the `ResetDatabase` trait, but it only creates the database tables for the entities that are tracked by Doctrine. However we have many database columns which are not tracked by Doctrine but should be there during a test.

I looked at the source code but it doesn't seem like there is any support for this. Is this intended? Do I need to implement my own solution?

Thanks!

kbond:
This feature gives the ability to have `ResetDatabase` run your migrations during the database reset (instead of `doctrine:schema:drop`/`doctrine:schema:create`). Enable in your `.env.test`: `FOUNDRY_RESET_MODE=migrate`.

When enabled, `ResetDatabase` does the following:
1. Before **first** test using the trait:
    1. drop database (if exists)
    2. create database
    3. run migrations
2. Before **each** test (skips if using DAMADoctrineTestBundle)
    1. drop database (if exists)
    2. create database (I couldn't find an easy way to drop all tables - see #144 for reference)
    3. run migrations

It is highly recommended to use DAMADoctrineTestBundle in conjunction with this feature as if not enabled, the entire database is dropped/created and migrations run before every test.

Todo:
- [x] Support DoctrineMigrationsBundle 2?
- [x] Docs

Fixes #144.

### DIFF
--- a/src/Test/DatabaseResetter.php
+++ b/src/Test/DatabaseResetter.php
@@ -34,11 +34,6 @@ final class DatabaseResetter
         return \class_exists(StaticDriver::class) && StaticDriver::isKeepStaticConnections();
     }
 
-    public static function isResetUsingMigrations(): bool
-    {
-        return 'migrate' === ($_SERVER['FOUNDRY_RESET_MODE'] ?? 'schema');
-    }
-
     public static function resetDatabase(KernelInterface $kernel): void
     {
         $application = self::createApplication($kernel);
@@ -57,6 +52,11 @@ final class DatabaseResetter
 
         self::dropSchema($application, $registry);
         self::createSchema($application, $registry);
+    }
+
+    private static function isResetUsingMigrations(): bool
+    {
+        return 'migrate' === ($_SERVER['FOUNDRY_RESET_MODE'] ?? 'schema');
     }
 
     private static function dropAndCreateDatabase(Application $application, ManagerRegistry $registry): void
--- a/src/Test/DatabaseResetter.php
+++ b/src/Test/DatabaseResetter.php
@@ -36,7 +36,7 @@ final class DatabaseResetter
 
     public static function isResetUsingMigrations(): bool
     {
-        return 'migrate' === ($_SERVER['FOUNDRY_RESET_MODE'] ?? null);
+        return 'migrate' === ($_SERVER['FOUNDRY_RESET_MODE'] ?? 'schema');
     }
 
     public static function resetDatabase(KernelInterface $kernel): void
--- a/src/Test/DatabaseResetter.php
+++ b/src/Test/DatabaseResetter.php
@@ -34,26 +34,17 @@ final class DatabaseResetter
         return \class_exists(StaticDriver::class) && StaticDriver::isKeepStaticConnections();
     }
 
+    public static function isResetUsingMigrations(): bool
+    {
+        return 'migrate' === ($_SERVER['FOUNDRY_RESET_MODE'] ?? null);
+    }
+
     public static function resetDatabase(KernelInterface $kernel): void
     {
         $application = self::createApplication($kernel);
         $registry = $kernel->getContainer()->get('doctrine');
 
-        foreach (self::connectionsToReset($registry) as $connection) {
-            $dropParams = ['--connection' => $connection, '--force' => true];
-
-            if ('sqlite' !== $registry->getConnection($connection)->getDatabasePlatform()->getName()) {
-                // sqlite does not support "--if-exists" (ref: https://github.com/doctrine/dbal/pull/2402)
-                $dropParams['--if-exists'] = true;
-            }
-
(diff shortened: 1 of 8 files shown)


kbond:
I've tested this on an app with a large test suite (+1400 tests). It works fine and the difference in test time was negligible compared to the default mode. In this app, I recently "rolled-up" several years worth of migrations (there are only 4 now) so as your migration count increases, the test time will also increase.

kbond:
I think this feature is ready. @weaverryan, @wouterj, mind giving it a high level review?

mpiot:
@kbond Just try your PR on my project and it work well. I need transition because one of them create a collation: `CREATE COLLATION fr_natural (provider = "icu", locale = "fr-u-kn-true");`.

With your PR, it works well.

kbond:
Great to hear @mpiot, thanks for testing!

mpiot:
Hi @kbond , just test on company project (2k tests, using paratest with WrapperRunner option):
 - Schema Strategy: 2:33, 2:02, 2:05, 2:22 (with some fails due to the lack of migrations)
 - Migration Strategy: 2:11, 2:22, 2:21, 2:17
 
 You're welcome, and thank you for this bundle :)

kbond:
Was this with or without dama?

So averaged out, they seem mostly equivalent performance-wise. That was my finding as well but I only have a few migrations. I imagine the more migrations you have, the slower it would be (unless using dama).

mpiot:
@kbond This is with DAMA yes, in the project the schema:update take about 7 seconds, the d:m:m take the same times, maybe 2-3 seconds more, it's similar (about 150 migration files).

src/Test/DatabaseResetter.php
```diff
@@ -34,11 +34,33 @@ public static function isDAMADoctrineTestBundleEnabled(): bool
         return \class_exists(StaticDriver::class) && StaticDriver::isKeepStaticConnections();
     }
 
+    public static function isResetUsingMigrations(): bool
+    {
+        return 'migrate' === ($_SERVER['FOUNDRY_RESET_MODE'] ?? null);
```
wouterj, line 39:
```suggestion
        return 'migrate' === ($_SERVER['FOUNDRY_RESET_MODE'] ?? 'schema');
```

Maybe to be more explicit on what the default behavior is?

src/Test/DatabaseResetter.php
```diff
@@ -34,11 +34,33 @@ public static function isDAMADoctrineTestBundleEnabled(): bool
         return \class_exists(StaticDriver::class) && StaticDriver::isKeepStaticConnections();
     }
 
+    public static function isResetUsingMigrations(): bool
```
wouterj, line 37:
```suggestion
    private static function isResetUsingMigrations(): bool
```

I don't think there is a reason to use this method outside this class (and makes it easier to modify this in future)

src/Test/DatabaseResetter.php
```diff
@@ -53,27 +75,18 @@ public static function resetDatabase(KernelInterface $kernel): void
...
+            self::runCommand($application, 'doctrine:migrations:migrate', ['-n' => true]);
+        } else {
+            foreach (self::objectManagersToReset($registry) as $manager) {
+                self::runCommand($application, 'doctrine:schema:create', [
+                    '--em' => $manager,
+                ]);
+            }
         }
 
         if (!Factory::isBooted()) {
```
wouterj, line 92: I guess this code has to be moved to `resetDatabase()`? It strictly has nothing to do with what the method name is suggesting (`createSchema`)

tests/Fixtures/Kernel.php
```diff
@@ -36,6 +37,10 @@ public function registerBundles(): iterable
         if (\getenv('USE_DAMA_DOCTRINE_TEST_BUNDLE')) {
             yield new DAMADoctrineTestBundle();
         }
+
+        if ('migrate' === \getenv('FOUNDRY_RESET_MODE')) {
```
wouterj, line 41: I believe it's more safe to rely on `$_SERVER` or `$_ENV` (https://github.com/laravel/framework/pull/7425)

wouterj (commented):
I cannot easily test this in our application, so only did some static review on this one :)

src/Test/DatabaseResetter.php
kbond: This is part of the `TestState::flushGlobalState();` below (which requires foundry to be booted). I think it should be kept together (but this is all changing in #153 so I'll re-evaluate there.)

tests/Fixtures/Kernel.php
kbond: I'll address in a separate PR (created issue #154 to track).

kbond:
Perfect, thanks @wouterj!

PR #151, opened 2021-04-13
[feature] Create ChainManagerRegistry in order to use doctrine/odm

nikophil:


src/Bundle/DependencyInjection/ChainManagerRegistryPass.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Zenstruck\Foundry\Bundle\DependencyInjection;
+
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+use Zenstruck\Foundry\ChainManagerRegistry;
+
+class ChainManagerRegistryPass implements CompilerPassInterface
```
kbond, line 12: Can you mark as final

src/Bundle/DependencyInjection/ChainManagerRegistryPass.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    public function process(ContainerBuilder $container)
```
kbond, line 14: `: void` return

src/Bundle/DependencyInjection/ZenstruckFoundryExtension.php
```diff
@@ -5,7 +5,9 @@
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
+use Symfony\Component\DependencyInjection\Reference;
```
kbond, line 8: no longer needed.

src/ChainManagerRegistry.php
```diff
@@ -0,0 +1,101 @@
+<?php
+
+namespace Zenstruck\Foundry;
+
+use Doctrine\Persistence\ManagerRegistry;
+use Doctrine\Persistence\ObjectManager;
+use Doctrine\Persistence\ObjectRepository;
+
+class ChainManagerRegistry implements ManagerRegistry
```
kbond, line 9: final and `@internal`.

src/ChainManagerRegistry.php
```diff
@@ -0,0 +1,101 @@
...
+    public function __construct(array $managerRegistries)
+    {
+        if (count($managerRegistries) === 0) {
+            throw new \InvalidArgumentException('no manager registry provided');
+        }
+
+        $this->managerRegistries = $managerRegistries;
+    }
+
+    public function getRepository($class, $persistentManagerName = null): ObjectRepository
```
kbond, line 24: I guess using `$persistentObject` instead if `$class` will fix.

src/Test/Factories.php
```diff
@@ -29,13 +30,15 @@ public static function _setUpFactories(): void
 
         TestState::bootFromContainer($kernel->getContainer());
         Factory::configuration()->setManagerRegistry(
-            new LazyManagerRegistry(static function() {
-                if (!static::$booted) {
-                    static::bootKernel();
-                }
+            new LazyManagerRegistry(
```
kbond, line 33: cs issues here

src/Test/Factories.php
```diff
@@ -29,13 +30,15 @@ public static function _setUpFactories(): void
...
+                static function () {
+                    if (!static::$booted) {
+                        static::bootKernel();
+                    }
 
-                return static::$kernel->getContainer()->get('doctrine');
-            })
+                    return static::$kernel->getContainer()->get(ChainManagerRegistry::class);
```
kbond, line 39: I think we're going to have a problem here with non-bundle usage: `ChainManagerRegistry` isn't in the container. I think we'll have to do what we do in the compiler pass manually here - maybe as a helper method on TestState to avoid duplication?

src/Test/TestState.php
```diff
@@ -122,7 +123,7 @@ public static function bootFromContainer(ContainerInterface $container): void
         $configuration = new Configuration();
 
         try {
-            $configuration->setManagerRegistry($container->get('doctrine'));
+            $configuration->setManagerRegistry($container->get(ChainManagerRegistry::class));
```
kbond, line 126: This logic is called in bundless-usage - as mentioned above, we'll have to do what the compiler pass does manually.

kbond (changes requested):
Looks good, some minor cleanup and some considerations for *bundle-less* usage.

src/Test/TestState.php
```diff
@@ -122,14 +124,34 @@ public static function bootFromContainer(ContainerInterface $container): void
...
-            $configuration->setManagerRegistry($container->get('doctrine'));
+            $configuration->setManagerRegistry(self::initializeChainManagerRegistry($container));
         } catch (NotFoundExceptionInterface $e) {
             throw new \LogicException('Could not boot Foundry, is the DoctrineBundle installed/configured?', 0, $e);
         }
 
         self::bootFoundry($configuration);
     }
 
+    public static function initializeChainManagerRegistry(ContainerInterface $container): ChainManagerRegistry
```
kbond, line 135: Can you mark `@internal`.

kbond (changes requested):
Awesome, looks great! Just one minor change.

What are you thinking for actually testing odm factories? Will be hard to do without the reset functionality. My only concern with merging this PR is we don't truly know that it works (in the context of this library's test suite). Have you tested this PR on your project?

kbond:
Also, don't worry about the test coverage drop - I only use this as a metric, not a strict requirement.

nikophil:
@kbond as I told you on slack, indeed I'll need to work on the reset functionality before being able to functionally test the PR.

I haven't tested the PR on my project yet. I think we really need to test everything before merging in master...
Maybe we can say this PR is OK, an I'll work on the reset / functional test stuff in another PR which will be based on this one? Then we'll merge in master only when we'll be ready.

For the testing strategy, I haven't thought about it really strongly but I feel like we can reuse a lot of existing tests. 
I do need to create some fixtures documents (ie: entities for ODM), and maybe I'll be able to play with an env var and data providers to decide what to test. 
IMO, only functional tests that interacts with database need to be tested towards ODM

FYI there is no relationship between documents in ODM, so it won't be necessary to test many-to-one, etc...

nikophil:
> I haven't tested the PR on my project yet. 

but I surely do before merging anything :wink:

kbond:
> Maybe we can say this PR is OK, an I'll work on the reset / functional test stuff in another PR which will be based on this one? Then we'll merge in master only when we'll be ready.

That would be great!

> IMO, only functional tests that interacts with database need to be tested towards ODM

Yes, I'd only expect a very minimum amount of tests - just as a sanity check for now - can always add more later as/if we find edge cases etc.

kbond:
> FYI there is no relationship between documents in ODM, so it won't be necessary to test many-to-one, etc...

What about "embedded documents" - will this be supported just "by default"?

nikophil:
`doctrine/odm` provides two annotations `EmbedMany` and `EmbedOne`.
They are helpers to create embedded documents inside main one:
```json
{
    "title": "My blog post",
    "body": "My loooong post body",
    "author": {
        "name": "nikophil"
    },
    "comments": [
        {"body": "comment 1"},
        {"body": "comment 2"},
    ]
}
```

but there is no external reference to them outside from main document. They can be easily created in user with the factories, and they will simply be persisted.

Maybe some helpers method could be created for them, I don't know how it could be useful, but it's definitely not necesarry

kbond:
Perfect, that's what I expected.

You're example above seems like the perfect initial fixture set for use in the test suite.

kbond (approved):

PR #152, opened 2021-04-16
[feature] Handle schema reset for mongo

nikophil:
Hello,

here is a proposition to support the `DatabaseResetter` with mongo.

I've ran into few problems for which I've made some trade-off, see my comments in the code.
I'm not quite happy with the implementation because `self::isDAMADoctrineTestBundleEnabled()` pops in every layers (`ResetDatabase` trait, `DatabaseResetter` and `ORMDatabaseResetter`) althought it is only really needed for orm...

I've not created automated tests for new classes yet: I'd rather wait we're OK on an implementation (I don"t even know if it's necessary since these class will be used into the test suite). 
I've also deactivated codecoverage because it resulted in too much noise in code review

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,7 @@ jobs:
   code-coverage:
     name: Code Coverage
     runs-on: ubuntu-latest
+    if: false
     services:
       mysql:
         image: mysql:5.7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,10 @@ jobs:
         ports:
           - 5432:5432
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+      mongo:
+        image: mongo:4
+        ports:
+          - 27017:27017
     strategy:
       matrix:
         php: [7.2, 7.3, 7.4, 8.0]
@@ -35,9 +39,9 @@ jobs:
           - php: 7.2
             stability: '@stable'
             versions: lowest
-          - php: 8.0
-            stability: '@dev'
-            versions: highest
+#          - php: 8.0
+#            stability: '@dev'
+#            versions: highest
     steps:
       - name: Checkout code
         uses: actions/checkout@v2.3.3
@@ -49,7 +53,7 @@ jobs:
         uses: shivammathur/setup-php@2.7.0
         with:
           php-version: ${{ matrix.php }}
-          extensions: pgsql, sqlite
+          extensions: pgsql, sqlite, mongodb
           coverage: none
 
       - name: Set minimum-stability to dev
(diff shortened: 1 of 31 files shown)


src/ChainManagerRegistry.php
```diff
@@ -51,7 +51,7 @@ public function getManagers(): array
         return \array_reduce(
             $this->managerRegistries,
             static function(array $carry, ManagerRegistry $managerRegistry) {
-                return \array_merge($carry, $managerRegistry->getManagers());
+                return \array_merge($carry, \array_values($managerRegistry->getManagers()));
```
nikophil, line 54: `ManagerRegistry::getManagers()` actually return `array<string, ObjectManager>`, where the key is the objectManager's name. So there is a risk of overlay.

This actually makes our method return `list<ObjectManager>` but the interface says `@return ObjectManager[]` so we don't break the contract...

src/Test/DatabaseResetter.php
```diff
@@ -37,72 +34,52 @@ public static function isDAMADoctrineTestBundleEnabled(): bool
...
     {
-        $application = self::createApplication($kernel);
-        $registry = $kernel->getContainer()->get('doctrine');
+        foreach (self::schemaResetters($kernel) as $databaseResetter) {
+            $databaseResetter->resetSchema();
+        }
+
+        if (self::isDAMADoctrineTestBundleEnabled()) {
+            return;
+        }
```
nikophil, line 54: This is where we have a problem with odm & orm & dama: if we have a global state for odm, it wouldn't be persisted.

src/Test/DatabaseResetter.php
```diff
@@ -37,72 +34,52 @@ public static function isDAMADoctrineTestBundleEnabled(): bool
     public static function resetDatabase(KernelInterface $kernel): void
     {
         $application = self::createApplication($kernel);
-        $registry = $kernel->getContainer()->get('doctrine');
+        $databaseResetter = new ORMDatabaseResetter($application, $kernel->getContainer()->get('doctrine'));
```
nikophil, line 37: There is no such things like `doctrine:database:drop` or `doctrine:database:create` in `doctrine/odm`. 
That's why we only use ORM here, and why the interface I've created only refers to "reset schema"

src/Test/ODMSchemaResetter.php
```diff
@@ -0,0 +1,52 @@
...
+                    '--dm' => $manager,
+                ]
+            );
+        }
+    }
+
+    /** @return list<string> */
+    private function objectManagersToReset(): array
+    {
+        return [$this->registry->getDefaultManagerName()];
```
nikophil, line 50: for now, we only deal with `default` manager, but we can imagine to introduce new `FOUNDRY_RESET_ODM_MANAGERS` env var

src/Test/ResetDatabase.php
```diff
@@ -50,11 +50,6 @@ public static function _resetDatabase(): void
      */
     public static function _resetSchema(): void
     {
-        if (DatabaseResetter::isDAMADoctrineTestBundleEnabled()) {
-            // not required as the DAMADoctrineTestBundle wraps each test in a transaction
-            return;
-        }
```
nikophil, line 56: Now even with dama, the kernel will be booted: we need to know whether or not we have odm in the kernel, this condition has been moved into `ORMDatabaseResetter::resetSchame()`

src/Test/SchemaResetterInterface.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Zenstruck\Foundry\Test;
+
+/**
+ * @internal
+ */
+interface SchemaResetterInterface
```
kbond, line 8: Since this is an internal interface, let's just make it an abstract class (`SchemaResetter`) that includes the helper method from the `CommandRunner` trait and remove the trait.

kbond (changes requested):
Nice work! I did a quick review and I think there are some opportunities to cleanup/refactor but I'd rather wait until we have odm tests. This will allow us to refactor with confidence.

Once we have new, passing tests, I'll merge into the primary mongo-support branch and we can iterate on it with additional PRs.

kbond:
> I'm not quite happy with the implementation because self::isDAMADoctrineTestBundleEnabled() pops in every layers

Fine for now I think, we may be able to refactor away on a followup iteration.

.github/workflows/ci.yml
```diff
@@ -128,9 +132,39 @@ jobs:
 #          USE_FOUNDRY_BUNDLE: 1
           DATABASE_URL: sqlite:///%kernel.cache_dir%/app.db
 
+      - name: 'Test: MySQL & Mongo'
```
nikophil, line 135: maybe it's time to use a matrix for all test cases?

kbond: Let's save for a future PR.

kbond (changes requested):
Looks good, couple comments/suggestions:
1. Can we just check `DATABASE_URL`/`MONGO_URL` to determine if they are enabled? (and remove `USE_ORM`/`USE_ODM`).
2. Can you disable psalm for now (and remove any psalm related comments added). We'll do a separate PR after to re-enable and fix.

kbond (approved):
LGTM, when you're ready, unmark as draft and I'll merge into the `mongo-support` branch.

nikophil:
hi @kbond 
I think the PR is ready

PR #155, opened 2021-04-21
Fix example $posts for Attributes section

babeuloula:
Hello,

Just a little fix on the documentation.


kbond (approved):
Thanks @babeuloula!

PR #157, opened 2021-05-02
[feature] Support embedded objects for ODM

nikophil:


src/Factory.php
```diff
@@ -52,6 +53,10 @@ public function __construct(string $class, $defaultAttributes = [])
 
         $this->class = $class;
         $this->attributeSet[] = $defaultAttributes;
+
+        if ($this->classIsEmbeddedObject()) {
+            $this->persist = false;
+        }
```
nikophil, line 59: should be in `Factory` constructor in order to be applied to `AnonymousClass` as well

kbond: Can we not move this to `Factory::isPersisting()`?

kbond (commented):
Thanks for working on this! Just the one comment/question and failing tests. Factory can be booted without doctrine so you'll have to do a `self::configuration()->hasManagerRegistry()` check.

PR #158, opened 2021-05-05
Update .gitattributes file

ker0x:
These folders and files are not needed when fetching the package through Composer.


kbond:
Thank you @ker0x.

PR #162, opened 2021-05-15
Only list entities which do not have factories

### LINKED ISSUE #82: make:factory only list entities that don't have factories already
Related to #15

jschaedl:
Fixes #82

### DIFF
--- a/src/Bundle/Resources/skeleton/Factory.tpl.php
+++ b/src/Bundle/Resources/skeleton/Factory.tpl.php
@@ -50,7 +50,7 @@ final class <?= $class_name ?> extends ModelFactory
         ;
     }
 
-    public static function getClass(): string
+    protected static function getClass(): string
     {
         return <?= $entity->getShortName() ?>::class;
     }
--- a/src/Bundle/Resources/skeleton/Factory.tpl.php
+++ b/src/Bundle/Resources/skeleton/Factory.tpl.php
@@ -50,7 +50,7 @@ final class <?= $class_name ?> extends ModelFactory
         ;
     }
 
-    protected static function getClass(): string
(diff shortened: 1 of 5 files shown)


src/Bundle/Resources/skeleton/Factory.tpl.php
```diff
@@ -50,7 +50,7 @@ protected function initialize(): self
         ;
     }
 
-    protected static function getClass(): string
+    public static function getClass(): string
```
jschaedl, line 53: Not happy about that change. Maybe you have a better idea how we can access the entities class name?

src/ModelFactory.php
```diff
@@ -240,7 +240,7 @@ final protected function addState($attributes = []): self
     }
 
     /** @psalm-return class-string<TModel> */
-    abstract protected static function getClass(): string;
+    abstract public static function getClass(): string;
```
jschaedl, line 243: Not happy about that change. Maybe you have a better idea how to access the entities class names?

kbond: This is also a BC break.

It's kind of crappy but the only thing I can think of is adding a separate public static method that just calls the protected one. Marked as final and `@internal`.

kbond (changes requested):
Let's just get the CI passing and I'll merge!

jschaedl:
I still need to fix the tests.

kbond:
I fixed the tests and added an assertion to ensure existing factory entities aren't displayed.

Thanks @jschaedl!

kbond (approved):

wouterj:
Wohoo, this is great. Thanks @jschaedl!

Issue #163, opened 2021-05-28
Define different foldername in maker-command


DigitalTimK:
Since I have a "Factory"-folder already in use for other reasons, the standard-foldername of Foundry `Factory` is causing an issue.

**Example**
- Entity-class: `Entity/Person.php`
- Non-Foundry-Factory in use: `src/Factory/PersonFactory.php`

**Issue**
The maker-command (`./bin/console make:factory`) will not create a new FOUNDRY-Factory file, as the same filename is already exsting. So the only possiblity would be, to refactor my entire code from "Factory" to "FactoryOther" (and their corresponding namespaces)

**Question**
Would it be possible to have either the possibility:
- to adapt the target-folder in maker-command, or
- to define the target-folder in YAML-file, or
- to adapt the make-comand so it will create a file like `Factory/PersonFactory_2.php`, if file already exists (to be able to adapt this fresh file) 

Cheers!

kbond:
Hey @DigitalTimK, for now, I'm going to stick with [this decision](https://github.com/zenstruck/foundry/issues/89#issuecomment-753328558). Basically, these makers may eventually be migrated to the official maker bundle and they've decided not to allow this.

DigitalTimK:
ok, sounds not so good. What do you think about proposed solution "bullet 3"?

kbond:
Ok, I think I'm ok with a `--namespace` option, what do you think about:
```
make:factory Post --namespace=Custom\\Factory\\Folder => src/Custom/Factory/Folder/PostFactory.php
```

DigitalTimK:
that would would be wonderful ;-)

kbond:
This feature is added in #164.

I had a chat with @weaverryan who maintains the official maker-bundle. Some path customization have been permitted there so I decided not to take such a "hard line" here.

PR #164, opened 2021-05-30
Customize namespace for factories generated with make:factory

kbond:
_Closes #163, Closes #89_


kbond:
@wouterj how do you think I should handle the psalm error? Renaming `$criteria` to `$id` fixes - should I just do that?

PR #167, opened 2021-06-07
[tests] fixed how tests are skipped

nikophil:
When a whole test class is skipped with `setUpBeforeClass()`, the first test case inside this class is skipped with the good messages, and all other test cases are skipped with the following error displayed: `Test skipped because of an error in hook method`


kbond:
Great! Just the cs issue, don't worry about psalm, it's fixed in the main branch, it'll get picked up when we rebase the mongo-support branch. FYI, the rebase will be sticky because of #145 but we'll worry about that later...

This PR includes 58c6e8a or will that be in another PR?

PR #168, opened 2021-06-07
WIP: fix problem on global state with mongo & dama

nikophil:


nikophil:
this won't be made since it is an edge case and needs a lot of hacky stuff to work

PR #173, opened 2021-06-25
[feature] Auto populate ModelFactory::getDefaults() based on doctrine mapping

benblub:
For Doctrine Types generate with faker Factory defaults for required fields. Types which cannot be handled will be created with null and a TODO as comment.

### Supported Doctrine Types
```php
    const DOCTRINE_TYPES = [
        'ARRAY' => '[],',
        'ASCII_STRING' => 'self::faker()->text()',
        'BIGINT' => 'self::faker()->randomNumber()',
        'BLOB' => 'self::faker()->text()',
        'BOOLEAN' => 'self::faker()->boolean()',
        'DATE' => 'self::faker()->datetime()',
        'DATE_MUTABLE' => 'self::faker()->datetime()',
        'DATE_IMMUTABLE' => 'self::faker()->datetime()',
        'DATETIME_MUTABLE' => 'self::faker()->datetime()',
        'DATETIME_IMMUTABLE' => 'self::faker()->datetime()',
        'DATETIMETZ_MUTABLE' => 'self::faker()->datetime()',
        'DATETIMETZ_IMMUTABLE' => 'self::faker()->datetime()',
        'DECIMAL' => 'self::faker()->randomFloat()',
        'FLOAT' => 'self::faker()->randomFloat()',
        'INTEGER' => 'self::faker()->randomNumber()',
        'JSON' => '[]',
        'JSON_ARRAY' => '[]',
        'SIMPLE_ARRAY' => '[]',
        'SMALLINT' => 'self::faker()->randomNumber(1, 32767)',
        'STRING' => 'self::faker()->text()',
        'TEXT' => 'self::faker()->text()',
        'TIME_MUTABLE' => 'self::faker()->datetime()',
        'TIME_IMMUTABLE' => 'self::faker()->datetime()',
    ];
```


benblub:
convert to draft i missed that 🙄

benblub:
Im thinking about Symfony Constraints/Validator. The Validator Component is no depency in Foundry. I can remove it or we add the component as depency?

Some examble, i have an field named fooMail and have a Email Constraint for it. Our string list is not smart enough to check its an Email and if there is created a sentence for it, validation/tests would fail. Thats why i added Support for read Constraint and use them for generating defaults.

Currently the Extractor just use this both. There are some more..
use Symfony\Component\Validator\Constraints\Email;
use Symfony\Component\Validator\Constraints\Url;

kbond:
Hey @benblub, thanks for working on this. Some quick notes:

1. Your generated code example above seems to show the output of faker as the defaults. It should be the faker call itself to ensure all created factories are different:
    ```php
    {
        return [
            'title' => self::faker()->title(),
            'company' => self::faker()->company(),
            // etc...
        ];
    }
    ```
2. I think, at least as a start, I'd like to be super dumb and only look at the doctrine metadata to determine the faker formatter to use (and only use if the property is non-nullable). There is going to be some complexity with nested factories and I'd like to keep the entire system simple until this is solid.

Here's what I'm thinking with regards to [2] above:

| Doctrine Type | Default |
| --- | --- |
| ARRAY | `[]` (can't know what kind of array user would want) |
| ASCII_STRING | `self::faker()->text()` |
| BIGINT | `self::faker()->randomNumber()` |
| BINARY | add todo comment |
| BLOB | `self::faker()->text()` |
| BOOLEAN | `self::faker()->boolean()` |
| DATE_MUTABLE | `self::faker()->datetime()` |
| DATE_IMMUTABLE | `self::faker()->datetime()` |
| DATEINTERVAL | add todo comment |
| DATETIME_MUTABLE | `self::faker()->datetime()` |
| DATETIME_IMMUTABLE | `self::faker()->datetime()` |
| DATETIMETZ_MUTABLE | `self::faker()->datetime()` |
| DATETIMETZ_IMMUTABLE | `self::faker()->datetime()` |
| DECIMAL | `self::faker()->randomFloat()` |
| FLOAT | `self::faker()->randomFloat()` |
| GUID | add todo comment |
| INTEGER | `self::faker()->randomNumber()` |
| JSON | `[]` (can't know what kind of array user would want) |
| JSON_ARRAY | `[]` (can't know what kind of array user would want) |
| OBJECT | add todo comment |
| SIMPLE_ARRAY | `[]` (can't know what kind of array user would want) |
| SMALLINT | `self::faker()->randomNumber()` (within range SMALLINT allows) |
| STRING | `self::faker()->text()` |
| TEXT | `self::faker()->text()` |
| TIME_MUTABLE | `self::faker()->datetime()` |
| TIME_IMMUTABLE | `self::faker()->datetime()` |

I'm really a fan of keeping things as simple as possible - in this case, adding _just enough_ to allow the entity to be saved without error.

src/Bundle/Extractor/DoctrineTypes.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Zenstruck\Foundry\Bundle\Extractor;
+
+class DoctrineTypes
+{
+    public const DOCTRINE_TYPES = [
+        'ARRAY' => '[]',
+        'ASCII_STRING' => 'self::faker()->text()',
+        'BIGINT' => 'self::faker()->randomNumber()',
+        'BINARY' => '// @TODO add Binary manualy',
```
kbond, line 11: Actually, let's remove the unsupported types from this array and in `Property::createFakerMethodFromDoctrineType()`, when the type is not found, generate a todo.

src/Bundle/Extractor/Property.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public function createFakerMethodFromDoctrineType(string $doctrineType)
+    {
+        $doctrineType = \mb_strtoupper($doctrineType);
+
+        if (\array_key_exists($doctrineType, DoctrineTypes::DOCTRINE_TYPES)) {
+            return DoctrineTypes::DOCTRINE_TYPES[$doctrineType];
+        }
+
+        throw new Exception('DOCTRINE_TYPE not found: '.$doctrineType);
```
kbond, line 72: We can't throw an exception here as it is possible/likely, a user could have their own doctrine types defined. We don't want to prevent the factory from being created.

Instead, let's do `return "// @TODO add {$doctrineType} manually";`

src/Bundle/Resources/skeleton/Factory.tpl.php
```diff
@@ -39,6 +39,11 @@ protected function getDefaults(): array
     {
         return [
             // TODO add your default values here (https://github.com/zenstruck/foundry#model-factories)
+<?php
+foreach ($defaultProperties as $fieldname => $type) {
+        echo "            '".$fieldname."' => ".$type.",\n";
```
kbond, line 44: As it stands now, the todo comments will generate invalid php syntax:

```php
return [
    'field1' => self::faker()->text(),
    'field2' => // TODO... (missing a value and comma)
    'field3' => self::faker()->text(),
];
```

I'm thinking the value returned from `Property::createFakerMethodFromDoctrineType()` should include the comma and the todo value should be `null` (?). The above would look like this:

```php
return [
    'field1' => self::faker()->text(),
    'field2' => null, // TODO...
    'field3' => self::faker()->text(),
];
```

src/Bundle/Extractor/Property.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    private $em;
+
+    public function __construct(EntityManagerInterface $em)
+    {
+        $this->em = $em;
+    }
+
+    /**
+     * @TODO Support for defaults is missing
```
kbond, line 32: What are you meaning by "defaults" here?

kbond (changes requested):
Looking good!

src/Bundle/Extractor/Property.php
benblub: i removed defaults. This was from early step where i forgot thats a default will be always available on instantiates classes. Defaults meaning: `$foo = 'bar` as examble.

benblub:
i take later a view into the issues from phpunit which not happen on my dev machine. Would be great if the CI checks starts automatic after a push :)

kbond:
> Would be great if the CI checks starts automatic after a push :)

Ugh, I know. GH added this to help stem bitcoin mining abuse... Annoying I have to "allow" after every commit you do.. You'd think the first time would be enough. I _think_ once I merge this, any subsequent PRs by you won't have this restriction (but I'm not positive).

tests/Functional/Bundle/Extractor/PropertyTest.php
```diff
@@ -0,0 +1,75 @@
...
+{
+    /**
+     * @var Property|object|null
+     */
+    private $property;
+
+    protected function setUp(): void
+    {
+        self::bootKernel();
+        $em = static::getContainer()->get('doctrine.orm.default_entity_manager');
```
kbond, line 19: You need to use `static::$container` here. `getContainer` is only available in Symfony 5.3.

tests/Functional/Bundle/Extractor/PropertyTest.php
```diff
@@ -0,0 +1,75 @@
...
+        $this->property = new Property($em);
+    }
+
+    /**
+     * @before
+     */
+    public function skipIfNotUsingFoundryBundle(): void
```
kbond, line 26: I don't _think_ this is required.

kbond (changes requested):
Just a minor thing to get the tests passing (I think).

benblub:
mhh im not sure why this error happen: Error: Call to undefined method Zenstruck\Foundry\Tests\Functional\Bundle\Extractor\PropertyTest::getContainer()

```php
        if (!\getenv('KERNEL_CLASS')) {
            $this->markTestSkipped('App Kernel not found.');
        }
```

Kernel ENV is there but it looks symfony Frameworkbundle is not installed on this test 🤔

tests/Functional/Bundle/Extractor/PropertyTest.php
benblub: ahh ok ... dang got it

tests/Functional/Bundle/Extractor/PropertyTest.php
benblub: ok remove lets try it will pass now. fingers crossed

src/Bundle/Extractor/Property.php
```diff
@@ -0,0 +1,79 @@
...
+            return DoctrineTypes::DOCTRINE_TYPES[$doctrineType];
+        }
+
+        return 'null, // @TODO add '.$doctrineType.' manually';
+    }
+
+    /**
+     * @TODO get relations which cant be NULL
+     */
+    public function getPropertiesFromDoctrineRelations()
```
kbond, line 66: What do you think about, for this PR, just adding the code to detect non-nullable relations and adding a todo to the user's factory (just like we do with unsupported types)?

src/Bundle/Extractor/Property.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+    }
+
+    public function getProperties(): array
```
kbond, line 70: This is unused and can be removed.

src/Bundle/Extractor/Property.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        return $this->properties;
+    }
+
+    public function setProperties(array $properties): void
```
kbond, line 75: This is unused and can be removed.

src/Bundle/Extractor/Property.php
```diff
@@ -0,0 +1,79 @@
...
+use Doctrine\ORM\EntityManagerInterface;
+use Exception;
+use ReflectionClass;
+
+class Property
+{
+    /**
+     * @var array
+     */
+    private $properties = [];
```
kbond, line 14: Let's remove this property.

kbond (changes requested):

src/Bundle/Extractor/Property.php
benblub: if we merge this PR first we can test next CI checks run automatic? And yes add todo for not nullable relations is a good next todo. Also some naming refactoring.

kbond: > if we merge this PR first we can test next CI checks run automatic

I think so, but I'm not positive. Want me to merge as is and you can add in a follow-up PR?

benblub: yes would be great 👍

kbond:
Before I merge, do you mind updating the PR description to reflect the current state? I link to these in my release's changelog so I'd like users to be able to get a proper overview of the feature.

kbond (approved):

kbond:
Thanks @benblub! Hopefully your future PR's actions will run automatically now.

benblub:
i hope so. We will see it soon.

Issue #179, opened 2021-07-27
Entity cascading unexpected behavior


mpiot:
In the case we have 3 entities linked together likes:

```php
class A
{
  /**
   * @ORM\OneToMany(targetEntity=B::class, mappedBy="a", cascade={"persist"})
   */
  private Collection $bs;

  private ?string $reference = null;

  public function __construct() {
    ​$this->bs = new ArrayCollection();
  ​}
}

class B
{
  /**
    * @ORM\ManyToOne(targetEntity=A::class, inversedBy="bs")
    * @ORM\JoinColumn(nullable=false)
    */
  ​private ?A $a = null;

  /**
   * @ORM\ManyToOne(targetEntity=C::class, inversedBy="bs", cascade={"persist"})
   * @ORM\JoinColumn(nullable=false)
   */
  ​private ?C $c = null;

  ​private ?string $reference = null;
}

class C
{
  /**
   * @ORM\OneToMany(targetEntity=B::class, mappedBy="c")
   */
  private Collection $bs;

  ​private ?string $reference = null;

  public function __construct() {
    ​$this->bs = new ArrayCollection();
  ​}
}
```

There is listener on A and C to generate tha A, B and C references. The A listener loop over the collection of B. It works well in the application, but when I want to create a fixture for A... It totally explode, when Foundry call the persist, there is no B inside the collection (in A class). But to generate the A reference I need the first B class of the collection and the linked C class.

Someone know how we can do this ? Actually I've something similar to:

AFactory::createOne();

Internally:
 - AFactory call BFactory::new()->many(3);
 - BFactory call CFactory::new();

kbond:
Does it work if you create the factories manually on their own? Create C, create A, then create B with these?

I find using foundry for one-to-many a bit finicky for complex scenarios. I tend to favor creating many-to-one's only.

mpiot:
It didn't for A and B, it's ok for C. In this case, C can live alone. But a B can't exists without a A, and a A can't exists without at least a B.

All work without the prePersist listener that generate the reference. Because to generate the A reference the listener get the first B, then the attached C. And to define the B reference, get the C then the A. A and B must be persisted together with a pre-existing C or by saving the C to.

[...]

PR #181, opened 2021-07-30
Doctrine cascade persist usage

mpiot:
This PR fix (#179) by adding the detection of `cascade={"persist"}` in Entities, this permit to Foundry to not call `->persist()` and `->flush()` on cascaded Factories.

1. An entity do not cascade persist a relation: Factory create the related Factory, then **call**  `->persist()` and `->flush()` on the entity, then attach this one to the parent entity via the `PropertyAccessor` component.
2. An entity has a cascade persist relation: Factory create the related Factory, **do not call**  `->persist()` and `->flush()` on the entity, it just attach this one to the parent entity via the `PropertyAccessor` component.

To do:
 - [x] ManyToOne
 - [x] OneToMany
 - [x] OneToOne (owner side)
 - [x] OneToOne (reverse side
 - [x] ManyToMany (owner side)
 - [x] ManyToMany (reverse side)
 - [x] Functionnal tests


mpiot:
@kbond Sorry for the break, I just came back from vacation. Do you have an idea about manage ManytoMany reverse side, I do not find a way to do that.

I start to write tests for the others.

kbond:
What exactly is the problem with "ManytoMany reverse side"?

mpiot:
I don't find where this case is handle, I just see that I never reach this case.

I start testing, maybe easier to do that after having a test that not passes, but for the momemt I would like tests passes first: Psalm (I don't know how fix it, because I use Interfaces attributes to know which fields are cascaded or not), and many tests fail with: `RuntimeException: Foundry was booted without doctrine. Ensure your TestCase extends Symfony\Bundle\FrameworkBundle\Test\KernelTestCase`.

src/Factory.php
```diff
@@ -302,6 +312,13 @@ function($value) {
             $value = $value->withoutPersisting();
         }
 
+        // Check if the attribute is cascade persist
+        $relationField = $this->relationshipField($value) ?? $this->inverseRelationshipField($value);
```
kbond, line 316: This is causing the `RuntimeException: Foundry was booted without doctrine.` failure. It is possible to use foundry without doctrine being available. It just creates the object w/o persisting. These functions try to access doctrine which causes the exception.

mpiot: Thanks, I add a condition with `self::configuration()->hasManagerRegistry()`.

src/Factory.php
```diff
@@ -202,6 +205,13 @@ final public function instantiateWith(callable $instantiator): self
         return $cloned;
     }
 
+    final public function setCascadePersist(bool $cascadePersist): self
```
kbond, line 208: I'd prefer this not be part of the public api. Just access and set the private property directly.

src/FactoryCollection.php
```diff
@@ -57,12 +57,15 @@ static function(Factory $factory) use ($attributes) {
      *
      * @psalm-return list<Factory<TObject>>
      */
-    public function all(): array
+    public function all(bool $cascadePersist = false): array
```
kbond, line 60: Let's remove this and add it directly to the factories after calling ->all()

src/Factory.php
```diff
@@ -326,7 +348,23 @@ private function normalizeCollection(FactoryCollection $collection): array
             return [];
         }
 
-        return $collection->all();
+        return $collection->all($cascadePersist);
```
kbond, line 351: Use an array map here to set the private property directly.

kbond (changes requested):

kbond:
I'd like to understand the `$cascadePersist` property and why you don't just use `->withoutPersisting()` for the nested factory when "cascade persist" is detected?

A => B (cascade persist)

In the example above, simply calling `->withoutPersisting()` on B before creating would be enough wouldn't it?

I'm thinking the following would be a problem though:

A => B (cascade persist) => C (no cascade persist)

During creating A, B would be created without persist, but so would C (so C wouldn't be persisted). Is this the reasoning?

Another question: could usage of `->withoutPersisting()` solve your problem in #179 (without the need for this PR)?

Below assumes: A => B (cascade persist) <=> C (cascade persist)

```php
AFactory::create([
    'b' => BFactory::new([
        'c' => CFactory::new()->many(5), // because B has withoutPersisting, this would inherit to C
    ])->withoutPersisting(),
]);
```

mpiot:
Yes, that's related to:
```php
 if (!$this->isPersisting()) {
    // ensure attribute Factory's are also not persisted
    $value = $value->withoutPersisting();
}
```

In the case of the cascade persistance, each relation should be analyzed then persisted or not depending it. Actually, the scope of `isPersisting` is on all children.

---

The problem is that some relations are cascaded other none:
A cascade persist B that cascade persist C, but each of them has relations from existing entities (that must be persisted before). I've use that workaround on simple things, but here a single A need arround at leats 20 entities, because A is linked on others, linked on others, ..., thats the same in C.

kbond:
> The problem is that some relations are cascaded other none:

So, is this accurate to what your describing?

```php
AFactory::create([
    'b' => BFactory::new([
        'c' => CFactory::new()->many(5), // !!! no way to force c to be persisted
    ])->withoutPersisting(),
]);
```

src/Factory.php
```diff
@@ -302,6 +305,15 @@ function($value) {
             $value = $value->withoutPersisting();
         }
 
+        // Check if the attribute is cascade persist
+        if (self::configuration()->hasManagerRegistry()) {
+            $relationField = $this->relationshipField($value) ?? $this->inverseRelationshipField($value);
+            $cascadePersist = $this->hasCascadePersist($value, $relationField);
+            if (true === $cascadePersist) {
+                $value->setCascadePersist(true);
+            }
```
kbond, line 314:
```suggestion
            $value->cascadePersist = $this->hasCascadePersist($value, $relationField);
```

mpiot:
I can retry to remember why, but I've try lots of solutions, and the withoutPeristing haven't help to solve it.

kbond:
Ok, fair enough. I was just trying to evaluate if the `$cascadePersist` is required but I think it is.

mpiot:
No problem, I've try without it, but its goal is different of the `$persist` and not find a way to do without it. I mainly looked for how to find the `cascade` with Doctrine from entities.

src/Factory.php
```diff
@@ -344,6 +379,32 @@ private function inverseRelationshipField(self $factory): ?string
...
+
+        $collectionClass = $factory->class;
+        $factoryClass = $this->class;
+        $collectionMetadata = self::configuration()->objectManagerFor($collectionClass)->getClassMetadata($collectionClass);
+        $classMetadataFactory = self::configuration()->objectManagerFor($factoryClass)->getMetadataFactory()->getMetadataFor($factoryClass);
+
+        if (false === $collectionMetadata->hasAssociation($field)) {
+            return false;
+        }
+
```
kbond, line 396: I think the below psalm errors can be solved with:

```php
if (!$classMetadata instanceof \Doctrine\ORM\Mapping\ClassMetadataInfo) {
    return false;
}
```

mpiot: Thanks, I've start by use method instead of public attributes to avoid 3 Psalm errors, but this one appear after, and I've not check it.

mpiot:
I will now take a look at the tests, just a question about them. We need to keep relations without cascade (actual tests), and add some with cascades (6 relations, or maybe 3 with reverse side, we don't do that in real app, not sure it works, but avoid to add lot of Entities/new relations) to test if Foundry detect them properly and do not persist them.

Do I create new entities only for them ? Or prefer to add cascaded relations on existing entities ? Like same attributes with `Cascaded` suffix ?

src/Factory.php
```diff
@@ -344,6 +379,36 @@ private function inverseRelationshipField(self $factory): ?string
...
+        if (null === $field) {
+            return false;
+        }
+
+        $collectionClass = $factory->class;
+        $factoryClass = $this->class;
+        $collectionMetadata = self::configuration()->objectManagerFor($collectionClass)->getClassMetadata($collectionClass);
+        $classMetadataFactory = self::configuration()->objectManagerFor($factoryClass)->getMetadataFactory()->getMetadataFor($factoryClass);
+
+        if (!$collectionMetadata instanceof \Doctrine\ORM\Mapping\ClassMetadataInfo || !$classMetadataFactory instanceof \Doctrine\ORM\Mapping\ClassMetadataInfo) {
```
kbond, line 393: Sorry, didn't show this in my example but can you import these?

kbond:
For the tests, I'd prefer all new entities to keep it separate. Suffixing them with `Cascade` is fine - don't feel you have to duplicate the ones we have (with all the fields). For these new entities, just add the fields required to test the cascade.

One thing though: how are you going to prove the cascade is "working"? _I believe_, unless you're doing something like in #179 (event listener), cascaded entities _just work_ with foundry without this change. We'll need to ensure the tests fail if we remove the existing code in this PR.

mpiot:
> One thing though: how are you going to prove the cascade is "working"? I believe, unless you're doing something like in #179 (event listener), cascaded entities just work with foundry without this change. We'll need to ensure the tests fail if we remove the existing code in this PR.

This is a good question, for the moment you're right the only way it to check it like in #179. Because, the result is the same: all is persisted.

I've think about Listener, because this is this Doctrine Feature that cause trouble in my case. By doing a simple case: `A -> B`. The `prePersist` event on the A class should give us: `A->getB()` return non persisted class if cascaded, and a persisted class if not cascaded (by checking if the id is `null` or not).

kbond:
> I've think about Listener, because this is this Doctrine Feature that cause trouble in my case. By doing a simple case: A -> B. The prePersist event, in that event A->getB() should return non persisted entity if cascaded, and a persisted entity if not cascaded (by checking if the id is null or not).

Good idea, that should work!

mpiot:
For entities, I`ve think to create independent entities (not same names at all) with a little logic about the cascade.

 - Brand (OneToMany $products)
 - Product (ManyToOne $brand  and OneToMany $variants)
 - Variant (ManyToOne $product and OneToOne $image as owner)
 - Image (OneToOne $variant as reverse)
 
In a `Brand` we have many `Products` that have many `Variant` with an `Image` each. For test purpose, we consider we can create all the chain from `Brand` that handle the creation of the Product, Variant and Image, or because we like strange creation forms create all from the Image that create the Variant, Product and Brand.

I`ve try to find a logical example to keep things comprehensive in tests. The only problem is the inverse that is not really logic (create from Image), we also can create other relations. Depends if the preference is to keep it simple (less new Entities, Factory as possible) or keep things full logic.

The idea of create something different was just to keep it separated without having same name: User, Post, Tag and find relations that interest us in that case. Plus there is no OneToOne relation in actual tests, and ManyToMany (reverse side). That might be confusing to have same entity names but different inside.

kbond:
I agree, test as much as you can but if we don't cover every scenario, we can catch these later and add logical tests if/when we find issues.

So for the tests you are creating, should they cover all the scenarios you have checked in this PR's description?

For the untested scenario(s), do you believe this fix _should_ work?

mpiot:
Actually the PR handle:
 - ManyToOne
 - OneToMany
 - OneToOne (owner side)
 - OneToOne (reverse side
 - ManyToMany (owner side)

If I remember before vacations, the ManyToMany (reverse) do not work, I want to create tests for all of them, then if possible find a way to fix the only test that not pass ( ManyToMany (reverse)).

mpiot:
Do you have an idea why just addig a new Entity class in tests/Fixtures/Entity like:
```php
<?php

namespace Zenstruck\Foundry\Tests\Fixtures\Entity;

use Doctrine\ORM\Mapping as ORM;

/**
 * @ORM\Entity
 * @ORM\Table(name="toto")
 */
class Toto
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @ORM\Column(type="string", length=255)
     */
    private $name;

    public function getName(): ?string
    {
        return $this->name;
    }

    public function setName($name)
    {
        $this->name = $name;
    }
}
```

Throw errors in some tests, and skip the lot of tests (compare no add entity) ? If I remove the entity all works well, but just add an entity, all became slow and fail.

```shell
mpiot@milou:~/Development/foundry$ DATABASE_URL=postgresql://website:website@127.0.0.1:5432/zenstruck_foundry?charset=utf8 vendor/bin/simple-phpunit -c phpunit-dama-doctrine.xml
PHPUnit 9.4.4 by Sebastian Bergmann and contributors.

Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!

Testing
ESSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSESSSSESESSSESSSSSSSSSSSSSSSSSS  63 / 194 ( 32%)
SSSSSSSSSSESSSSSSSSSSSSSSSSESSSSSSSSSSSSSSSSSSESSSSS........... 126 / 194 ( 64%)
............................................................... 189 / 194 ( 97%)
....

Time: 01:20.679, Memory: 36.00 MB
```

kbond (approved):

kbond:
Looks good! Let me know when/if you are finished and I'll merge.

mpiot:
It's okay for me ! Thanks for your help during this PR and issue :)

kbond:
Thanks @mpiot!

PR #185, opened 2021-08-20
Be able to remove most of method annotations on user factories

Nyholm:
PHPStorm 2021.2 supports generics just like psalm. 

It make sure we can use `@extends ModelFactory<\App\Entity\Post>` to remove the long list of annotations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,18 +196,7 @@ use Zenstruck\Foundry\ModelFactory;
 use Zenstruck\Foundry\Proxy;
 
 /**
- * @method static Post|Proxy createOne(array $attributes = [])
- * @method static Post[]|Proxy[] createMany(int $number, $attributes = [])
- * @method static Post|Proxy find($criteria)
- * @method static Post|Proxy findOrCreate(array $attributes)
- * @method static Post|Proxy first(string $sortedField = 'id')
- * @method static Post|Proxy last(string $sortedField = 'id')
- * @method static Post|Proxy random(array $attributes = [])
- * @method static Post|Proxy randomOrCreate(array $attributes = []))
- * @method static Post[]|Proxy[] all()
- * @method static Post[]|Proxy[] findBy(array $attributes)
- * @method static Post[]|Proxy[] randomSet(int $number, array $attributes = []))
- * @method static Post[]|Proxy[] randomRange(int $min, int $max, array $attributes = []))
+ * @extends ModelFactory<\App\Entity\Post>
  * @method static PostRepository|RepositoryProxy repository()
  * @method Post|Proxy create($attributes = [])
  */
--- a/README.md
+++ b/README.md
@@ -212,7 +212,6 @@ use Zenstruck\Foundry\Proxy;
  * @method static Post[]|Proxy[] randomRange(int $min, int $max, array $attributes = []))
  * @method static PostRepository|RepositoryProxy repository()
  * @method Post|Proxy create(array|callable $attributes = [])
- * @method Post[]|Proxy[] many(int $number, array|callable $attributes = [])
  */
 final class PostFactory extends ModelFactory
 {
--- a/README.md
+++ b/README.md
@@ -211,7 +211,8 @@ use Zenstruck\Foundry\Proxy;
  * @method static Post[]|Proxy[] randomSet(int $number, array $attributes = []))
  * @method static Post[]|Proxy[] randomRange(int $min, int $max, array $attributes = []))
  * @method static PostRepository|RepositoryProxy repository()
- * @method Post|Proxy create($attributes = [])
+ * @method Post|Proxy create(array|callable $attributes = [])
+ * @method Post[]|Proxy[] many(int $number, array|callable $attributes = [])
  */
 final class PostFactory extends ModelFactory
 {
--- a/README.md
+++ b/README.md
@@ -198,6 +198,7 @@ use Zenstruck\Foundry\Proxy;
 /**
  * @extends ModelFactory<\App\Entity\Post>
  * @method static PostRepository|RepositoryProxy repository()
+ * @method static Post[]|Proxy[] createMany(int $number, $attributes = [])
  * @method Post|Proxy create($attributes = [])
  */
 final class PostFactory extends ModelFactory
--- a/README.md
+++ b/README.md
@@ -196,9 +196,21 @@ use Zenstruck\Foundry\ModelFactory;
 use Zenstruck\Foundry\Proxy;
 
 /**
- * @extends ModelFactory<\App\Entity\Post>
- * @method static PostRepository|RepositoryProxy repository()
+ * @extends ModelFactory<Post>
+ *
+ * @method static Post|Proxy createOne(array $attributes = [])
  * @method static Post[]|Proxy[] createMany(int $number, $attributes = [])
+ * @method static Post|Proxy find($criteria)
+ * @method static Post|Proxy findOrCreate(array $attributes)
+ * @method static Post|Proxy first(string $sortedField = 'id')
+ * @method static Post|Proxy last(string $sortedField = 'id')
+ * @method static Post|Proxy random(array $attributes = [])
+ * @method static Post|Proxy randomOrCreate(array $attributes = []))
+ * @method static Post[]|Proxy[] all()
+ * @method static Post[]|Proxy[] findBy(array $attributes)
+ * @method static Post[]|Proxy[] randomSet(int $number, array $attributes = []))
+ * @method static Post[]|Proxy[] randomRange(int $min, int $max, array $attributes = []))
+ * @method static PostRepository|RepositoryProxy repository()
  * @method Post|Proxy create($attributes = [])
  */
 final class PostFactory extends ModelFactory
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -5,7 +5,7 @@ namespace Zenstruck\Foundry;
 use Faker;
 
 /**
- * @template TObject as object
+ * @template TObject of object
  * @abstract
  *
  * @author Kevin Bond <kevinbond@gmail.com>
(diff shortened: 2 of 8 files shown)


Nyholm:
The unrelated style changes are fixed in https://github.com/zenstruck/foundry/pull/186 (including editorconfig)

src/Factory.php
```diff
@@ -5,7 +5,7 @@
 use Faker;
 
 /**
- * @template TObject as object
```
Nyholm, line 8: "as" is basically a substitution. "of object" means that we cannot use non-objects

src/Factory.php
```diff
@@ -68,8 +68,7 @@ public function __call(string $name, array $arguments)
     /**
      * @param array|callable $attributes
      *
-     * @return Proxy|object
```
Nyholm, line 71: This will tell PHPStan that the returned object is both a Proxy and a TObject.

src/ModelFactory.php
```diff
@@ -95,6 +92,9 @@ final public static function findOrCreate(array $attributes): Proxy
     /**
      * @see RepositoryProxy::first()
      *
+     * @return Proxy<TModel>&TModel
+     * @psalm-return Proxy<TModel>
```
Nyholm, line 96: Psalm will (correctly) complain on `@return Proxy<TModel>&TModel`. That is why we need tell psalm (only psalm) the return type is actually `Proxy<TModel>`.

src/ModelFactory.php
```diff
@@ -18,11 +15,13 @@ public function __construct()
         parent::__construct(static::getClass());
     }
 
-    public static function __callStatic(string $name, array $arguments)
+    /**
+     * @return list<TModel&Proxy<TModel>>
+     * @psalm-return list<Proxy<TModel>>
+     */
+    public static function createMany(): array
```
kbond, line 22: Shouldn't we use the real method parameters here? (`int $number, array|callable $attributes = []`)?

kbond (commented):
Cool, thanks! Just the one question.

src/ModelFactory.php
Nyholm: Yes, you are correct. I've updated the PR.

src/ModelFactory.php
```diff
@@ -18,13 +15,13 @@ public function __construct()
         parent::__construct(static::getClass());
     }
 
-    public static function __callStatic(string $name, array $arguments)
```
Nyholm, line 21: Im not sure why this method was "magic" from the beginning.

kbond: This is because there used to be a `createMany` instance method on `Factory` that we deprecated (ref #111).

README.md
```diff
@@ -196,18 +196,7 @@ use Zenstruck\Foundry\ModelFactory;
...
- * @method static Post|Proxy findOrCreate(array $attributes)
- * @method static Post|Proxy first(string $sortedField = 'id')
- * @method static Post|Proxy last(string $sortedField = 'id')
- * @method static Post|Proxy random(array $attributes = [])
- * @method static Post|Proxy randomOrCreate(array $attributes = []))
- * @method static Post[]|Proxy[] all()
- * @method static Post[]|Proxy[] findBy(array $attributes)
- * @method static Post[]|Proxy[] randomSet(int $number, array $attributes = []))
- * @method static Post[]|Proxy[] randomRange(int $min, int $max, array $attributes = []))
+ * @extends ModelFactory<\App\Entity\Post>
```
kbond, line 199: The `Post` entity is imported above so the FQCN is not required here.

Have your confirmed auto-completion works as expected in the latest version of PHPStorm? I've had a lot of problems with generic support in this version.

If so, should we update the maker template?

src/ModelFactory.php
kbond: I looked into this a bit further to see why in #111 we didn't do this. There was a typo in a test proving this `__callStatic()` is required (see #187).

Once I merge #187 and you rebase this PR, this test will fail. So we do need to keep `__callStatic()` (until 2.0 anyway).

src/Factory.php
```diff
@@ -68,8 +68,7 @@ public function __call(string $name, array $arguments)
...
-     *
+     * @return Proxy<TObject>&TObject
```
wouterj, line 71: No need for `&TObject`, as the `Proxy` mixin is already taking care of this?

src/ModelFactory.php
wouterj: There already is a non-static `createMany` method, so we cannot define it normally (like you tried to do now). The BC layer has to use the magic method.

wouterj: There must be a `createMany` static and a `createMany` non-static method (in `Factory`). The non-static one is deprecated. Because of this, we have to use `__callStatic` and `__call` magic and cannot define the method directly afaics.

src/ModelFactory.php
wouterj: Same as previous, is there a need for `&TModel`? The `@mixin` on `Proxy` should take care of that.

README.md
kbond: As discussed with @Nyholm in slack, I think we need to keep the existing annotations here and in the factory maker template. At least for the foreseeable future. I'd like to have auto-completion work for the widest spectrum of users as possible.

src/Factory.php
Nyholm: You are correct. However, PHPStorm do not understand that. 

So we add the `@return` to make sure PHPStorm is auto completing correctly, then we do `@psalm-return` with the "correct" return type so psalm does not complain.

src/ModelFactory.php
Nyholm: I dont understand the `@mixin`, could you tell me how that works?

Nyholm:
Screenshot 2021-08-20 at 13 50 31
Screenshot 2021-08-20 at 13 32 24


------

I've noticed that PHPStorm does not understand inheritance. Ie. `Factory::create()` and `Factory::many()` does not get auto completion support.

kbond:
> I've noticed that PHPStorm does not understand inheritance. Ie. Factory::create() and Factory::many() does not get auto completion support.

Do you mean something like `CandidateFactory::new()->create()->getUuid()` (`getUuid()` would not be auto-completable)?

Nyholm:
Yes. But I may fix that.. I'll check

src/ModelFactory.php
Nyholm: I spent over an hour on this. 

There is no way for PHP to detect if you call a static method in an non-static way. The current implementation (with magic methods) is the only way to detect that.

Nyholm:
So https://github.com/zenstruck/foundry/pull/185/commits/9f3d7cfeefd24a0396aacbdc93b3b311983b7176 fixed the issue with `create()` and `many()`. 

This is as good as we can get it because PHPStorm does not support inheritance. 
The readme shows **all** annotations. 

Here is the minimum list (that I use in my applications)

```php
/**
 * @extends ModelFactory<Candidate>
 *
 * @method static Candidate[]|Proxy[] createMany(int $number, $attributes = [])
 * @method static CandidateRepository|RepositoryProxy repository()
 * @method Candidate|Proxy create(array|callable $attributes = [])
 * @method Candidate[]|Proxy[] many(int $number, array|callable $attributes = [])
 */
class CandidateFactory extends ModelFactory
{
```

README.md
```diff
@@ -209,7 +211,8 @@ use Zenstruck\Foundry\Proxy;
  * @method static Post[]|Proxy[] randomSet(int $number, array $attributes = []))
  * @method static Post[]|Proxy[] randomRange(int $min, int $max, array $attributes = []))
  * @method static PostRepository|RepositoryProxy repository()
- * @method Post|Proxy create($attributes = [])
+ * @method Post|Proxy create(array|callable $attributes = [])
+ * @method Post[]|Proxy[] many(int $number, array|callable $attributes = [])
```
kbond, line 215: Hmm, I don't think this is right. `many()` returns a `FactoryCollection`. Not sure if `FactoryCollection<Post>|FactoryCollection<Proxy>` would work..?

src/ModelFactory.php
Nyholm: From the release notes of https://github.com/vimeo/psalm/releases/tag/3.8.2

> Support for the @mixin docblock annotation, which allows a class to proxy the methods and properties of another.
>
> I also made a slight change in how magic getters & setters are dealt with – Psalm will emit an UndefinedMagicProperty if you define a docblock @property string $foo and then request the unlisted property $bar. If you wanted this behaviour previously you had to also annotate your class with @psalm-seal-properties. This is now assumed if you provide a @mixin or @property annotation for the given class. In the event the property is missing, Psalm now emits an UndefinedMagicProperty issue that you can suppress in a more granular fashion.

Nyholm: So again, you are correct that psalm will understand that. But the `&TModel` is for PHPStorm. 

I think this change is valid.

Nyholm (commented):
This PR is ready for review

README.md
Nyholm: Is there a `Proxy` in the `FactoryCollection`? Isn't it just `Post`?

Nyholm: Neither will help PHPStorm and psalm will do well without this. 

I've removed this line.

kbond (approved):
LGTM, @wouterj, do you mind doing a final review?

src/ModelFactory.php
wouterj: As far as I can see, PHPstorm understands mixin? (not a phpstorm user myself though, so I can't test)

https://youtrack.jetbrains.com/issue/WI-1730
https://stackoverflow.com/a/48298989/1149495

wouterj:
If Psalm agrees, I agree :)

PHPdocs have always been "intentionally wrong" about types to improve autocompletion, so the intersection type is OK by me (but I'm still not sure if it's really needed).

kbond:
Thanks @Nyholm!

PR #187, opened 2021-08-20
Ensure legacy test works as expected

kbond:
This fixes a typo in a legacy test introduced in #111. It wasn't actually triggering the deprecation.


wouterj:
A suggestion: if we use the `ExpectDeprecationTrait` from the PHPunit bridge, we can add assertions for a deprecation being triggered: `$this->expectDeprecation('the expected message');`. This makes sure I won't ever write these type of bad tests again :wink:

kbond:
Good idea! There is one other place we don't do this. I'll fix.

PR #189, opened 2021-08-23
Update maker factory template, add documentation about using with PhpStorm 2021.2+

kbond:


src/Bundle/Resources/skeleton/Factory.tpl.php
```diff
@@ -10,9 +10,11 @@
 use Zenstruck\Foundry\Proxy;
 
 /**
+ * @extends ModelFactory<<?= $entity->getShortName() ?>>
+ *
  * @method static <?= $entity->getShortName() ?>|Proxy createOne(array $attributes = [])
```
maks-rafalko, line 15: should it be `&Proxy` rather than `|Proxy`? After generating factories, PHPStan complains with errors unless I change it to intersection type instead of union.

kbond: This is a tough one. PHPStorm 2021.3.2 doesn't autocomplete the object when an intersection is used.

I think the proper type should be `Proxy<<?= $entity->getShortName() ?>>` but that doesn't autocomplete either...

PR #205, opened 2021-10-21
Update MakeFactory.php

### LINKED ISSUE #204: `make:factory` with YAML mapping - `Undefined index: nullable` for ID field
Using similar mapping example to this:

https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/yaml-mapping.html#example

`make:factory` won't work because `nullable` key is not part of the id field. 

https://github.com/zenstruck/foundry/blob/c483b2ebed14ac95c0b1ca23b92909fe75f853ef/src/Bundle/Maker/MakeFactory.php#L201

I apologize for raising this issue if this works for other mapping formats (and you dont support yaml) and/or similar issue already exists.

domagoj-orioly:
Check if index exists before accessing it.

Fixes #204

### DIFF
--- a/src/Bundle/Maker/MakeFactory.php
+++ b/src/Bundle/Maker/MakeFactory.php
@@ -198,7 +198,7 @@ final class MakeFactory extends AbstractMaker
 
         foreach ($metadata->fieldMappings as $property) {
             // ignore identifiers and nullable fields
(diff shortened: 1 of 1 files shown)


kbond:
Thanks for the report/fix @domagoj-orioly!

Issue #210, opened 2021-10-25
Factory create extra unintended records when cascade persist is enabled


anjeylink:
Following documentation on how to work with doctrine relations I got unintended records for reverse side of relation with configured cascade persist. https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#doctrine-relationships

For example I have **Post** with **One-To-Many relation with Comment**.
Post field `comments` is configured **cascade persist**.

PostFactory:

```php

    protected function getDefaults(): array
    {
        return [
            'name' => self::faker()->word(),
        ];
    }
```

CommentFactory:

```php

    protected function getDefaults(): array
    {
        return [
            'text' => self::faker()->word(),
            'post' => PostFactory::new(),
        ];
    }
```

Test:

```php

    public function testThatCreatePostWithCommentsWillCreateOnlyOnePostRecord()
    {
        PostFactory::createOne([
            'comments' => CommentFactory::new()->many(5),
        ]);

        PostFactory::assert()->count(1);
    }
```

Expect that there will be only one Post entity but there are 6.
Tests failed starting from 1.13 version. It's working in 1.12.

kbond:
Closing for now.

PR #211, opened 2021-10-25
Cascade persist creates extra entities when factory has default

kbond:
See #210 for example of when this failure occurs. #181 introduced this problem.


kbond:
Going to leave the solution as [using withoutPersisting](https://github.com/zenstruck/foundry/issues/210#issuecomment-951089529) for now.

Issue #212, opened 2021-10-27
Create entity with OneToOne  not nullable association gives an error


bastien70:
Hello, in my entity, I've a **OneToOne** relation with another entity, like this : 

**UserPlanning.php**

```php
class UserPlanning
{
  //...

    #[ORM\OneToOne(inversedBy: 'planning', targetEntity: User::class, cascade: ['persist', 'remove'])]
    #[ORM\JoinColumn(nullable: false)]
    private ?User $user = null;

    public function getUser(): ?User
    {
        return $this->user;
    }

    public function setUser(?User $user): self
    {
        $this->user = $user;

        return $this;
    }
}
```

And the other side in **User.php** : 

```php
class User
{
  //...
  #[ORM\OneToOne(mappedBy: 'user', targetEntity: UserPlanning::class, cascade: ['persist', 'remove'])]
    #[Assert\Valid]
    private ?UserPlanning $planning = null;

    public function getPlanning(): ?UserPlanning
    {
        return $this->planning;
    }

    public function setPlanning(?UserPlanning $planning): self
    {
        $this->planning = $planning;

        // set the owning side of the relation if necessary
        if ($planning->getUser() !== $this) {
            $planning->setUser($this);
        }

        return $this;
    }
}
```

I've in my **UserFactory.php** : 

```php
class UserFactory
{
  //...

     protected function getDefaults(): array
    {
        return [
            //...
            'planning' => UserPlanningFactory::new(),
        ];
    }
}
```

But when calling my **UserFactory** to create some users and initialize at the same time their planning, like this : 

```php
$employees = UserFactory::new()
      ->many(10)
      ->create();
```

I've the following error : 

```
An exception occurred while executing 'INSERT INTO user_planning (updated_at, user_id, monday_id, tuesday_id, wednesday_id, thursday_id, friday_id, saturday_i  
  d, sunday_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)' with params ["2021-10-27", null, 96, 97, 98, 99, 100, null, null]:                                            
                                                                                                                                                                  
  SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'user_id' cannot be null
```

If I replace in my **UserPlanning**, the `nullable: false` to `nullable: true`, it works, but the property shouldn't be nullable, so I can't. Is there an alternative to initialize at the same time the **UserPlanning** with the **User** object ?

kbond:
[...]

PR #213, opened 2021-10-27
One-to-One non-nullable reverse problem

kbond:
_I think_ demonstrates the issue in #212.


kbond:
This was a mis-configuration of one-to-one.

Issue #215, opened 2021-11-23
Remove doctrine event listeners/subscribers temporarily


CharloMez:
I have an issue with some testing cases:

*  When i want to test my doctrine event subscribers in a functional test, I want predefined data to perform a PUT request.
My subscribers are fired when i instantiate foundry fixtures, so i could verify nothing with my call. 
* Same for Functional tests on symfony command

So sometimes I would like to store fixtures that goes raw in database, without any listeners on their creation.

Maybe we can already does it, but I didn't found how,
So I made a little code snippet for that:
```
use Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface;
use Doctrine\ORM\EntityManagerInterface;

...

    public function withoutDoctrineListeners()
    {
        /** @var EntityManagerInterface $manager */
        $manager = FooFactory::configuration()->objectManagerFor(Foo::class);
        $listeners = $manager->getConnection()->getEventManager()->getListeners();
        foreach ($listeners as $type => $listenersByType) {
            foreach ($listenersByType as $listener) {
                if ($listener instanceof EventSubscriberInterface) {
                    $manager->getConnection()->getEventManager()->removeEventSubscriber($listener);

                    continue;
                }

                $manager->getConnection()->getEventManager()->removeEventListener($type, $listener);
            }
        }
    }
```

So do you think is it something you can propose in foundry ?

I thought about something like (function name is probably bad):
```
        FooFactory::new()->withoutDoctrineListeners()->create();
```

I can submit a pull request if you like the idea

kbond:
I do like the idea but there are some nuances we need to think about.

When calling your code that actually tests the listeners, do you do anything to put the event listeners "back"?

CharloMez:
Yeah I agree with that, we need to put back listeners after creation.  
I thought about something else, the function could have an optional argument, "$eventTypeName":
```
FooFactory::new()->withoutDoctrineListeners(Events::postPersist)->create();
```

The function "getListeners" have this optional argument, so it would be nice to allow it and follow it to this function.

Do you want to add the feature yourself, or it is ok for you to let me make a pull request for this ?

kbond:
> I thought about something else, the function could have an optional argument, "$eventTypeName":

Sure, makes sense.

>  or it is ok for you to let me make a pull request for this

Go ahead with a PR if you're willing!

nikophil:
stale issue, feel free to reopen it

PR #216, opened 2021-11-25
feature(doctrine events): allow factories to disable doctrine events.

CharloMez:
feature(doctrine events): allow factories to disable doctrine events before creation. (issue #215)

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -128,9 +128,7 @@ class Factory
             return $proxy;
         }
 
-        $proxy->save();
-
-        $proxy->withoutAutoRefresh(function(Proxy $proxy) use ($attributes) {
+        $proxy->save()->withoutAutoRefresh(function(Proxy $proxy) use ($attributes) {
             foreach ($this->afterPersist as $callback) {
                 $proxy->executeCallback($callback, $attributes);
             }
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -2,10 +2,11 @@
 
 namespace Zenstruck\Foundry;
 
-use Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface;
+use Doctrine\Common\EventSubscriber;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Faker;
+use Symfony\Bridge\Doctrine\ContainerAwareEventManager;
 
 /**
  * @template TObject of object
@@ -338,6 +339,11 @@ class Factory
             $value = $value->withoutPersisting();
         }
 
+        if (false === $this->doctrineEvents) {
+            // ensure attribute Factory's doctrine events are also not fired
+            $value = $value->withoutDoctrineEvents();
+        }
+
         // Check if the attribute is cascade persist
         if (self::configuration()->hasManagerRegistry()) {
             $relationField = $this->relationshipField($value);
@@ -481,16 +487,18 @@ class Factory
 
         /** @var EntityManagerInterface $manager */
         $manager = self::configuration()->objectManagerFor($this->class);
+        /** @var ContainerAwareEventManager $eventManager */
+        $eventManager = $manager->getConnection()->getEventManager();
 
         foreach ($this->disabledDoctrineEvents as $type => $listenersByType) {
             foreach ($listenersByType as $listener) {
-                if ($listener instanceof EventSubscriberInterface) {
-                    $manager->getConnection()->getEventManager()->addEventSubscriber($listener);
+                if ($listener instanceof EventSubscriber) {
+                    $eventManager->addEventSubscriber($listener);
 
                     continue;
                 }
 
-                $manager->getConnection()->getEventManager()->addEventListener($type, $listener);
+                $eventManager->addEventListener($type, $listener);
             }
         }
     }
@@ -499,17 +507,27 @@ class Factory
     {
         /** @var EntityManagerInterface $manager */
         $manager = self::configuration()->objectManagerFor($this->class);
+        /** @var ContainerAwareEventManager $eventManager */
+        $eventManager = $manager->getConnection()->getEventManager();
 
         $this->disabledDoctrineEvents = $manager->getConnection()->getEventManager()->getListeners($this->doctrineEventTypeName);
         foreach ($this->disabledDoctrineEvents as $type => $listenersByType) {
-            foreach ($listenersByType as $listener) {
-                if ($listener instanceof EventSubscriberInterface) {
+            foreach ($listenersByType as $name => $listener) {
+                if ($listener instanceof EventSubscriber) {
                     $manager->getConnection()->getEventManager()->removeEventSubscriber($listener);
 
                     continue;
                 }
 
-                $manager->getConnection()->getEventManager()->removeEventListener($type, $listener);
+                if (false === \mb_strpos($name, '_service_')) {
+                    // Handle listeners with hash as name, ex: '000000001f7b9da2000000001c719ccc'
+                    $eventManager->removeEventListener($type, $listener);
+                } else {
+                    // Handle listeners with name like:
+                    // '_service_Zenstruck\Foundry\Tests\Fixtures\Event\CommentEventListener'
+                    // '_service_doctrine.orm.default_listeners.attach_entity_listeners'
+                    $eventManager->removeEventListener($type, \str_replace('_service_', '', $name));
+                }
             }
         }
     }
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -2,6 +2,8 @@
 
 namespace Zenstruck\Foundry;
 
+use Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Faker;
 
@@ -33,6 +35,9 @@ class Factory
 
     /** @var string|null */
     private $doctrineEventTypeName = null;
+
+    /** @var array */
+    private $disabledDoctrineEvents = [];
 
     /** @var bool */
     private $cascadePersist = false;
@@ -119,13 +124,13 @@ class Factory
         }
 
         if (false === $this->doctrineEvents) {
-            $proxy->disableDoctrineEvents($this->doctrineEventTypeName);
+            $this->disableDoctrineEvents();
         }
 
         $proxy->save();
 
         if (false === $this->doctrineEvents) {
-            $this->resetDoctrineEvents($proxy);
+            $this->resetDoctrineEvents();
         }
 
         return $proxy->withoutAutoRefresh(function(Proxy $proxy) use ($attributes) {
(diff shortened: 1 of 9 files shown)


kbond:
Thanks for working on this. Don't worry about the CS issues - I can fix before merging.

CharloMez:
@kbond actually I tried to build a docker environment to make tests works in local env, I partially succeed (I can't run functional tests).  
So it's almost a blind devlopment, do you have something to help me on it ? Maybe a docker-compose environment ? Makefile ? .env ?

CharloMez:
I set up a docker environment with php/pqsql/nginx, I configured it well, the DATABASE_URL is well set in the container, but I still don't succeed to run your test suite. I always have this error
`Error running "doctrine:database:drop":  in /var/www/src/Test/DatabaseResetter.php:123`
I miss something but I don't understand what.

Except that, I can see that my functional tests doesn't work on the CI, my doctrine EventSubscriber fixture is not executed.

kbond:
Unfortunately, I don't have docker experience. I run everything locally. I was going to suggest doing what you did but am unsure why you're getting that error. Any more context? That error should output the command output.

kbond:
For the test event subscriber: try having it implement `Doctrine\Bundle\DoctrineBundle\EventSubscriber\EventSubscriberInterface`. It doesn't look like `Doctrine\Common\EventSubscriber` can be autowired.

kbond:
Regarding running the tests locally, could https://github.com/nektos/act help? I haven't used it myself but see it referenced quite a bit.

CharloMez:
ok finally I succeed to make it run locally (with my docker setup), and it works locally with a php8.

kbond:
Great to hear. I'd like to make the test suite easier to run for contributors. Is there anything you can share that could help? Are there other/bundles libraries you know if that I can use as inspiration?

tests/Fixtures/Kernel.php
```diff
@@ -58,6 +59,10 @@ protected function configureContainer(ContainerBuilder $c, LoaderInterface $load
             ->setAutoconfigured(true)
             ->setAutowired(true)
         ;
+        $c->register(CommentEventSubscriber::class)
```
kbond, line 62: You'll have tag this as an event subscriber manually instead of relying on autoconfiguration. Early versions of doctrine bundle don't register this interface for autoconfiguration.

CharloMez:
@kbond finally CI is ok (except scrutinizer, I don't know why).  
I will try to share my docker-compose environment after some cleaning

src/Proxy.php
```diff
@@ -222,6 +226,47 @@ public function disableAutoRefresh(): self
         return $this;
     }
 
+    public function disableDoctrineEvents(?string $eventTypeName = null): self
```
kbond, line 229: Do you have a use-case for having these methods on the `Proxy`? If not, I'd prefer this code to be in `Factory` (and remove this public API).

kbond (changes requested):

src/Proxy.php
CharloMez: done :+1:

kbond (commented):
Just a few comments:
1. Relation factory's created during `create()` would currently have events called even if disabled by the parent. There are certain scenarios where they would be disabled though (ie if cascade: persist).
2. The `afterPersist` hook is called with disabled events re-enabled.

Are the above two things intentional or what are your thoughts on these?

If a user has chosen to disable events, I'm wondering if all of create should have events disabled. WDTY?

kbond:
Just an idea but what do you think of this API:

```php
Factory::withoutDoctrineEvents(function() {
    PostFactory::createOne(); // events disabled
    PostFactory::createOne(); // events disabled
});

PostFactory::createOne(); // events enabled
```

CharloMez:
> Just a few comments:
> 
> 1. Relation factory's created during `create()` would currently have events called even if disabled by the parent. There are certain scenarios where they would be disabled though (ie if cascade: persist).
> 2. The `afterPersist` hook is called with disabled events re-enabled.
> 
> Are the above two things intentional or what are your thoughts on these?
> 
> If a user has chosen to disable events, I'm wondering if all of create should have events disabled. WDTY?

No it was not intentional, I didn't really thought about it. Actually I thought to do it like the  `withoutPersisting()` functionality. 

I think the functionality is good like that for some reasons:
* I feel like it is same as `withoutPersisting` functionality, and different way to use them will loose a little the user
* In this way, if the user want all the process to be without events he can do it (he explicite `withoutDoctrineEvents` on all factory). But if we force all the process to be without events, the user can do nothing if he just want the first entity without events.
* It's personal but I don't like much this kind of declaration
* And finally we can still correct anytime the feature if the community have troubles to use it like that

kbond:
> I feel like it is same as withoutPersisting functionality, and different way to use them will loose a little the user

FYI, `withoutPersisting` is "passed" down to sub-factory's

> It's personal but I don't like much this kind of declaration

Is [this](https://github.com/zenstruck/foundry/pull/216#issuecomment-981696281) what you're referring to?

> And finally we can still correct anytime the feature if the community have troubles to use it like that

If we merge as-is, then decide to disable events for the entire create, this would be a BC break.

CharloMez:
> FYI, `withoutPersisting` is "passed" down to sub-factory's

Ah ok I didn't notice, my bad. But when I said that, I thought more about different way to declare it in tests

> Is [this](https://github.com/zenstruck/foundry/pull/216#issuecomment-981696281) what you're referring to?

Yes

> If we merge as-is, then decide to disable events for the entire create, this would be a BC break.

Yeah true... 

> In this way, if the user want all the process to be without events he can do it (he explicite withoutDoctrineEvents on all factory). 
> But if we force all the process to be without events, the user can do nothing if he just want the first entity without events.

And what do you think about that ?

You have a much better vision of this, and where does it go, so if you think it's better to follow the `withoutDoctrineEvents` to child creations, let's do it.

kbond:
> In this way, if the user want all the process to be without events he can do it (he explicite withoutDoctrineEvents on all factory). But if we force all the process to be without events, the user can do nothing if he just want the first entity without events.

I'm a bit torn to be honest, you're right in that we couldn't provide a way to disable just for the primary factory but:

```php
PostFactory::new()->withoutDoctrineEvents()->create();
```

My initial instinct expects all doctrine events to be disabled during "create".

CharloMez:
ok let's do it like that

kbond:
Ok, yeah, I'm thinking let's disable events for all of create. The solution for allowing sub-factory's to be created with events will be to create them first:

```php
$categroy = CategoryFactory::createOne(); // created with events
PostFactory::new()->withoutDoctrineEvents()->create(['category' => $category]);
```

WDYT?

CharloMez:
> Ok, yeah, I'm thinking let's disable events for all of create. The solution for allowing sub-factory's to be created with events will be to create them first:
> 
> ```
> $categroy = CategoryFactory::createOne(); // created with events
> PostFactory::new()->withoutDoctrineEvents()->create(['category' => $category]);
> ```
> 
> WDYT?

It's ok for me :+1: 
 i will try to do it soon. thanks

CharloMez:
@kbond sorry it took me times to do it...
I figured out that this dev didn't work on listener, so I handle it in the last commit, and I add specific tests for listeners.  
Usually I don't like to comment my code, but I feel that this new specific part needed some.  
And of course I also add what we said about sub-factories, included tests.

Don't hesitate to tell me if something is not good for you.

CharloMez:
@kbond I have a repository for a docker-compose environnement on php8/pgsql/nginx, and i made a specific branch for foundry.
https://github.com/CharloMez/php8-nginx-pg-bootstrap_for_symfony/tree/feature/zenstruckFoundryMakefile

Tell me what do you think about it.  Feel free to ask me some improvement on it.

src/Factory.php
```diff
@@ -440,4 +479,56 @@ private function isPersisting(): bool
...
+                    continue;
+                }
+
+                if (false === \mb_strpos($name, '_service_')) {
+                    // Handle listeners with hash as name, ex: '000000001f7b9da2000000001c719ccc'
+                    $eventManager->removeEventListener($type, $listener);
+                } else {
+                    // Handle listeners with name like:
+                    // '_service_Zenstruck\Foundry\Tests\Fixtures\Event\CommentEventListener'
+                    // '_service_doctrine.orm.default_listeners.attach_entity_listeners'
```
kbond, line 528:
```suggestion
                    // Symfony\Bridge\Doctrine\ContainerAwareEventManager allows container service id's to be registered as event listeners
                    // these are prefixed with "_service_"
```

kbond (changes requested):

kbond:
> Tell me what do you think about it. Feel free to ask me some improvement on it.

I'm not very familiar with docker but a few comments:
1. The test suite doesn't require a web server so probably shouldn't include php-fpm/nginx
2. Should include a mysql container to be able to run test suite on mysql

kbond:
> sorry it took me times to do it...

No worries and no pressure/rush - we're all doing this in our spare time!

The test failures must be related to this PR but I can't see how. I merged a PR a few hours ago that didn't have failures. Maybe a rebase would fix?

src/Factory.php
```diff
@@ -440,4 +479,56 @@ private function isPersisting(): bool
...
+        $this->disabledDoctrineEvents = $manager->getConnection()->getEventManager()->getListeners($this->doctrineEventTypeName);
+        foreach ($this->disabledDoctrineEvents as $type => $listenersByType) {
+            foreach ($listenersByType as $name => $listener) {
+                if ($listener instanceof EventSubscriber) {
+                    $manager->getConnection()->getEventManager()->removeEventSubscriber($listener);
+
+                    continue;
+                }
+
+                if (false === \mb_strpos($name, '_service_')) {
```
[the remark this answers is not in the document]
CharloMez: Yeah actually I'm not really satisfied by this code, but I don't see how to do better.  
It is well tested, if this keyword change, it will break my test.

src/Factory.php
```diff
@@ -440,4 +479,56 @@ private function isPersisting(): bool
...
+                    // Handle listeners with hash as name, ex: '000000001f7b9da2000000001c719ccc'
+                    $eventManager->removeEventListener($type, $listener);
```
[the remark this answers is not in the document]
CharloMez: This is well tested because in the test I call twice `withoutDoctrineEvents()`, when the `create()` method of the first one finish, it re enable listeners and register them with that kind of name.  
So the second  `withoutDoctrineEvents()` test this part.

CharloMez:
> > sorry it took me times to do it...
> 
> No worries and no pressure/rush - we're all doing this in our spare time!
> 
> The test failures must be related to this PR but I can't see how. I merged a PR a few hours ago that didn't have failures. Maybe a rebase would fix?

I rebased it, but still fail, I don't really understand why...

CharloMez:
> I'm not very familiar with docker but a few comments:
> 
> 1. The test suite doesn't require a web server so probably shouldn't include php-fpm/nginx
> 2. Should include a mysql container to be able to run test suite on mysql

yeah true, I will remove nginx for this branch.  

It is a very simple docker-compose conf, to be able to execute the tests suite.  

You are right, a perfect configuration would be to have all different kind of database and all different php version handled by your project, but doing such environment will require to create a system to be able to switch between all of that.  
It require much more reflexion and skills in dockerisation, I can think of it but I'm not sure I have the skill for that.

src/Factory.php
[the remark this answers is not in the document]
kbond: I guess this is all we can do.

src/Factory.php
```diff
@@ -112,11 +128,19 @@ function($value) {
             return $proxy;
         }
 
-        return $proxy->save()->withoutAutoRefresh(function(Proxy $proxy) use ($attributes) {
+        $proxy->save();
+
+        $proxy->withoutAutoRefresh(function(Proxy $proxy) use ($attributes) {
```
kbond, line 133:
```suggestion
        $proxy->save()->withoutAutoRefresh(function(Proxy $proxy) use ($attributes) {
```

kbond (changes requested):
Small fix.

Since the failing test is happening when running the test suite with dama-doctrine-bundle, I wonder if messing with the events is breaking something. When I have a chance, I'll checkout the code and take a look.

Edit: I think this [event](https://github.com/dmaicher/doctrine-test-bundle/blob/master/src/DAMA/DoctrineTestBundle/Doctrine/DBAL/PostConnectEventListener.php) is being removed and causing problems. We'll need to somehow only remove ORM events (this might be easier said than done).

CharloMez:
> Small fix.
> 
> Since the failing test is happening when running the test suite with dama-doctrine-bundle, I wonder if messing with the events is breaking something. When I have a chance, I'll checkout the code and take a look.
> 
> Edit: I think this [event](https://github.com/dmaicher/doctrine-test-bundle/blob/master/src/DAMA/DoctrineTestBundle/Doctrine/DBAL/PostConnectEventListener.php) is being removed and causing problems. We'll need to somehow only remove ORM events (this might be easier said than done).

hmm yeah I see. Maybe a more reasonable way to handle this functionality is to give explicitly the subscriber or listener to remove. like:
`->withoutDoctrineEvent(CommentEventListener::class)` 
And allow the user to chain call if needed, like:
`->withoutDoctrineEvent(CommentEventListener::class)->withoutDoctrineEvent(CommentEventSubscriber::class)->create()`

src/Factory.php
CharloMez: This don't solve the problem

kbond: I didn't think it would, this was just a minor thing.

kbond:
> Maybe a more reasonable way to handle this functionality is to give explicitly the subscriber or listener to remove. like:

I agree, while not ideal, I don't see an easy way to distinguish between ORM and DBAL events.

kbond:
I'm closing this as stale but do think it's a valid feature. Feel free to reopen!

PR #218, opened 2021-12-03
Add all-fields option to MakeFactory

gnito-org:
This PR adds the option to tell MakeFactory that it must create default values for all required and nullable fields of the entity, not only required fields.

It's easier to manually remove unwanted defaults than it is to manually add additional defaults.

The default behavior of MakeFactory has not changed. Without the option on the command line, MakeFactory will create defaults only for required fields.

If the PR is approved, I will make a Symfony documentation PR to add the option there as well.

### DIFF
--- a/src/Bundle/Maker/MakeFactory.php
+++ b/src/Bundle/Maker/MakeFactory.php
@@ -99,7 +99,7 @@ final class MakeFactory extends AbstractMaker
         }
 
         if (!$input->getOption('all-fields')) {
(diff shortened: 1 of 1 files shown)


kbond:
Thanks for the PR @gnito-org. What do you think about an interactive message so this is discoverable.

gnito-org:
> Thanks for the PR @gnito-org. What do you think about an interactive message so this is discoverable.

Good idea. I think the commits will do it.

kbond:
Thanks @gnito-org!

kbond:
> If the PR is approved, I will make a Symfony documentation PR to add the option there as well.

FYI, those doc's are hosted in this repo

PR #221, opened 2021-12-10
make:factory auto defaults for non-nullable 1-1/n-1 relationships

benblub:
Its not finished yet.


benblub:
@kbond 
Short Question how to use doctrine migrations in this project?
i tried `php vendor/bin/doctrine-migrations migrations:generate` but that isnt working without a configuration File.

benblub:
Im not sure what new Entity names should named. So i just added Foo & Bar Entity to can test the different relations. (OneToOne, ManyToOne, OneToMany & ManyToMany). As i need them to be non nullable for my tests i didnt put them in existing Entities. So nothing breaks.

benblub:
Some more Questions
1. Why is the composer.lock on gitignore? I added to project, we can remove is there is some reason.
2. i replaced the hardcoded class_exist factory with symfony finder search factory in src & tests. Not sure this can leak to performance issues on large Projects. But as the cmd argument --namespace exist on creating factories we need to search over all folders/files in src & tests folder.. 
3. local i used some dockerized setup to setup & run this project easily. I would like to add this to the repo?

kbond:
> Short Question how to use doctrine migrations in this project?

I believe I just used symfony/demo to create the same entities, created the migration and copy and pasted the generated code into foundry.

> Im not sure what new Entity names should named. So i just added Foo & Bar Entity that i can test the different relations.

That's fine.

> Why is the composer.lock on gitignore? I added to project, we can remove is there is some reason.

For packages, there is no reason so keep it in the repo - it would change every time someone runs composer update in different environments.

> local i used some dockerized setup to setup & run this project easily. I would like to add this to the repo?

I would love something like this in the repo but in a different PR

.gitignore
```diff
@@ -8,4 +8,5 @@
 /.phpunit.result.cache
 /tests/Fixtures/tmp
 /var/
+/.idea/
```
kbond, line 11: Please remove this change - you can add this to your global .gitignore:

```
echo "/.idea/" > ~/.gitignore_global
```

src/Bundle/Maker/MakeFactory.php
```diff
@@ -202,6 +203,40 @@ private function defaultPropertiesFor(string $class, bool $allFields): iterable
...
+
+            $joinedColumns = $item['joinColumns'];
+            $fieldName = $item['fieldName'];
+
+            $targetEntityArray = \explode('\\', $item['targetEntity']);
+            $targetEntity = \end($targetEntityArray);
+            $factory = \ucfirst($targetEntity).'Factory';
+
+            if ($this->isFactory($factory)) {
+                yield \lcfirst($fieldName) => \ucfirst($targetEntity).'Factory::createOne(),';
```
kbond, line 236:
```suggestion
                yield \lcfirst($fieldName) => \ucfirst($targetEntity).'Factory::new(),';
```

This will prevent factory's being created even when defaults are overridden.

kbond (changes requested):
I know you're still working on this but I found a few minor things after a quick review.

benblub:
@kbond 
im not sure about the feature ask user to create missing factory. i added some init code for it. But it looks a bit complex..
The create missing factory function has on his side same problem that other side isnt created yet. 

examble create Foo mentioned Bar is missing and ask the user to create Bar too. But Bar is on this step missing Foo 🤦 
Maybe we use same approach like option all-fields..? One problem still exist. We create Bar but its missing Foo. Workaround is hasFactory need some flag like $ignore = null. And if ignore is true we can generate fields for missing factory which is created soon. But what i mentioned code looks still some complex with the feature create missing factories.

Dunno we want build this in this PR too or not?

kbond:
Let's leave _asking the user_ for another PR. This at least lays the foundation for that.

src/Bundle/Maker/MakeFactory.php
```diff
@@ -218,4 +248,31 @@ private function defaultPropertiesFor(string $class, bool $allFields): iterable
             yield $property['fieldName'] => $value;
         }
     }
+
+    private function hasFactory($factory)
```
kbond, line 252: As discussed in slack, let's use the injected factories instead. `$entitiesWithFactories` should be an array map of `Entity FQCN: Factory FQCN`

tests/Fixtures/Entity/Bar.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Zenstruck\Foundry\Tests\Fixtures\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="bars")
+ */
+class Bar
```
kbond, line 11: Let's remove this entity and just use the `Category` fixture for Foo's relationships. It already has it's factory as a service so it will just work for the test.

src/Bundle/Maker/MakeFactory.php
```diff
@@ -202,6 +203,35 @@ private function defaultPropertiesFor(string $class, bool $allFields): iterable
...
+            }
+
+            $fieldName = $item['fieldName'];
+
+            $targetEntity = \explode('\\', $item['targetEntity']);
+            $targetEntity = \end($targetEntity);
+            $factory = \ucfirst($targetEntity).'Factory';
+
+            if ($this->hasFactory($factory)) {
+                yield \lcfirst($fieldName) => \ucfirst($factory).'::new(),';
```
kbond, line 230: Let's yield the Factory's FQCN if not in the same namespace as this new factory.

tests/Fixtures/Entity/Foo.php
```diff
@@ -0,0 +1,132 @@
...
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="foos")
+ */
+class Foo
```
kbond, line 13: Let's rename to `EntityWithRelations` (not a great name but not as generic as Foo).

tests/Fixtures/Entity/Foo.php
```diff
@@ -0,0 +1,132 @@
...
+     * @ORM\GeneratedValue
+     * @ORM\Column(type="integer")
+     */
+    private $id;
+
+    /**
+     * @ORM\OneToOne(targetEntity=Bar::class, cascade={"persist", "remove"})
+     * @ORM\JoinColumn(nullable=false)
+     */
+    private $oneToOne;
```
kbond, line 26: Can you add a nullable one-to-one (to ensure it doesn't get created)

tests/Fixtures/Entity/Foo.php
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * @ORM\OneToMany(targetEntity=Bar::class, mappedBy="foo")
+     */
+    private $oneToMany;
+
+    /**
+     * @ORM\ManyToOne(targetEntity=Bar::class)
+     * @ORM\JoinColumn(nullable=false)
+     */
+    private $manyToOne;
```
kbond, line 37: Again, can you add a nullable many-to-one?

tests/Fixtures/Entity/Foo.php
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * @ORM\ManyToMany(targetEntity=Bar::class)
+     */
+    private $manyToMany;
+
+    public function __construct()
```
kbond, line 44: I think we can remove the boilerplate below - we aren't actually creating these entities.

tests/Fixtures/Factories/BarFactory.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Zenstruck\Foundry\Tests\Fixtures\Factories;
+
+use Zenstruck\Foundry\ModelFactory;
+use Zenstruck\Foundry\Tests\Fixtures\Entity\Bar;
+
+/**
+ * @author Benjamin Knecht <benjaminknecht@wearewondrous.com>
+ */
+final class BarFactory extends ModelFactory
```
kbond, line 11: Can be removed with `Bar`.

tests/Fixtures/Migrations/Version20210820131819.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Tests\Fixtures\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20210820131819 extends AbstractMigration
```
kbond, line 10: Adding/updating/renaming the entities will change this migration, I know generating these migrations are a pain - sorry about that.

tests/Functional/Bundle/Maker/MakeFactoryTest.php
```diff
@@ -6,13 +6,89 @@
...
+
+        // TODO inject services if required (https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#factories-as-services)
+    }
+
+    protected function getDefaults(): array
+    {
+        return [
+            // TODO add your default values here (https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#model-factories)
+            'oneToOne' => BarFactory::new(),
+            'manyToOne' => BarFactory::new(),
```
kbond, line 69: I believe this will change to:

```suggestion
            'oneToOne' => Zenstruck\Foundry\Tests\Fixtures\Factories\CategoryFactory::new(),
            'manyToOne' => Zenstruck\Foundry\Tests\Fixtures\Factories\CategoryFactory::new(),
```

kbond (changes requested):

src/Bundle/Maker/MakeFactory.php
kbond: I'm not sure what you mean. Are you trying to actually run the maker to test the changes?

benblub:
- [ ] Create new migration File

benblub:
@kbond Do you something about that Psalm error? 
````
Run vendor/bin/psalm --output-format=github
  vendor/bin/psalm --output-format=github
  shell: /usr/bin/bash -e {0}
Scanning files...
Analyzing files...

E
Error: Cannot find file composer-bin-proxy:/home/runner/work/foundry/foundry/bin/tools/psalm/vendor/vimeo/psalm/psalm to include (see https://psalm.dev/107)
Error: Process completed with exit code 2.
````

locally i can run psalm..

kbond:
Don't worry about psalm, it's failing on the main branch as well.

nikophil:
Hello @benblub 

I think for now we'll just inform the user that a factory is missing. In the future we might directly propose to create the missing factories.

BTW, do you have any plans on finishing this PR? If no, no problem, I can do it :)

benblub:
> Hello @benblub
> 
> I think for now we'll just inform the user a factory is missing. In the future we might directly propose to create the missing factories.
> 
> BTW, do you have any plans on finishing this PR? If no, no problem, I can do it :)

Hi @nikophil 
Would be nice if you work on this. I didn't see any time slot in near future to continue my work.

nikophil:
I'm closing this PR in favor of #351

PR #223, opened 2021-12-13
Fix smallint generated Faker

jmsche:
Hi,

The currently configured default faker for smallints is wrong, it should be numberBetween() for specifying an int range instead of randomNumber().

Very small PR :)


kbond:
Thank you @jmsche!

PR #226, opened 2021-12-18
Correct spelling in index.rst

gnito-org:


kbond:
Thank you @gnito-org.

PR #228, opened 2021-12-20
Add dependent field caution in index.rst

gnito-org:


kbond:
Thanks for the PR but I'm hesitant to add this as is. As I mentioned in the issue, I believe this is a domain issue. If the order in which setters can be called is important, this is a code smell.

I think a note in [this section](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#attributes) about how attributes are built would be best. Something like:

"Attributes passed to the `create*` methods are merged with any set via `getDefaults()` and `withAttributes()`."

gnito-org:
I changed the wording.

docs/index.rst
```diff
@@ -187,6 +187,11 @@ This command will generate a ``PostFactory`` class that looks like this:
         }
     }
 
+.. note::
```
kbond, line 190: Can you move to the "Attributes" section below (right above the "Faker" title)

kbond (changes requested):

docs/index.rst
gnito-org: Done.

kbond:
Thank you @gnito-org.

PR #229, opened 2021-12-22
Fix event hook argument name

Aeet:


docs/index.rst
```diff
@@ -455,7 +455,7 @@ they were added.
             // $object is the instantiated object
             // $attributes contains the attributes used to instantiate the object and any extras
         })
-        ->afterPersist(function(Proxy $object, array $attributes) {
+        ->afterPersist(function(Proxy $proxy, array $attributes) {
             /* @var Post $object */
```
kbond, line 459: This should be changed as well.

```suggestion
            /* @var Post $proxy */
```

kbond (changes requested):
Thanks! Just one minor thing.

kbond:
Thanks @Aeet.

PR #231, opened 2021-12-29
Allow any type for Story States

wouterj:
Our use-case: we generate multiple fixtures with a specific relationship (e.g. "updated_by" must be set). In another story, we want to use one of these fixtures when creating another fixture.  
"updated_by" is a dynamic value, so we can't use `randomOrCreate(['updated_by' => ...])`. Due to the strict `object` type-hint, we can also not use Story states to transfer the knowledge to the other story.

If we relax the type, it would be possible to store the full list of generated fixtures in the state and use `array_rand()` to pick a random fixture from the list in the second story.

What do you think?

### DIFF
--- a/src/Story.php
+++ b/src/Story.php
@@ -7,8 +7,8 @@ namespace Zenstruck\Foundry;
  */
 abstract class Story
 {
-    /** @var array<string, Proxy> */
-    private $objects = [];
+    /** @var array<string, mixed> */
+    private $state = [];
 
     final public function __call(string $method, array $arguments)
     {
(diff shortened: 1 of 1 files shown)


kbond:
This makes sense to me. The _story system_ was probably the least thought out part of Foundry. It likely has more potential to be exploited.

kbond:
@wouterj am I missing something or should `Story::get()` have it's type hint removed?

wouterj:
Uh-oh, yes you're correct. In the end, we ended up solving this in a different way (using specialized helper classes in the story) so I never got to test this feature :grimacing:

kbond:
Do you think we should revert or fix? (Are you still wanting this?)

wouterj:
If Foundry implements https://github.com/zenstruck/foundry/issues/250 , this is no longer needed.

PR #241, opened 2022-01-04
[doc] mongo-support documentation

kbond:
The integration with ORM/MongoDB/both is pretty seamless. Not much we have to say except "announce" we support these `doctrine/persistence` implementations. @nikophil, wdyt?


nikophil:
I kinda agree, I think there not much things to add...

nikophil (approved):

PR #245, opened 2022-01-19
[Experiment] Use gh action default mysql service

kbond:
Ref: https://twitter.com/dunglas/status/1483878551749152768?s=21

- [x] MySQL
- [ ] PostgreSQL


kbond:
Does not appear worth it. May even slow down tests?

Issue #248, opened 2022-02-02
Ignore abstract classes in maker


abeal-hottomali:
Suggestion: In `\Zenstruck\Foundry\Bundle\Maker\MakeFactory::entityChoices`, currently is:
```php
        foreach ($this->managerRegistry->getManagers() as $manager) {
            foreach ($manager->getMetadataFactory()->getAllMetadata() as $metadata) {
                if (!\in_array($metadata->getName(), $this->entitiesWithFactories, true)) {
                    $choices[] = $metadata->getName();
                }
            }
        }
```
Should include:
```php
        foreach ($this->managerRegistry->getManagers() as $manager) {
            foreach ($manager->getMetadataFactory()->getAllMetadata() as $metadata) {
                if ($metadata->getReflectionClass()->isAbstract()) {
                    continue;
                }
                if (!\in_array($metadata->getName(), $this->entitiesWithFactories, true)) {
                    $choices[] = $metadata->getName();
                }
            }
        }
```

kbond:
Good call! Do you mind creating a PR?

abeal-hottomali:
Sure, [added](https://github.com/zenstruck/foundry/pull/249)

PR #249, opened 2022-02-03
Ignores abstract classes in the maker

abeal-hottomali:
Updates `\Zenstruck\Foundry\Bundle\Maker\MakeFactory::entityChoices` to ignore abstract classes when considering which to generate factories for.  Should close https://github.com/zenstruck/foundry/issues/248


kbond:
Thank you @abeal-hottomali

PR #252, opened 2022-02-10
Story pools

### LINKED ISSUE #250: Random "named"/"prefixed" references
Hi,

Is there a simple way to get random "named"/"prefixed" items from a Story?

Eg. I have fixtures that generate some states in several countries, and then I'd like to create cities for some random Belgian states.

So something like the wildcard reference in nelmio/alice: https://github.com/nelmio/alice/blob/master/doc/relations-handling.md#references

kbond:
Closes #250.

Adds the following API to Story's:

```php
final class ProvinceStory extends Story
{
    public function build(): void
    {
        // add collection to a "pool"
        $this->addToPool('be', ProvinceFactory::createMany(5, ['country' => 'BE']));

        // equivalent to above
        $this->addToPool('be', ProvinceFactory::new(['country' => 'BE'])->many(5));

        // add single object to a pool
        $this->addToPool('be', ProvinceFactory::createOne(['country' => 'BE']));

        // add single object to single pool and make available as "state"
        $this->addState('be-1', ProvinceFactory::createOne(['country' => 'BE']), 'be');
    }
}
```

Usage:

```php
ProvinceStory::getRandom('be'); // random Proxy from "be" pool
ProvinceStory::getRandomSet('be', 3); // 3 random Proxy's from "be" pool
ProvinceStory::getRandomRange('be', 1, 4); // between 1 and 4 random Proxy's from "be" pool
```

TODO:
- [x] Tests
- [x] Docs

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1656,7 +1656,9 @@ Later, you can access the story's state when creating other fixtures:
 
 .. code-block:: php
 
-    PostFactory::createOne(['category' => CategoryStory::load()->get('php')]);
+    PostFactory::createOne([
+        'category' => CategoryStory::load()->get('php') // Category Proxy
+    ]);
 
     // or use the magic method (functionally equivalent to above)
     PostFactory::createOne(['category' => CategoryStory::php()]);
@@ -1664,6 +1666,46 @@ Later, you can access the story's state when creating other fixtures:
 .. note::
 
     Story state is cleared after each test (unless it is a :ref:`Global State Story <global-state>`).
+
+Story Pools
+^^^^^^^^^^^
+
+Stories can store (as state) *pools* of objects:
+
+.. code-block:: php
+
+    // src/Story/ProvinceStory.php
+
+    namespace App\Story;
+
+    use App\Factory\ProvinceFactory;
+    use Zenstruck\Foundry\Story;
+
+    final class ProvinceStory extends Story
+    {
+        public function build(): void
+        {
+            // add collection to a "pool"
+            $this->addToPool('be', ProvinceFactory::createMany(5, ['country' => 'BE']));
+
+            // equivalent to above
+            $this->addToPool('be', ProvinceFactory::new(['country' => 'BE'])->many(5));
+
+            // add single object to a pool
+            $this->addToPool('be', ProvinceFactory::createOne(['country' => 'BE']));
+
+            // add single object to single pool and make available as "state"
+            $this->addState('be-1', ProvinceFactory::createOne(['country' => 'BE']), 'be');
+        }
+    }
+
+Objects can be fetched from pools in your tests, fixtures or other stories:
+
+.. code-block:: php
+
+    ProvinceStory::getRandom('be'); // random Province|Proxy from "be" pool
+    ProvinceStory::getRandomSet('be', 3); // 3 random Province|Proxy's from "be" pool
+    ProvinceStory::getRandomRange('be', 1, 4); // between 1 and 4 random Province|Proxy's from "be" pool
 
 Bundle Configuration
 --------------------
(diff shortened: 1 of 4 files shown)


src/Story.php
```diff
@@ -29,40 +32,124 @@ final public static function load(): self
...
+        return \array_slice($values, 0, \random_int($min, $max));
+    }
+
+    /**
+     * @param object|Proxy|Factory $object
+     *
      * @return static
      */
-    final public function add(string $name, $state): self
+    final public function add(string $name, object $object, string ...$pools): self
```
kbond, line 88: Should I make this `string|array $pools` (we've previously chosen to avoid variadic variables)?

jmsche: Wouldn't it be better to force the array type instead? And also have `addToPools()` in addition to `addToPool()`

Variadic arguments are an issue if you want to add more args after that.

wouterj: Adding it to multiple pools seems very advanced to me. I would suggest only implementing `string $pool` now and then accepting a community contribution if someone really needs to add one fixture to multiple pools.

kbond: Ok, I've simplified - the PR description is updated to the new API.

src/Story.php
```diff
@@ -29,40 +32,124 @@ final public static function load(): self
     }
 
     /**
+     * Get a random item from a pool.
+     */
+    final public static function getRandom(string $pool): Proxy
```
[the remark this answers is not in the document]
kbond: Leaving off for now - can be added later if desired.

kbond:
I believe this to be ready, @jmsche, @wouterj: mind giving a final review?

docs/index.rst
```diff
@@ -1665,6 +1667,46 @@ Later, you can access the story's state when creating other fixtures:
...
     Story state is cleared after each test (unless it is a :ref:`Global State Story <global-state>`).
 
+Story Pools
+^^^^^^^^^^^
+
+Stories can store (as state) *pools* of objects:
+
+.. code-block:: php
+
+    // src/Story/CategoryStory.php
```
jmsche, line 1677:
```suggestion
    // src/Story/ProvinceStory.php
```

src/Story.php
```diff
@@ -29,47 +32,134 @@ final public static function load(): self
...
+    }
+
+    /**
+     * Get a random set of items from a pool.
+     *
+     * @return Proxy[]
+     */
+    final public static function getRandomSet(string $pool, int $number): array
+    {
+        if ($number < 0) {
```
jmsche, line 49: Just wondering: is there any reason to allow "0" as $number?

src/Story.php
```diff
@@ -29,47 +32,134 @@ final public static function load(): self
...
+            throw new \InvalidArgumentException(\sprintf('$max (%d) cannot be less than $min (%d).', $max, $min));
+        }
+
+        $story = static::load();
+        $values = $story->pools[$pool] ?? [];
+
+        \shuffle($values);
+
+        if (\count($values) < $max) {
+            throw new \RuntimeException(\sprintf('At least %d items must be on pool "%s" (%d items found).', $max, $pool, \count($values)));
```
jmsche, line 77: "in" pool? (not sure about that)

src/Story.php
kbond: I changed to `$number < 1`.

kbond:
Thanks for the review @jmsche! Would you say this adequately solves #250 for you?

jmsche:
> Would you say this adequately solves https://github.com/zenstruck/foundry/issues/250 for you?

Yes, I think it should solve this issue.

> Thanks for the review @jmsche!

And thank you for everything :)

jmsche:
Hi Kevin, do you think you can create a new release including this new feature? Or are you waiting to finish an other PR first?

kbond:
Done! https://github.com/zenstruck/foundry/releases/tag/v1.17.0

jmsche:
Thank you for your fast reaction :)

PR #253, opened 2022-02-10
Make `Story::get()` static (implies `Story::load()->get()`)

kbond:
Instead of calling `MyStory::load()->get('some-state')`, can call `MyStory::get('some-state')`. This deprecates calling `->get()` instance method.

### DIFF
--- a/src/Story.php
+++ b/src/Story.php
@@ -4,6 +4,8 @@ namespace Zenstruck\Foundry;
 
 /**
  * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @method static Proxy get(string $name)
  */
 abstract class Story
 {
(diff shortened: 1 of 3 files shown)


wouterj:
> @wouterj, what do you think about this? There is now no longer a way to access story state items without loading the story. Do you think this is a problem?

I would ask this question the other way (and I'm sure you did already and know the answer, but I'm missing some context): What does this change solve?

Having a `$story->get('countries')` method with proper PHP declaration (and thus SA and auto completion) seems better than `$story->countries()` to me personally.

kbond:
> What does this change solve?

This PR is purely a quality of life improvement:

```diff
- MyStory::load()->get('countries');
+ MyStory::get('countries');
```

> Having a $story->get('countries') method with proper PHP declaration (and thus SA and auto completion) seems better than $story->countries() to me personally.

This PR doesn't change this ability. The idea is you'd add `@method` annotations to your story to help with autocompletion.

src/Story.php
```diff
@@ -4,6 +4,8 @@
 
 /**
  * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @method static Proxy get(string $name)
```
nikophil, line 8: I guess there is no solution to "type" this `Proxy`?

nikophil (approved):

src/Story.php
kbond: I don't think so, a user would have to add a custom `@method`

PR #254, opened 2022-02-10
Deprecate `Story:add()` and add protected `Story::addState()`

kbond:
This method was never intended to be called from outside. This is technically a BC break but calling it from outside could do weird things. @wouterj, do you think we should deprecate instead? Is there a deprecation pattern for changing a method's visibility?

Another option could be to deprecate `Story::add()` and add `final protected function addState(string $name)`. This would line up with #252's `Story::addToPool()`.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1645,10 +1645,10 @@ Another feature of *stories* is the ability for them to *remember* the objects t
     {
         public function build(): void
         {
-            $this->add('php', CategoryFactory::createOne(['name' => 'php']));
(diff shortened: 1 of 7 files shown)


wouterj:
> This is technically a BC break but calling it from outside could do weird things. @wouterj, do you think we should deprecate instead? Is there a deprecation pattern for changing a method's visibility?

Yes, adding a new method and deprecating the current one is the best BC path here (if you want to provide one).

kbond:
Thanks @wouter, I've added a deprecation path.

PR #257, opened 2022-02-19
feat: add return type for command (deprecated in symfony 6)

Chris53897:
`Method "Symfony\Component\Console\Command\Command::execute()" might add "int" as a native return type declaration in the future. Do the same in child class "Zenstruck\Foundry\Bundle\Command\StubCommand" now to avoid errors or add an explicit @return annotation to suppress this message.`


kbond:
Thanks!

Issue #270, opened 2022-04-21
Factory for embeddable classes only working when using "withoutPersisting"


mergeMarc:
I have multiple embedded classes and factories for those. I get an exception like this when i don't initialize all those factories with 'withoutPersisting()' (in this example ProcessedSms is a embeddable class):

```
In ORMInvalidArgumentException.php line 63:
                                                                               
  [Doctrine\ORM\ORMInvalidArgumentException]                                   
  The given entity of type 'MyVendor\MyProject\Message\ProcessedSms' (MyVendor\MyProject\Message\ProcessedSms@541)
  has no identity/no id values set. It cannot be added to the identity map.  
```

Is this intended behavior? I think that #192 says otherwise.
If this is intended behavior it would be nice if the docs would mention having to use `withoutPersisting()` with embeddables.

Please note that i'm using Foundry for the first time so it could be an issue on my side.

kbond:
You should not have to call `->withoutPersisting()` on embeddables. Can you share some example code? Perhaps you discovered an edge case.

mergeMarc:
I quickly threw together an example project to replicate the issue: [FoundryExample](https://github.com/mergeMarc/FoundryExample)

When i run `php bin/console doctrine:fixtures:load` i get the following exception in that project:
```
In ORMInvalidArgumentException.php line 66:
                                                                               
  The given entity of type 'App\Entity\IcEngine' (App\Entity\IcEngine@218) has no 
  identity/no id values set. It cannot be added to the identity map.
```

I commented out the `->withoutPersisting()` in the factories. If those are uncommented everything works fine.

kbond:
Perfect, thank you. Will debug deeper but my first thought is the nested embeddables is the issue. I've never come across this but if Doctrine supports, then foundry should as well.

PR #271, opened 2022-04-22
Fix embeddable support when used with file (ie xml) mapping

kbond:
The annotaion/attribute drivers do not register embeddables in class metadata but the file drivers do.

Reference: https://github.com/doctrine/orm/pull/8138
Fixes: #270

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -449,7 +449,13 @@ class Factory
             return false;
         }
 
+        if ($classMetadata instanceof ORMClassMetadata && $classMetadata->isEmbeddedClass) {
+            // embedded entity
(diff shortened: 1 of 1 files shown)


kbond:
I opted not to include tests as it would require _yet another_ test matrix element (duplicating all the mapping), further complicating the test suite.

kbond:
@mergeMarc, can you confirm this fixes #270 in your _real_ app? Then I'll merge/release.

mergeMarc:
Tested the patch with my application and everything works now even when removing `->withoutPersisting()`.

Thank you so much for looking into this.

PR #274, opened 2022-04-28
Add note about phpstan docblocks

### LINKED ISSUE #272: Foundry and PHPStan errors
I have implemented Foundry to generate all the test fixtures for my app. It's working nicely - thank you.

However, when I run my static analysis with PHPStan, I see hundreds of errors such as:

```
Call to an undefined method App\Domain\Model\User|Zenstruck\Foundry\Proxy::getId()
```

This is because I am using the `getId()` method in my test - for example to build the route for a controller in a functional test.

I could of course add an annotation to make PHPStan happy -  e.g.

```
/** @var User $user */
$user = UserFactory::last();
```

However, this would have to be done many, many times and it does not seem like a smart way to deal with it.

Can you suggest any potential solutions? Am I missing something obvious? Thanks.

### LINKED ISSUE #273: "Short" version of PHPDoc annotations does not allow PHPStorm to auto-complete
When creating a factory using the MakerBundle command, or copying from the docs, you get the "long" version of the PHPDoc annotations.

I'm running the latest PHPStorm (2022.1), and the docs state that a "short" version of the annotations can be used if you are running PHPStorm 2021.2+.

However, this short version of the annotations does NOT give me auto-complete in PHPStorm (although PHPStan is happy).

So it seems the "short" annotations are not working as a replacement for the "long" version of the annotations.

### Example Long Version
```php
/**
 * @extends ModelFactory<User>
 *
 * @method static     User|Proxy createOne(array $attributes = [])
 * @method static     User[]|Proxy[] createMany(int $number, array|callable $attributes = [])
 * @method static     User|Proxy find(object|array|mixed $criteria)
 * @method static     User|Proxy findOrCreate(array $attributes)
 * @method static     User|Proxy first(string $sortedField = 'id')
 * @method static     User|Proxy last(string $sortedField = 'id')
 * @method static     User|Proxy random(array $attributes = [])
 * @method static     User|Proxy randomOrCreate(array $attributes = [])
 * @method static     User[]|Proxy[] all()
 * @method static     User[]|Proxy[] findBy(array $attributes)
 * @method static     User[]|Proxy[] randomSet(int $number, array $attributes = [])
 * @method static     User[]|Proxy[] randomRange(int $min, int $max, array $attributes = [])
 * @method User|Proxy create(array|callable $attributes = [])
 */
```

### Example Short Version
```php
/**
 * @extends ModelFactory<User>
 *
 * @method static User[]|Proxy[] createMany(int $number, array|callable $attributes = [])
 * @method static UserRepository|RepositoryProxy repository()
 * @method User|Proxy create(array|callable $attributes = [])
 */
```

kbond:
Turns out with the latest PhpStorm we can't support autocompletion and PHPStan with one set of `@method` docblocks.

Close #273, Closes #272.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -206,19 +206,19 @@ This command will generate a ``PostFactory`` class that looks like this:
         /**
          * ...
          *
-         * @phpstan-method static User&Proxy createOne(array $attributes = [])
-         * @phpstan-method static User[]&Proxy[] createMany(int $number, array|callable $attributes = [])
-         * @phpstan-method static User&Proxy find(object|array|mixed $criteria)
-         * @phpstan-method static User&Proxy findOrCreate(array $attributes)
-         * @phpstan-method static User&Proxy first(string $sortedField = 'id')
-         * @phpstan-method static User&Proxy last(string $sortedField = 'id')
-         * @phpstan-method static User&Proxy random(array $attributes = [])
-         * @phpstan-method static User&Proxy randomOrCreate(array $attributes = [])
-         * @phpstan-method static User[]&Proxy[] all()
-         * @phpstan-method static User[]&Proxy[] findBy(array $attributes)
-         * @phpstan-method static User[]&Proxy[] randomSet(int $number, array $attributes = [])
-         * @phpstan-method static User[]&Proxy[] randomRange(int $min, int $max, array $attributes = [])
-         * @phpstan-method User&Proxy create(array|callable $attributes = [])
+         * @phpstan-method static Post&Proxy createOne(array $attributes = [])
+         * @phpstan-method static Post[]&Proxy[] createMany(int $number, array|callable $attributes = [])
+         * @phpstan-method static Post&Proxy find(object|array|mixed $criteria)
+         * @phpstan-method static Post&Proxy findOrCreate(array $attributes)
+         * @phpstan-method static Post&Proxy first(string $sortedField = 'id')
+         * @phpstan-method static Post&Proxy last(string $sortedField = 'id')
+         * @phpstan-method static Post&Proxy random(array $attributes = [])
+         * @phpstan-method static Post&Proxy randomOrCreate(array $attributes = [])
+         * @phpstan-method static Post[]&Proxy[] all()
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -195,21 +195,30 @@ This command will generate a ``PostFactory`` class that looks like this:
...
+         * @phpstan-method static User&Proxy findOrCreate(array $attributes)
+         * @phpstan-method static User&Proxy first(string $sortedField = 'id')
+         * @phpstan-method static User&Proxy last(string $sortedField = 'id')
+         * @phpstan-method static User&Proxy random(array $attributes = [])
+         * @phpstan-method static User&Proxy randomOrCreate(array $attributes = [])
+         * @phpstan-method static User[]&Proxy[] all()
+         * @phpstan-method static User[]&Proxy[] findBy(array $attributes)
+         * @phpstan-method static User[]&Proxy[] randomSet(int $number, array $attributes = [])
+         * @phpstan-method static User[]&Proxy[] randomRange(int $min, int $max, array $attributes = [])
+         * @phpstan-method User&Proxy create(array|callable $attributes = [])
```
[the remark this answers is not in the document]
kbond: Yes, thank you.

docs/index.rst
```diff
@@ -195,21 +195,30 @@ This command will generate a ``PostFactory`` class that looks like this:
...
 
-.. tip::
+.. note::
 
-    PhpStorm 2021.2+ has support for
-    `generics annotations <https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-release/#generics>`_,
-    with it, your factory's annotations can be reduced to the following and still have the same auto-completion
-    support:
+    The generated ``@method`` docblocks above enable autocompletion with PhpStorm but
+    cause errors with PHPStan. To support PHPStan for your factory's, you need to *also*
```
benr77, line 201: Should be "factories" in the plural

benr77: Actually, no, I think you are right - you are referring to a specific example.

benr77 (approved):

PR #277, opened 2022-05-23
[feature] Handle variadic constructor arguments

ndench:
When instantiating an object that has variadic constructor arguments, we get an error like the following:

```
TypeError : Foo::__construct(): Argument #3 must be of type string, array given
```

We can use the Reflection API to identify if a parameter is variadic so that we can pass it in correctly.


kbond:
This is great, thanks @ndench!

PR #283, opened 2022-06-22
Use DateTimeImmutable with immutable ORM types

HypeMC:
Immutable Doctrine types should use `\DateTimeImmutable` instead of `\DateTime`.


kbond:
Thanks @HypeMC!

Issue #288, opened 2022-07-21
Strange error message about deprecation.


gazzatav:
I'm working through the SymfonyCasts tutorials for Symfony 5 and have v1.21.0 installed.  When I create a factory and add state like this:

`QuestionFactory::new()->unpublished()->createMany(5);`

I get this weird message (when loading fixtures).  I wouldn't mind if what it told me to do actually worked but the instruction is to do exactly the same as what has been deprecated (except it doesn't work because of the single colon).

> "User Deprecated: Since zenstruck/foundry 1.7: Calling instance method "App\Factory\QuestionFactory::createMany()" is deprecated and will be removed in 2.0, use the static "App\Factory\QuestionFactory:createMany()" method instead.

I realise this may be out of date by now  but would appreciate an explanation and/or suggestion as to how I can make Factory.php happy again. 

EDIT: I see this is related to #115 and the error disappears after changing the line to:

`QuestionFactory::new()->unpublished()->many(5)->create();`

I still think the actual message could be improved!

kbond:
Yes, I agree, what about 

> Calling instance method "App\Factory\QuestionFactory::createMany(x)" is deprecated and will be removed in 2.0, use "$factory->many(x)->create()" instead.

gazzatav:
Yes, `$factory->many(x)->create()"` would have been helpful to me.  It is the replacement for the function called rather than a different function, which cannot be used in the same situation.  How much of the error is programmatically generated though?  I notice that it names the user class in the error.

kbond:
> How much of the error is programmatically generated though?

You can see where the error is generated here: https://github.com/zenstruck/foundry/blob/a535f83ef65680a63f615d64b04219cdeee35f1b/src/Factory.php#L69

> I notice that it names the user class in the error.

Yes, I use `static::class` to get the name of your `ModelFactory` class.

---

Would you be willing to create a PR?

gazzatav:
Yes, no problem.  I'm not a PHP expert but it looks like it's just that string that changes.   Are you happy to have a longer, more comprehensive message, or would you prefer to keep it to the above i.e. `$factory->many(x)->create()`?

kbond:
Let's just change it exactly what I have above. The current message is confusing as you've found and doesn't show the proper upgrade path.

gazzatav:
I was just thinking that as you have the static class there and as all the comments I've seen on this were people in the same boat as me, the 'what to do' part of the message could be:

[...]

PR #289, opened 2022-07-22
Clarifying deprecation of non-static createMany()

weaverryan:
If you're using, for example, a state method, then you want to instantiate the factory, call `->many()` and then `->create()`

Based on feedback over at - https://symfonycasts.com/screencast/symfony-doctrine/foundry-tricks#comment-5925292669 - from our deprecated code usage there :).

Cheers!


kbond:
Hah, had a conversation with @gazzatav about this in #288.

kbond:
Thanks Ryan, closing in favor of #291.

PR #291, opened 2022-07-22
Improve deprecation message for createMany().

### LINKED ISSUE #288: Strange error message about deprecation.

gazzatav:
#288 Message now suggests to use the decomposed functions many() and create() instead of createMany() and gives an example showing how this would look with a common use case, that of adding state.

Fixes #288.

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -66,7 +66,7 @@ class Factory
             throw new \BadMethodCallException(\sprintf('Call to undefined method "%s::%s".', static::class, $name));
         }
 
-        trigger_deprecation('zenstruck/foundry', '1.7', 'Calling instance method "%1$s::createMany()" is deprecated and will be removed in 2.0, use e.g. "%1$s::new()->stateAdapter()->many()->create()" instead.', static::class, $arguments[0]);
+        trigger_deprecation('zenstruck/foundry', '1.7', 'Calling instance method "%1$s::createMany()" is deprecated and will be removed in 2.0, use e.g. "%1$s::new()->stateAdapter()->many(%2$d)->create()" instead.', static::class, $arguments[0]);
 
         return $this->many($arguments[0])->create($arguments[1] ?? []);
     }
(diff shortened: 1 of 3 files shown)


kbond:
Thanks @gazzatav!

gazzatav:
Thanks.  A first for me, although I've used Git for years I've never done a pull request on github. Was there anything I could have done better?  When the tests failed or timed out I wasn't sure what to do, was there anything I could have done?  I noticed you changed a unit test concerning PostFactory.

kbond:
Nice!

> When the tests failed or timed out I wasn't sure what to do, was there anything I could have done?

A couple of the tests checked the text of the deprecation message, that's why the tests were failing.

Here's where you could have seen that: https://github.com/zenstruck/foundry/runs/7476656717?check_suite_focus=true#step:7:133

Admittedly, the failure message was a bit confusing...

> Was there anything I could have done better?

I guess you could have run the test suite locally to find/fix the failure but running it locally is a pain right now. It's on my to-do list to improve.

I think you did great!

PR #294, opened 2022-08-27
Use field limit length when specified

### LINKED ISSUE #286: [Make:factory] Use field limit length when specified
Hi, 

Command `make:factory` create Factory with default Faker method according to field type, like `self::faker()->text()` for a `string` field

What would you think about use `limit` property of fields and in order to generate more accurate Factory ? 

For example, this field

```
#[ORM\Column(type: 'string', length: 40)] 
private string $label;
```

could be rendered this way in Factory

```'label' => self::faker()->text(40)```

I think i can work on a PR for this if you agree with this idea

MrYamous:
Related to issue #286

I think this will require new tests, and maybe adapting existing ones, but I haven't managed to run them locally yet. Still working on how to do it

### DIFF
--- a/src/Bundle/Maker/MakeFactory.php
+++ b/src/Bundle/Maker/MakeFactory.php
@@ -26,7 +26,7 @@ final class MakeFactory extends AbstractMaker
 {
     private const DEFAULTS = [
         'ARRAY' => '[],',
-        'ASCII_STRING' => 'self::faker()->text(),',
+        'ASCII_STRING' => 'self::faker()->text({length}),',
         'BIGINT' => 'self::faker()->randomNumber(),',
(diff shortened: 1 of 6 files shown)


kbond (dismissed):
> I think this will require new tests, and maybe adapting existing ones, but I haven't managed to run them locally yet. Still working on how to do it

Sorry about that, the test suite for this library is a bear. I think if you find/replace all instances of `self::faker()->text()` with `self::faker()->text(255)` in `MakeFactoryTest` that will make the test suite pass. Don't think we need new tests.

nikophil (approved):

kbond:
Thanks @MrYamous!

PR #295, opened 2022-08-31
Update index.rst

Chris53897:
PhpStorm did not find this method. And i do not see it in the proxy-class.


kbond (approved):
Good catch, thanks!

PR #298, opened 2022-09-14
[feat] introduce sequences

### LINKED ISSUE #177: Factory collection "sequences"
https://laravel.com/docs/8.x/database-testing#sequences

nikophil:
fixes #177

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -148,6 +148,7 @@ This command will generate a ``PostFactory`` class that looks like this:
      *
      * @method static Post|Proxy createOne(array $attributes = [])
      * @method static Post[]|Proxy[] createMany(int $number, array|callable $attributes = [])
+     * @method static Post[]&Proxy[] createSequence(array|callable $sequence)
      * @method static Post|Proxy find(object|array|mixed $criteria)
      * @method static Post|Proxy findOrCreate(array $attributes)
      * @method static Post|Proxy first(string $sortedField = 'id')
@@ -208,6 +209,7 @@ This command will generate a ``PostFactory`` class that looks like this:
          *
          * @phpstan-method static Post&Proxy createOne(array $attributes = [])
          * @phpstan-method static Post[]&Proxy[] createMany(int $number, array|callable $attributes = [])
+         * @phpstan-method static Post[]&Proxy[] createSequence(array|callable $sequence)
          * @phpstan-method static Post&Proxy find(object|array|mixed $criteria)
          * @phpstan-method static Post&Proxy findOrCreate(array $attributes)
          * @phpstan-method static Post&Proxy first(string $sortedField = 'id')
@@ -417,6 +419,43 @@ instantiation.
 
     Attributes passed to the ``create*`` methods are merged with any attributes set via ``getDefaults()``
     and ``withAttributes()``.
+
+
+Sequences
+~~~~~~~~~
+
+Sequences help to create different objects in one call:
+
+.. code-block:: php
+
+    use App\Factory\PostFactory;
+
+    // create/persist 2 posts based on a sequence of attributes
+    PostFactory::createSequence(
+        [
+            ['name' => 'title 1'],
+            ['name' => 'title 2'],
+        ]
+    );
+
+    // create 10 posts using a sequence callback with an incremental index
+    PostFactory::createSequence(
+        function() {
+            foreach (range(1, 10) as $i) {
+                yield ['name' => "title $i"];
+            }
+        }
+    );
+
+    // sequences could also be used with a factory with states
+    $posts = PostFactory::new()
+        ->unpublished()
+        ->sequence(
+            [
+                ['name' => 'title 1'],
+                ['name' => 'title 2'],
+            ]
+        )->create();
 
 Faker
 ~~~~~
@@ -1405,6 +1444,72 @@ It is possible to use factories in
 
     For the same reason as above, it is not possible to use `Factories as Services`_ with required
     constructor arguments (the container is not yet available).
+
+.. tip::
+
+    ``ModelFactory::new()->many()`` and ``ModelFactory::new()->sequence()`` return a special ``FactoryCollection`` object
+    which can be used to generate data providers:
+
+    .. code-block:: php
+
+        use App\Factory\PostFactory;
+
+        /**
+         * @dataProvider postDataProvider
+         */
+        public function test_post_via_data_provider(PostFactory $factory): void
+        {
+            $factory->create();
+
+            // ...
+        }
+
+        public static function postDataProvider(): iterable
+        {
+            yield from PostFactory::new()->sequence(
+                [
+                    ['title' => 'foo'],
+                    ['title' => 'bar'],
+                ]
+            )->asDataProvider();
+        }
+
+    The ``FactoryCollection`` could also be passed directly to the test case in order to have several objects available in the same test:
+
+    .. code-block:: php
+
+        use App\Factory\PostFactory;
+
+        /**
+         * @dataProvider postDataProvider
+         */
+        public function test_post_via_data_provider(FactoryCollection $factoryCollection): void
+        {
+            $factoryCollection->create();
+
+            // ...
+        }
+
+        public static function postDataProvider(): iterable
+        {
+            // 3 posts will be created for the first test case
+            yield PostFactory::new()->sequence(
+                [
+                    ['title' => 'foo 1'],
+                    ['title' => 'bar 1'],
+                    ['title' => 'baz 1'],
+                ]
+            );
+
+            // 2 posts will be created for the second test case
+            yield PostFactory::new()->sequence(
+                [
+                    ['title' => 'foo 2'],
+                    ['title' => 'bar 2'],
+                ]
+            );
+        }
+
 
 Performance
 ~~~~~~~~~~~
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -131,13 +131,31 @@ class Factory
     }
 
     /**
-     * @see FactoryCollection::__construct()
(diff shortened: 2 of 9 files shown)


src/Factory.php
```diff
@@ -140,6 +140,26 @@ final public function many(int $min, ?int $max = null): FactoryCollection
         return new FactoryCollection($this, $min, $max);
     }
 
+    /**
+     * @param list<array<string, mixed>>|callable(): iterable<array<string, mixed>> $sequence
```
kbond, line 144:
```suggestion
     * @param iterable<array<string, mixed>>|callable(): iterable<array<string, mixed>> $sequence
```

?

docs/index.rst
```diff
@@ -276,6 +276,21 @@ Using your Factory
     PostFactory::createMany(5); // returns Post[]|Proxy[]
     PostFactory::createMany(5, ['title' => 'My Title']);
 
+    // create/persist several posts based on a sequence of attributes
```
kbond, line 279:
```suggestion
    // create/persist 2 posts based on a sequence of attributes
```

docs/index.rst
```diff
@@ -276,6 +276,21 @@ Using your Factory
...
+    Post::sequence(
```
kbond, line 280:
```suggestion
    Post::createSequence(
```

docs/index.rst
```diff
@@ -276,6 +276,21 @@ Using your Factory
...
+        [
+            ['name' => 'title 1'],
+            ['name' => 'title 2'],
+        ]
+    );
+    Post::sequence(
```
kbond, line 286:
```suggestion

    // create 10 posts using a sequence callback with an incrementing index
    Post::createSequence(
```

docs/index.rst
```diff
@@ -276,6 +276,21 @@ Using your Factory
...
+        function() {
+            foreach (range(0, 9) as $i) {
```
kbond, line 288:
```suggestion
            foreach (range(1, 10) as $i) {
```

kbond (changes requested):
Looks great, a few minor things.

Regarding the docs, I was trying to find a place where we could document `PostFactory::new()->sequence()` but couldn't find a great spot. Any ideas?

nikophil:
I've made another documentation proposal

src/ModelFactory.php
```diff
@@ -74,6 +74,19 @@ final public static function createOne(array $attributes = []): Proxy
         return static::new()->create($attributes);
     }
 
+    /**
+     * A shortcut to create multiple models, based on a sequence, without states.
+     *
+     * @param iterable<array<string, mixed>>|callable(): iterable<array<string, mixed>> $sequence
+     *
+     * @return list<Proxy<TModel>&TModel>
```
kbond, line 82:
```suggestion
     * @return list<TModel&Proxy<TModel>>
```

Maybe reversing these will fix the cs-fixer issue/bug? `randomSet()` below has it like this with no issue.

docs/index.rst
```diff
@@ -1406,6 +1444,41 @@ It is possible to use factories in
...
+        yield from PostFactory::new()->sequence(
+            [
+                ['title' => 'foo'],
+                ['title' => 'bar'],
+            ]
+        )->dataProvider(
+            [
+                ['foo'],
+                ['bar'],
+            ]
```
nikophil, line 1477: not sure about this, since it can create data providers difficult to reason about...

docs/index.rst
```diff
@@ -1406,6 +1444,67 @@ It is possible to use factories in
...
+            yield from PostFactory::new()->sequence(
+                [
+                    ['title' => 'foo'],
+                    ['title' => 'bar'],
+                ]
+            )->dataProvider(
+                [
+                    ['foo'],
+                    ['bar'],
+                ]
```
nikophil, line 1477: not sure about this, since it can create data providers difficult to reason about...

docs/index.rst
```diff
@@ -208,6 +208,7 @@ This command will generate a ``PostFactory`` class that looks like this:
          *
          * @phpstan-method static Post&Proxy createOne(array $attributes = [])
          * @phpstan-method static Post[]&Proxy[] createMany(int $number, array|callable $attributes = [])
+         * @phpstan-method static Post[]&Proxy[] createSequence(array|callable $sequence)
```
kbond, line 211: Let's add to the docblock above as well.

src/FactoryCollection.php
```diff
@@ -18,12 +18,18 @@ final class FactoryCollection
...
-     * @param int|null $max If set, when created, the collection will be a random size between $min and $max
+     * @param int|null                       $max      If set, when created, the collection will be a random size between $min and $max
+     * @param iterable<array<string, mixed>> $sequence
      *
      * @psalm-param Factory<TObject> $factory
+     *
+     * @deprecated using directly FactoryCollection's constructor is deprecated. It will be private in v2. Use named constructors instead.
      */
-    public function __construct(Factory $factory, int $min, ?int $max = null)
+    public function __construct(Factory $factory, ?int $min = null, ?int $max = null, ?iterable $sequence = null)
```
kbond, line 32:
```suggestion
    public function __construct(Factory $factory, ?int $min = null, ?int $max = null, ?iterable $sequence = null, bool $calledInternally = false)
```

Then trigger a deprecation if `$calledInternally = false`

src/FactoryCollection.php
```diff
@@ -32,6 +38,25 @@ public function __construct(Factory $factory, int $min, ?int $max = null)
         $this->factory = $factory;
         $this->min = $min;
         $this->max = $max ?? $min;
+        $this->sequence = $sequence;
+    }
+
+    public static function set(Factory $factory, int $count): self
+    {
+        return new self($factory, $count);
```
kbond, line 46:
```suggestion
        return new self($factory, $count, null, null, true);
```

src/FactoryCollection.php
```diff
@@ -32,6 +38,25 @@ public function __construct(Factory $factory, int $min, ?int $max = null)
...
+    }
+
+    public static function range(Factory $factory, int $min, int $max): self
+    {
+        return new self($factory, $min, $max);
```
kbond, line 51:
```suggestion
        return new self($factory, $min, $max, null, true);
```

src/FactoryCollection.php
```diff
@@ -32,6 +38,25 @@ public function __construct(Factory $factory, int $min, ?int $max = null)
...
+    }
+
+    /**
+     * @param iterable<array<string, mixed>> $sequence
+     */
+    public static function sequence(Factory $factory, iterable $sequence): self
+    {
+        return new self($factory, 0, null, $sequence);
```
kbond, line 59:
```suggestion
        return new self($factory, 0, null, $sequence, true);
```

src/Factory.php
```diff
@@ -131,13 +131,29 @@ function($value) {
     }
 
     /**
-     * @see FactoryCollection::__construct()
-     *
      * @return FactoryCollection<TObject>
      */
     final public function many(int $min, ?int $max = null): FactoryCollection
```
kbond, line 136: Let's put the docblocks from `FactoryCollection::__construct()` here.

kbond (changes requested):

docs/index.rst
kbond: Yeah, you're right. Let's drop the `dataProvider()` for now and in the docs, show how to manually create a provider:

```php
        public static function postDataProvider(): iterable
        {
            $sequence = PostFactory::new()->sequence(
                [
                    ['title' => 'foo'],
                    ['title' => 'bar'],
                ]
            );

            foreach ($sequence as $factory) {
                yield [$factory];
            }
        }
```

src/FactoryCollection.php
nikophil: ok, that's clever, I was actually thinking the annotation was enough

kbond (changes requested):
Just a few minor things.

docs/index.rst
```diff
@@ -1406,6 +1445,66 @@ It is possible to use factories in
...
+        public function test_post_via_data_provider(FactoryCollection $factoryCollection): void
+        {
+            $factoryCollection->create();
+
+            // ...
+        }
+
+        public static function postDataProvider(): iterable
+        {
+            yield PostFactory::new()->sequence(
```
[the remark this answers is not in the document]
nikophil: I've added a second yield, and some comments

docs/index.rst
```diff
@@ -1406,6 +1445,76 @@ It is possible to use factories in
...
+            $sequence = PostFactory::new()->sequence(
+                [
+                    ['title' => 'foo'],
+                    ['title' => 'bar'],
+                ]
+            );
+
+            foreach ($sequence as $factory) {
+                yield [$factory];
+            }
```
kbond, line 1478: How common do you think this pattern would be? Just returning a single factory (wrapped in an array) from a data provider? I think the `dataProvider()` method we had before was confusing because of the extra parameters it takes, but what about:

```php
            yield from PostFactory::new()->sequence(
                [
                    ['title' => 'foo'],
                    ['title' => 'bar'],
                ]
            )->asDataProvider();
```

As a shortcut to this block?

nikophil: but I can definitely introduce the `asDataProvider()` method (I prefer this new name, btw)

kbond: Ok, let's do this.

nikophil: done :heavy_check_mark:

nikophil: done :heavy_check_mark:

kbond (approved):

nikophil:
:tada:

kbond:
Thanks @nikophil! 🎉

PR #300, opened 2022-09-20
fix: discover relations with inheritance

NorthBlue333:
Hi,
I have stumbled upon this bug.

In my entities, I have a "polymorphic" relationship which is typed with an Abstract class, implemented by 2 entities. The package is not able to detect their relation and normalize the factories. This fixes it.

I have not added tests because it needs changes in entities and I do not know how you wish to implement it.

```php
abstract class CourseLocation {
#[ORM\OneToMany(mappedBy: 'courseLocation', targetEntity: CourseLocationContact::class, cascade: ['persist'], orphanRemoval: true)]
    protected Collection $contacts;

    /**
     * @return Collection|CourseLocationContact[]
     */
    public function getContacts(): array|Collection
    {
        return $this->contacts;
    }

    public function addContact(CourseLocationContact $contact): static
    {
        if (!$this->contacts->contains($contact)) {
            $this->contacts->add($contact);
            $contact->setCourseLocation($this);
        }

        return $this;
    }

    public function removeContact(CourseLocationContact $contact): static
    {
        if ($this->contacts->removeElement($contact)) {
            if ($contact->getCourseLocation() === $this) {
                $contact->setCourseLocation(null);
            }
        }

        return $this;
    }
}

class TrainingCenter extends CourseLocation {}

class TemporarySite extends CourseLocation {}

class CourseLocationContact
{
    #[ORM\ManyToOne(targetEntity: CourseLocation::class, inversedBy: 'contacts')]
    private ?CourseLocation $courseLocation;

    public function getCourseLocation(): ?CourseLocation
    {
        return $this->courseLocation;
    }

    public function setCourseLocation(?CourseLocation $courseLocation): self
    {
        $this->courseLocation = $courseLocation;

        return $this;
    }
}
```

Acutal error : 
> A new entity was found through the relationship 'App\Entity\CourseLocationContact#courseLocation' that was not configured to cascade persist operations for entity: TrainingCenter. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example `@ManyToOne(..,cascade={"persist"})`. 

---
Below will be addressed is addressed in #302.

As a side note, I dont think the actual code (nothing to do with the fix) can work with this situation, as I think it will arbitrary choose the first relation with corresponding type (here `User`) :
```php
class Post {
    private User $author;
    private User $reader;
}

class User {
    private Collection $authoredPosts;
    private Collection $readerPosts;
}

UserFactory::new(['readerPosts' => PostFactory::new()->many(5)]);
```

Is it possible to **also** check the original relation name ?

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -15,6 +15,8 @@ use Faker;
  */
 class Factory
 {
+    private const NULL_VALUE = '__null_value';
+
     /** @var Configuration|null */
     private static $configuration;
 
@@ -96,7 +98,11 @@ class Factory
         // filter each attribute to convert proxies and factories to objects
         $mappedAttributes = [];
         foreach ($attributes as $name => $value) {
-            $mappedAttributes[$name] = $this->normalizeAttribute($value, $name);
+            $normalizedAttribute = $this->normalizeAttribute($value, $name);
+            if (self::NULL_VALUE === $normalizedAttribute) {
+                $normalizedAttribute = null;
+            }
+            $mappedAttributes[$name] = $normalizedAttribute;
         }
         $attributes = $mappedAttributes;
 
@@ -321,11 +327,16 @@ class Factory
 
         if (\is_array($value)) {
             // possible OneToMany/ManyToMany relationship
-            return \array_map(
+            return \array_filter(
+                \array_map(
+                    function($value) use ($name) {
+                        return $this->normalizeAttribute($value, $name);
+                    },
+                    $value
+                ),
                 function($value) {
-                    return $this->normalizeAttribute($value);
-                },
-                $value
+                    return self::NULL_VALUE !== $value;
+                }
             );
         }
 
@@ -340,8 +351,27 @@ class Factory
 
         // Check if the attribute is cascade persist
         if (self::configuration()->hasManagerRegistry()) {
-            $relationField = $this->relationshipField($name, $value);
-            $value->cascadePersist = $this->hasCascadePersist($value, $relationField);
+            $ownedRelationshipField = $this->ownedRelationshipField($name, $value);
+
+            if (null !== $ownedRelationshipField) {
+                $cascadePersist = $this->hasCascadePersist(null, $ownedRelationshipField);
+            } else {
+                $isCollection = false;
+                $relationshipField = $this->inverseRelationshipField($name, $value, $isCollection);
+                $cascadePersist = $this->hasCascadePersist($value, $relationshipField);
+
+                if ($this->isPersisting() && null !== $relationshipField && false === $cascadePersist) {
+                    $this->afterPersist[] = static function(Proxy $proxy) use ($value, $relationshipField, $isCollection) {
+                        $value->create([$relationshipField => $isCollection ? [$proxy] : $proxy]);
+                        $proxy->refresh();
+                    };
+
+                    // creation delegated to afterPersist event - return null here
+                    return self::NULL_VALUE;
+                }
+            }
+
+            $value->cascadePersist = $cascadePersist;
         }
 
         return $value->create()->object();
@@ -356,10 +386,10 @@ class Factory
         }
     }
 
-    private function normalizeCollection(?string $fieldName, FactoryCollection $collection): array
+    private function normalizeCollection(?string $relationName, FactoryCollection $collection): array
     {
         if ($this->isPersisting()) {
-            $field = $this->inverseRelationshipField($fieldName, $collection->factory());
+            $field = $this->inverseCollectionRelationshipField($relationName, $collection->factory());
             $cascadePersist = $this->hasCascadePersist($collection->factory(), $field);
 
             if ($field && false === $cascadePersist) {
@@ -383,7 +413,7 @@ class Factory
         );
     }
 
-    private function relationshipField(?string $relationName, self $factory): ?string
+    private function ownedRelationshipField(?string $relationName, self $factory): ?string
     {
         $factoryClass = $this->class;
         $relationClass = $factory->class;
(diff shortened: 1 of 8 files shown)


nikophil (commented):
:+1: for the fix, but maybe we can reproduce this kind of relationship in tests in order to assert everything is OK? WDYT @kbond ?

kbond:
Thanks for fixing this @NorthBlue333!

To clarify, you are describing two separate issues here?
1. A polymorphic relationship with a mapped super class? This PR fixes this?
2. Creating a fixture for an entity with two of the same relationships but with different property names doesn't set on the correct property?

> but maybe we can reproduce this kind of relationship in tests in order to assert everything is OK?

Yeah, definitely would like some tests but I know this library's test suite is complex and adding new entities even more so... @NorthBlue333, could you create the test entities that demonstrate the two above issues (as minimal as possible)? Could you also create two tests that show specifically the problems? Don't worry about making them pass - you can even use pseudo code. I'll work on making them pass.

@nikophil, these two problems valid for Mongo as well?

NorthBlue333:
Sure will do the tests! I'll try to do it before the end of the week.

Yes I am describing two issues.
About the other issues, I will create a separate PR/issue. My bad there, I was in the middle of my work when I wrote the PR. Sorry!

nikophil:
I've just discovered that [reference mapping](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.3/reference/reference-mapping.html) exists in doctrine

these are not (yet) supported in foundry. But once they'd be supported I think the use case could exist

kbond:
Hey @NorthBlue333, let me know when you've gone as far as you can, then I can fix up the tests.

tests/Functional/FactoryTest.php
```diff
@@ -34,8 +34,8 @@ public function many_to_one_relationship(): void
     {
         $categoryFactory = factory(Category::class, ['name' => 'foo']);
         $category = create(Category::class, ['name' => 'bar']);
-        $postA = create(Post::class, ['title' => 'title', 'body' => 'body', 'category' => $categoryFactory]);
-        $postB = create(Post::class, ['title' => 'title', 'body' => 'body', 'category' => $category]);
+        $postA = create(Post::class, ['type' => 'simple', 'title' => 'title', 'body' => 'body', 'category' => $categoryFactory]);
```
kbond, line 37: This shouldn't be required. The discriminator column is internal to doctrine.

NorthBlue333: Sure is, but I got a strange error in the tests. Gonna check that. I will update when I'm done with my changes and will state my changes :)

NorthBlue333:
Ok, I've figured out the issue, the migration is missing. May I let you create it @kbond ?

I have added the necessary tests for the ORM.

```
one_to_many_with_nested_collection_relationship_polymorphic
one_to_many_with_nested_collection_relationship_polymorphic_mixed
many_to_one_relationship_polymorphic
many_to_many_with_nested_collection_relationship_polymorphic
many_to_many_with_nested_collection_relationship_polymorphic_mixed
inverse_many_to_many_with_nested_collection_relationship_polymorphic
inverse_many_to_many_with_nested_collection_relationship_polymorphic_mixed
```

It would be really useful to have a PHP environment in the docker-compose as well. The tests need specific PHP extensions.

NorthBlue333:
> Ok, I've figured out the issue, the migration is missing. May I let you create it @kbond ?

Nevermind, I have managed to add it.

kbond:
I don't think the tests actually are covering this fix. I pushed a temporary commit to remove the fix and all the tests still pass.

NorthBlue333:
Oh you're right. Gonna fix this right now.

NorthBlue333:
I have added a new part of the fix, concerning creating entities via `OneToMany` without `cascade: persist`.

I have added the following tests and kept the others (in order to always have them succeed if there are changes):

```php
one_to_many_relationship_without_cascade # fails
one_to_one_relationship_polymorphic # already works
inverse_one_to_one_relationship_polymorphic # already works but partially, see below
one_to_many_polymorphic_with_nested_collection_relationship # fails
```

`inverse_one_to_one_relationship_polymorphic` already works but only because it takes the wrong relation due to #302

Needs #302 for all tests to be ok.

NorthBlue333:
> So if we re-add the fix, the test suite will pass?
> 
> > Needs #302 for all tests to be ok.
> 
> Should we merge that one first?

Yes all tests should pass with the merge of #302 first

kbond:
Ready to be rebased

NorthBlue333:
Doing it!

NorthBlue333:
Adding #302 made some tests succeed with updates to `inverseRelationshipField` and usage of `is_a`, but the fix is still necessary.

I have not added more tests as it concerns creating a factory with a non-existing relation name, in example:

```php
class Comment {
User $user;
// User $secondUser;
}

class User {
// no $comments property defined
}

UserFactory::createOne([
// works because it finds the `user` inverse relationship, but would fail if `secondUser` is the targeted property, as we cannot determine inversedBy field.
  'comments' => CommentFactory::new()->many(4)
])
```

This is the original behavior before both this PR and #302. I do not really know if this is intended, and there are no tests asserting this behavior. @kbond Could you state if these tests are necessary ?

kbond:
> I do not really know if this is intended, and there are no tests asserting this behavior. Could you state if these tests are necessary?

Hmm, this was not intended but let's keep it out of the scope of this PR.

NorthBlue333:
Just gonna fix the tests (must have done a bad merge).

kbond:
Looks like we're all good here. Ok to merge after the final action completes?

Question: did you ever get the tests running locally?

NorthBlue333:
Looks ok to me!

> Question: did you ever get the tests running locally?

No I did not. Sorry for that, but did not have enough time to setup an env with correct versions of PHP and extensions.
I could have though, needs a custom Dockerfile for PHP and a compose with all needed services. I recall that when trying, there was also some packages missing event after `composer install`, but that must be an error on my side.

kbond:
> Sorry for that, but did not have enough time to setup an env with correct versions of PHP and extensions.

No worries! We really need #301!

kbond:
Thanks so much for your patience here @NorthBlue333!

Issue #222, opened 2021-12-11
docs: Improve initial Setup to contribute on this Bundle


benblub:
Maybe its a good idea so write somewhere how to setup/use this repo in an easy way to prevent searching all whats needed.

### Running tests
locally i installed phpunit instead of running `php vendor/bin/simple-phpunit`

Problem
Step by step Debugging in PHPStorm IDE which uses per default normal phpunit and its more handy then using the CLI.

Solution
composer require phpunit/phpunit --dev
Then you can run `php vendor/bin/phpunit` which also fixes running unit tests in PHPStorm via the small icons

### Setup
Problem
Needs time to setup this here to start contribute

Solution
Add some dockerized files that we can fast start working.

### CSFixer
Problem
Missing binary in bin folder

Solution
composer require friendsofphp/php-cs-fixer --dev
then you can run php vendor/bin/php-cs-fixer fix ..

### Running migration
Problem
How to generate new migrations?

Solution
Not found yet
run php vendor/bin/doctrine-migrations migrations:generate didnt work -> missing configuration.

**edit**
I have to closer look into this issue, first steps works now 💪 
![Bildschirmfoto 2021-12-12 um 00 30 40](https://user-images.githubusercontent.com/1756031/145694818-67eed60d-d1b5-4d2b-9de4-1231a9bca91f.png)

kbond:
Thanks for jotting down your issues. Yes, I'd like to add a _**contributing**_ section to the readme.

> Running tests

I'd like to maintain the bridge to avoid dependency conflicts. For PhpStorm, there is a solution: https://stackoverflow.com/questions/47569173/how-to-configure-phpstorm-to-use-symfony-phpunit-bridge (happy to include a link to this)

> Setup

I agree but have no/little experience with docker but create a PR with your setup and we can discuss further.

> CSFixer

I'm planning to remove this as a PR check and just use an action to auto-commit fixes after merge. This shouldn't be a roadblock to contributing.

> Running migration

I'd be happy to formalize these steps.

benblub:
For the CSFixer i like Husky. Then you can setup a git pre-commit hook which can be commited to the repo. But its required npm..

https://github.com/typicode/husky

kbond:
For docker, I feel something like https://github.com/nektos/act would be best (assuming it works as described). What would be nice about this solution is we wouldn't have to keep docker config and gh actions config in sync.

benblub:
needs to test.

nikophil:
[...]

PR #301, opened 2022-09-22
Make it easier to contribute

kbond:
Fixes https://github.com/zenstruck/foundry/issues/222

TODO:
- [ ] `test:full` working
- [ ] `build:fixcs` working
- [ ] `build:sca` working
- [ ] Add contributor section to readme


composer.json
```diff
@@ -63,9 +63,22 @@
...
-        }
+        "build:docs": [
+            "composer bin docs install",
+            "bin/tools/docs/vendor/symfony-tools/docs-builder/bin/console build:docs --fail-on-errors docs/ docs/output/"
+        ],
+        "build:fixcs": "echo TODO",
+        "build:sca": "echo TODO",
+        "test:full": [
+            "DATABASE_URL=mysql://root:root@127.0.0.1:33061/foundry MONGO_URL=mongodb://127.0.0.1:27017/dbName?compressors=disabled&amp;gssapiServiceName=mongodb USE_FOUNDRY_BUNDLE=1 vendor/bin/simple-phpunit --configuration phpunit-dama-doctrine.xml.dist"
+        ],
```
kbond, line 74: I'd like to be able to have this:

```suggestion
        "test:full": [
            "docker composer up -d",
            "(wait for mysql/mongo to be available??)"
            "DATABASE_URL=mysql://root:root@127.0.0.1:33061/foundry MONGO_URL=mongodb://127.0.0.1:27017/dbName?compressors=disabled&amp;gssapiServiceName=mongodb USE_FOUNDRY_BUNDLE=1 vendor/bin/simple-phpunit --configuration phpunit-dama-doctrine.xml.dist"
            "docker composer down",
        ],
```

wouterj: > "(wait for mysql/mongo to be available??)"

This can be achieved using [healthchecks](https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck) in the Docker compose.

kbond: Right, but I wasn't sure how to actually have docker `compose up -d` _pause_ until the health check passes. I was wondering if I could create a dummy service that `depends_on` `mysql` and `mongo`. Something like:

```yaml
services:
  dummy:
    depends_on:
      mysql:
        condition: service_healthy
      mongo:
        condition: service_healthy
```

NorthBlue333:
It would be really useful to have a PHP environment in the docker-compose as well. The tests need specific PHP extensions.

kbond:
> It would be really useful to have a PHP environment in the docker-compose as well. The tests need specific PHP extensions.

Good point! Same with sca/fixcs, they need specific PHP versions.

composer.json
nikophil: I'm actually wondering if we should not provide a php container. It has several benefits:
- It could be used to run the healthcheck
- it would be easier to contribute because all needed extensions will be already installed (mongo, pgsql, mysql...)
- we could fix the version for sttic analysis

WDYT?

kbond: I think this makes sense. It also would prevent users from using code that's incompatible with php 7.2.

@wouterj, wdyt?

PR #302, opened 2022-09-24
fix: multiple relationships with same entity

NorthBlue333:
Hello again,

I have confirmed there is an issue when an entity have two not owned relationships with the same target entity. (Mentionned it in #300).

```php
class Post {
    // many to one, owner, mapped by authoredPosts
    private User $author;
    // many to one, owner, mapped by readingPosts
    private User $reader;

    // one to one, owner, mapped by oneToOnePost
    private User $oneToOneUser;

    // many to many, owner, mapped by readPosts
    private Collection $readers;
    // many to many, owner, mapped by likedPosts
    private Collection $likers;
}

class User {
    private Collection $authoredPosts;
    private Collection $readingPosts;
    
    private Post $oneToOnePost;

    private Collection $readPosts;
    private Collection $likedPosts;
}

UserFactory::new([
  // will set author
  'readerPosts' => PostFactory::new()->many(5),
  // will ALSO set author
  'authoredPosts' => PostFactory::new()->many(5),
  // will ALSO set author
  'oneToOnePost' => PostFactory::new(),
  // will set readPosts
  'readPosts' => PostFactory::new()->many(5),
  // I think it will ALSO set readPosts but not sure there
  'likedPosts' => PostFactory::new()->many(5),
]);
```

I have implemented the corresponding tests and proposed a fix. The necessary migration for the tests is missing, as I would need #301 to have a correct environment. **EDIT** Nevermind, I have managed to add the migration.

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -94,12 +94,11 @@ class Factory
         }
 
         // filter each attribute to convert proxies and factories to objects
-        $attributes = \array_map(
-            function($value) {
-                return $this->normalizeAttribute($value);
-            },
-            $attributes
-        );
+        $mappedAttributes = [];
+        foreach ($attributes as $name => $value) {
+            $mappedAttributes[$name] = $this->normalizeAttribute($value, $name);
+        }
+        $attributes = $mappedAttributes;
 
         // instantiate the object with the users instantiator or if not set, the default instantiator
         $object = ($this->instantiator ?? self::configuration()->instantiator())($attributes, $this->class);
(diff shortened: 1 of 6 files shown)


kbond:
Hey @NorthBlue333, glad you got the migrations/test suite figured out. I hope to make it easier to contribute in the near future.

Regarding this PR, I pushed a temporary commit that removes the fix just to check if the tests cover it properly. `one_to_many_with_two_relationships_same_entity` is the only test that fails w/o the fix, is that expected?

kbond:
Ok, do you want to remove the temp commit then?

NorthBlue333:
Good for me!

kbond:
Thanks @NorthBlue333!!

Issue #305, opened 2022-09-29
[Help] How to create rich domain entities?


nikophil:
Hello,

that's not an issue, but more a request for help.

I have some "rich" domain entities (opposite of [anemic entities](https://thevaluable.dev/anemic-domain-model/)) and I struggle to create them with foundry.
Here is a simple example:
```php
// Post.php
#[ORM\Entity]
class Post
{
    #[ORM\Column]
    private string $title;

    #[ORM\OneToMany(targetEntity: Comment::class, mappedBy: 'post')]
    private iterable $comments;

    public function __construct(string $title);

    public function addComment(Comment $comment): void;

    public function getComments(): iterable<Comments>
}

#[ORM\Entity]
class Comment
{
    #[ORM\ManyToOne(targetEntity: Event::class, inversedBy: 'slots'), ORM\JoinColumn(nullable: false)]
    private Post $post;

    #[ORM\Column]
    private string $body;

    public function __construct(Post $post, string $body);
}
```

`Comment::$post` property can never be modified in the app, thus I have zero need to add a setter on this property. 

Is there a way to create a post with some comments with foundry without hacking the code?

thanks for your answer

kbond:
Hey!

Rich entities is something I've been trying to build myself. I don't like entities that are constructed in an invalid state and with unnecessary setters/getters.

With your example above, my first thought is using `Instantiator::alwaysForceProperties()->withoutConstructor()`. Have you looked at that?

I always use this in my apps. I tend to like the idea of creating fixtures based on their database values rather than having them perfectly created via constructor/methods. I know this perhaps is controversial but this _is_ how doctrine works when it hyrdates objects from the db.

nikophil:
thanks for you answer!

I think this may be the way to go, but... `Comment::$post` is mandatory is the db... so how could I tell foundry to hydrate it somehow?

kbond:
To clarify, you are tying to do the following?

```php
PostFactory::createOne([
    'title' => 'my title',
    'comments' => CommentFactory::new()->many(5),
]);
```

kbond:
I've confirmed this is, I think, a bug.

I've created [a reproducer](https://github.com/kbond/symfony-reproducer/tree/foundry-rich-entity) that demonstrates the bug ([commit](https://github.com/kbond/symfony-reproducer/commit/715ae2ee9d3e79dc4675e486d6523b10b6b160c0)).

[...]

PR #312, opened 2022-10-12
[feature] allow creating "rich domain models"

kbond:
Possible solution for #305. When forcing properties with the `Instantiator`, covert `array`'s to `ArrayCollection`'s if the typehint demands it.

Not a huge fan of this solution... I think I'd prefer `FactoryCollection` generate an `ArrayCollection` and then it can be set directly. I'll explore this option.

### DIFF
--- a/src/Instantiator.php
+++ b/src/Instantiator.php
@@ -11,6 +11,8 @@
 
 namespace Zenstruck\Foundry;
 
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
 use Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 use Symfony\Component\PropertyAccess\PropertyAccessor;
@@ -137,6 +139,19 @@ final class Instantiator
      */
     public static function forceSet(object $object, string $property, mixed $value): void
     {
+        $refProp = self::accessibleProperty($object, $property);
+
+        // if value is an array and the property expects a doctrine collection, convert to doctrine collection
+        if (\is_array($value) && $type = $refProp->getType()) {
+            foreach ($type instanceof \ReflectionNamedType ? [$type] : $type->getTypes() as $type) { // @phpstan-ignore-line
+                if (\is_a($type->getName(), Collection::class, true)) {
+                    $value = new ArrayCollection($value);
+
+                    break;
+                }
+            }
+        }
+
         self::accessibleProperty($object, $property)->setValue($object, $value);
     }
 
--- a/tests/Fixtures/Entity/RichComment.php
+++ b/tests/Fixtures/Entity/RichComment.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Zenstruck\Foundry\Tests\Fixtures\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ */
+class RichComment
+{
+    /**
+     * @ORM\Id
+     * @ORM\GeneratedValue
+     * @ORM\Column(type="integer")
+     */
+    private ?int $id = null;
+
+    /**
+     * @ORM\Column(type="text")
+     */
+    private string $body;
+
+    /**
+     * @ORM\ManyToOne(targetEntity=RichPost::class, inversedBy="comments")
+     * @ORM\JoinColumn(nullable=false, onDelete="CASCADE")
+     */
+    private RichPost $post;
+
+    public function __construct(string $body, RichPost $post)
+    {
+        $this->body = $body;
+        $this->post = $post;
+    }
+
+    public function getId(): ?int
+    {
+        return $this->id;
+    }
+
+    public function getBody(): string
+    {
+        return $this->body;
+    }
+
+    public function getPost(): RichPost
+    {
+        return $this->post;
+    }
+}
--- a/tests/Fixtures/Entity/RichPost.php
+++ b/tests/Fixtures/Entity/RichPost.php
@@ -1,0 +1,68 @@
+<?php
+
+namespace Zenstruck\Foundry\Tests\Fixtures\Entity;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ */
+class RichPost
+{
+    /**
+     * @ORM\Id
+     * @ORM\GeneratedValue
+     * @ORM\Column(type="integer")
+     */
+    private ?int $id = null;
+
+    /**
+     * @ORM\Column(type="string", length=255)
+     */
+    private string $title;
+
+    /**
+     * @ORM\OneToMany(targetEntity=RichComment::class, mappedBy="post", orphanRemoval=true)
+     */
+    private Collection $comments;
+
+    public function __construct(string $title)
+    {
+        $this->title = $title;
+        $this->comments = new ArrayCollection();
+    }
+
+    public function getId(): ?int
+    {
+        return $this->id;
+    }
+
+    public function getTitle(): string
+    {
+        return $this->title;
+    }
+
+    /**
+     * @return Collection<RichComment>
+     */
+    public function getComments(): Collection
+    {
+        return $this->comments;
+    }
+
+    public function addComment(RichComment $comment): void
+    {
+        if (!$this->comments->contains($comment)) {
+            $this->comments->add($comment);
+        }
+    }
+
+    public function removeComment(RichComment $comment): void
+    {
+        if ($this->comments->contains($comment)) {
+            $this->comments->removeElement($comment);
+        }
+    }
+}
--- a/tests/Fixtures/Factories/RichPostFactory.php
+++ b/tests/Fixtures/Factories/RichPostFactory.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Zenstruck\Foundry\Tests\Fixtures\Factories;
+
+use Zenstruck\Foundry\ModelFactory;
+use Zenstruck\Foundry\Tests\Fixtures\Entity\RichPost;
+
+final class RichPostFactory extends ModelFactory
+{
+    protected static function getClass(): string
+    {
+        return RichPost::class;
+    }
+
+    protected function getDefaults(): array
+    {
+        return [
+            'title' => self::faker()->text(),
+        ];
+    }
+}
--- a/tests/Functional/ORMModelFactoryTest.php
+++ b/tests/Functional/ORMModelFactoryTest.php
@@ -12,8 +12,11 @@
 namespace Zenstruck\Foundry\Tests\Functional;
 
 use Doctrine\ORM\EntityManagerInterface;
+use Zenstruck\Foundry\Instantiator;
 use Zenstruck\Foundry\Tests\Fixtures\Entity\Category;
 use Zenstruck\Foundry\Tests\Fixtures\Factories\AddressFactory;
+use Zenstruck\Foundry\Tests\Fixtures\Factories\CascadeRichCommentFactory;
+use Zenstruck\Foundry\Tests\Fixtures\Factories\CascadeRichPostFactory;
 use Zenstruck\Foundry\Tests\Fixtures\Factories\CategoryFactory;
 use Zenstruck\Foundry\Tests\Fixtures\Factories\CommentFactory;
 use Zenstruck\Foundry\Tests\Fixtures\Factories\ContactFactory;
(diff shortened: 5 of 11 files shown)


tests/Fixtures/Entity/RichPost.php
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @ORM\Column(type="string", length=255)
+     */
+    private $title;
+
+    /**
+     * @ORM\OneToMany(targetEntity=RichComment::class, mappedBy="post")
+     */
+    private Collection $comments;
```
kbond, line 29: This typehint causes problems when running tests on < PHP 7.4...

tests/Fixtures/Factories/RichPostFactory.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Zenstruck\Foundry\Instantiator;
+use Zenstruck\Foundry\ModelFactory;
+use Zenstruck\Foundry\Tests\Fixtures\Entity\RichPost;
+
+final class RichPostFactory extends ModelFactory
+{
+    protected function initialize()
+    {
+        return $this->instantiateWith((new Instantiator())->alwaysForceProperties());
```
nikophil: nice, I didn't know this one. This way no need for any setter :tada:

tests/Fixtures/Entity/RichPost.php
nikophil: I think we can skip the related test for lower versions

tests/Functional/ORMModelFactoryTest.php
```diff
@@ -493,6 +495,24 @@ public function embeddables_are_never_persisted(): void
...
+        $post = RichPostFactory::createOne([
+            'comments' => RichCommentFactory::new()->many(5),
+        ]);
+
+        RichPostFactory::assert()->count(1);
+        RichCommentFactory::assert()->count(5);
+        $this->assertCount(5, $post->getComments());
+
+        foreach (RichCommentFactory::all() as $comment) {
+            $this->assertSame($post->getId(), $comment->getPost()->getId());
```
nikophil: I don't understand how this works :upside_down_face: 
we never say "the post we're creating is the one to use for all comments". How is it possible that a new post is not created for each comment?

src/Instantiator.php
```diff
@@ -130,6 +132,19 @@ public function alwaysForceProperties(array $properties = []): self
      */
     public static function forceSet(object $object, string $property, $value): void
     {
+        $refProp = self::accessibleProperty($object, $property);
+
+        // if value is an array and the property expects a doctrine collection, convert to doctrine collection
+        if (\is_array($value) && \method_exists($refProp, 'getType') && $type = $refProp->getType()) {
```
nikophil: indeed, directly returning array collection would really be a better fit, and seems simpler (deprecation layer excepted...)

tests/Fixtures/Entity/RichPost.php
kbond: It's not that easy as the `RichPost` will be loaded on every version.

tests/Fixtures/Factories/RichPostFactory.php
kbond: You can set this globally in your config too.

tests/Functional/ORMModelFactoryTest.php
kbond: There is some magic in `Factory` that handles this when creating relationships.

nikophil: nice! it was this particular magic I didn't know about when I wrote the issue

tests/Fixtures/Entity/RichPost.php
nikophil: I got another idea :joy: 
https://www.php.net/supported-versions.php

nikophil: I got another idea :imp: 
Capture d’écran du 2022-10-13 17-40-32

kbond: I know. I think we should bump the min PHP version to 8.0 in `1.x` and leave a `1.?.x` branch for bug fixes.

nikophil: I'm a lot in favor for this move :+1:

tests/Functional/ORMModelFactoryTest.php
Kocal: What about asserting between objects instead of scalars?
```suggestion
            $this->assertSame($post, $comment->getPost());
```

tests/Fixtures/Entity/RichComment.php
```diff
@@ -0,0 +1,50 @@
...
+     */
+    private $post;
+
+    public function __construct(string $body, RichPost $post)
+    {
+        $this->body = $body;
+        $this->post = $post;
+    }
+
+    public function getId(): ?int
```
benblub, line 36: getId() return type should be `int`

kbond: What about when the id truly is null (before the entity is persisted)?

benblub: good point.

nikophil: if this entity is part of a "rich model" it should use uuids and set the uuid in its constructor :innocent: 

but I think we don't really care for these fixtures :shrug:

kbond: This is very true! (I didn't want to add a dep)

X-Coder264:
@kbond Is there anything left to be done before this can be merged?

We are currently trying to migrate to Foundry but this is blocking us.

As CommentFactory in our case has `'post' => PostFactory::new(),` in its `getDefaults` method (so that the `CommentFactory::createOne();` call can work) the piece of code below creates two post records and one comment record in the database (instead of the expected one post record with one comment). Plus the created comment record gets created for the other post, not the one that actually gets returned by the post factory call.

```php
        $post = PostFactory::createOne([
            'comments' => CommentFactory::new()->many(1),
        ])->object();

       $post->getComments()->isEmpty(); // currently true, the expected result is false
```

nikophil:
Hi @X-Coder264 

> Is there anything left to be done before this can be merged?

we would like to try another way of doing this: `FactoryCollection::create()` could return directly an `ArrayCollection`. We could give a try in the next few days if this feature is missing for someone

kbond:
Hey @X-Coder264, I'd like finish this one up shortly, just getting back from holidays.

kbond:
I think this is ready as-is. @X-Coder264, can you try this out in your project?

There's some things I'd still like to do:
1. Split `Instantiator` into separate instantiator/hydrator (#388)
2. Use `symfony/property-info` to detect doctrine collections (right now the prop needs to be type-hinted as `Collection`).
3. Add `FactoryCollection::asDoctrineCollection()` or something like that.

But these can be added in the future (probably when we split the instantiator).

kbond:
@nikophil, any idea on the CI failure? Related to #393 somehow?

nikophil:
very strange :thinking:
I already had these errors in #393 but they were fixed by adding a condition on the migrations related to PHP 8.1
https://github.com/zenstruck/foundry/blob/1.x/tests/Fixtures/Migrations/Version20230109134337.php#L19

nikophil:
would you mind rebasing on 1.x?

kbond:
> would you mind rebasing on 1.x?

Done but still a problem. I see the failures from #393 that had the same issue, was [this](https://github.com/zenstruck/foundry/pull/393/files#diff-9f2b7d5a434ac908dce450d97972203bbf496cab5cd60547ffac77b374559a4f) the solution for that one?

X-Coder264:
@kbond I've tried it out and for some reason I'm still getting the same old/wrong behavior even though I can see that the new `$value = new ArrayCollection($value);` code path gets successfully triggered when I put a Xdebug breakpoint there. I can't see any meaningful difference between your test case and my code (except that I use cascade persist, orphan removal and fetch extra lazy on my mapping but none of those should make a difference) :thinking: 

EDIT: It does make a difference.

This works (same scenario as your test case):

```php
    /**
     * @ORM\OneToMany(targetEntity="Comment", mappedBy="post")
     */
    private Collection $comments;
```

```php
    /**
     * @ORM\ManyToOne(targetEntity="Post", inversedBy="comments")
     * @ORM\JoinColumn(nullable=false)
     */
    private Post $post;
```

This doesn't work as expected:

```php
    /**
     * @ORM\OneToMany(targetEntity="Comment", mappedBy="post", cascade={"persist"}, orphanRemoval=true, fetch="EXTRA_LAZY")
     */
    private Collection $comments;
```

```php
    /**
     * @ORM\ManyToOne(targetEntity="Post", inversedBy="comments", cascade={"persist"})
     * @ORM\JoinColumn(nullable=false)
     */
    private Post $post;
```

X-Coder264:
Ok, found the issue. Removing the cascade persist "fixes" the issue. So if I change my mapping to this it works as expected:

```php
    /**
     * @ORM\OneToMany(targetEntity="Comment", mappedBy="post", orphanRemoval=true, fetch="EXTRA_LAZY")
     */
    private Collection $comments;
```

I'm not sure why does cascade persist change anything, the behavior with or without cascade persist should be the same (getting back only one post with X comments).

kbond:
Ok, thanks for testing and finding this edge case. I'll create a test case that uses cascade: persist and see if I can figure out why it doesn't work as expected.

There are some nuances with cascade persist that I don't fully understand (as I never use this feature myself). These are accounted for elsewhere in foundry but I didn't write this code. Maybe something with one of these is causing the problem.

kbond:
I found the issue but not sure yet how to solve: 

When _NOT_ using cascade persist, the instantiation of the _comment_ collection is done after persisting the _post_. The _comments_ are instantiated with the previously created _post_ (this overrides the default post so the extra post per comment aren't created).

When using cascade persist, the collection needs to be instantiated before creating the post so I don't have a chance to set the post (as it isn't instantiated yet).

tests/Functional/ORMModelFactoryTest.php
```diff
@@ -540,6 +545,102 @@ public function can_count_based_on_a_relationship(): void
         self::assertSame(2, PostFactory::count(['category' => $category]));
     }
 
+    /**
+     * @test
+     */
+    public function can_create_factories_for_rich_domain_models(): void
```
kbond, line 551: This test passes and adds the comments via `RichPost::setComment()` (property-accessor).

tests/Functional/ORMModelFactoryTest.php
```diff
@@ -540,6 +545,102 @@ public function can_count_based_on_a_relationship(): void
...
+        $post->save();
+
+        RichPostFactory::assert()->count(1);
+        RichCommentFactory::assert()->count(4);
+    }
+
+    /**
+     * @test
+     */
+    public function can_create_factories_for_rich_domain_models_using_force(): void
```
kbond, line 575: This test passes and adds the comments by forcing `RichPost::$comments`.

tests/Functional/ORMModelFactoryTest.php
```diff
@@ -540,6 +545,102 @@ public function can_count_based_on_a_relationship(): void
...
+        $post->save();
+
+        RichPostFactory::assert()->count(1);
+        RichCommentFactory::assert()->count(4);
+    }
+
+    /**
+     * @test
+     */
+    public function can_create_factories_for_cascade_rich_domain_models(): void
```
kbond, line 599: Fails due to the cascade persist on `CascadePost::$comments`.

tests/Functional/ORMModelFactoryTest.php
```diff
@@ -540,6 +545,102 @@ public function can_count_based_on_a_relationship(): void
...
+        $post->save();
+
+        CascadeRichPostFactory::assert()->count(1);
+        CascadeRichCommentFactory::assert()->count(4);
+    }
+
+    /**
+     * @test
+     */
+    public function can_create_factories_for_cascade_rich_domain_models_using_force(): void
```
kbond, line 623: Fails due to the cascade persist on `CascadePost::$comments`.

kbond:
I now test using `rich entities` with and without `alwaysForceProperties()` and this all works as expected.

The same tests using _cascade persist_ do not pass. Still not sure how to solve. We're kind of in a chicken-egg situation: to create the comment, we need the post but to create the post, we need the comment...

kbond:
I think when we [split](https://github.com/zenstruck/foundry/issues/388) the instantiator into an instantiator and a hydrator we might be able to make `cascade: persist` work here.

> We're kind of in a chicken-egg situation: to create the comment, we need the post but to create the post, we need the comment...

We could _instantiate_ the post (w/o hydrating), then _instantiate_ and _hydrate_ the comments (using the post object), then hydrate the post with the created comments. Complicated but doable I think.

nikophil:
I think we really should have this work with `cascade={"persist"}` since it is really common to have this

kbond:
> I think we really should have this work with cascade={"persist"} since it is really common to have this

Agreed, I'm hoping #395 will help solve.

nikophil:
@kbond I've found why some tests not related to this PR are failing. Could you add this code to all migration which does not have it?
```php
    public function isTransactional(): bool
    {
        return false;
    }
```
Migrations are only tested without DAMA, then they are only tested for "php8.0/sf4.4/lowest" and "php8.2/sf6.2/highest" but I don't understand why it does not fail in the second scenario... maybe `doctrine/migration` is clever enough to know transaction mode should be disabled in certain cases?

I'm considering to add a custom PHPStan rule since it's been multiple times I've lost time on this.

nikophil:
hey @kbond 
could we close this PR now, since we'e merged https://github.com/zenstruck/foundry/pull/645 ?

PR #318, opened 2022-10-19
[feat] pass an index to FactoryCollection attributes

### LINKED ISSUE #316: [Feature Request] Allow to pass index in `createMany()` callback
Hi! 

I'm currently migrating. from Alice (and AliceBundle) to Foundry, but I lack from an interesting feature, the [fixtures list](https://github.com/nelmio/alice/blob/master/doc/complete-reference.md#fixture-lists).

With Alice, the following code will generate 6 users with username `front_user1`, `front_user2`, etc...: 
```yaml
App\Core\Entity\User\User:
    forum_user_{1..6}:
        username: 'front_user<current()>'
        email: 'front_user<current()>\@foobar.com'
```

However I don't find the equivalent with Foundry, I was hoping to use `createMany` but it does not work as I expected: 
```php
UserFactory::createMany(6, function (int $number) {
    return [
        'username' => 'front_user' . $number,
        'email' => 'front_user' . $number . '@foobar.com',
    ];
});
```

It seems the current workaround is: 
```php
for ($i = 1; $i <= 6; ++$i) {
    UserFactory::createOne([
        'username' => 'front_user' . $i,
        'email' => 'front_user' . $i . '@foobar.com',
    ]);
}
```

It's not really a big deal, and maybe it was on purpose when developing Foundry, but WDYT about it?

Thanks

nikophil:
fixes #316 

:warning: this PR introduces a *small* BC break: when a callback was passed to `withAttributes()` we were giving it an instance of faker, which is not made anymore. This behavior was not even documented nor tested, and the faker instance can be retrieved in another way

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,13 +71,13 @@ fixcs: docker-start bin/tools/cs-fixer/vendor ### Run PHP CS-Fixer
 	@${DOCKER_PHP} bin/tools/cs-fixer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer --no-interaction --diff -v fix
 
 bin/tools/cs-fixer/vendor: vendor bin/tools/cs-fixer/composer.json bin/tools/cs-fixer/composer.lock
-	@${DOCKER_PHP} composer bin cs-fixer update
+	@${DOCKER_PHP} composer bin cs-fixer install
 
 sca: docker-start bin/tools/psalm/vendor ### Run Psalm
 	@${DOCKER_PHP} bin/tools/psalm/vendor/vimeo/psalm/psalm --config=./psalm.xml
 
 bin/tools/psalm/vendor: vendor bin/tools/psalm/composer.json bin/tools/psalm/composer.lock
-	@${DOCKER_PHP} composer bin psalm update
+	@${DOCKER_PHP} composer bin psalm install
 
 database-generate-migration: docker-start vendor ### Generate new migration based on mapping in Zenstruck\Foundry\Tests\Fixtures\Entity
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -71,13 +71,13 @@ fixcs: docker-start bin/tools/cs-fixer/vendor ### Run PHP CS-Fixer
 	@${DOCKER_PHP} bin/tools/cs-fixer/vendor/friendsofphp/php-cs-fixer/php-cs-fixer --no-interaction --diff -v fix
 
 bin/tools/cs-fixer/vendor: vendor bin/tools/cs-fixer/composer.json bin/tools/cs-fixer/composer.lock
-	@${DOCKER_PHP} composer bin cs-fixer update
+	@${DOCKER_PHP} composer bin cs-fixer install
```
nikophil, line 74: I had to change this, otherwise psalm and csfixer's composer.lock will always be updated

docs/index.rst
```diff
@@ -278,6 +278,14 @@ Using your Factory
     PostFactory::createMany(5); // returns Post[]|Proxy[]
     PostFactory::createMany(5, ['title' => 'My Title']);
 
+    // Create 5 posts with incremental title
+    PostFactory::createMany(
+        5,
+        static function(int $i) {
+            return ['title' => "Title $i"];
```
Kocal:
```suggestion
            return ['title' => "Title $i"]; // Title 1, Title 2, ... Title 5
```

Or in a better way, the idea behind this suggestion is to indicate if `$i` will starts at `0` or `1`.

Thanks!

Kocal (commented):
Thank you!

kbond (approved):
This is great! I'm fine with that BC break - seems like a relic from the very early days of working on this library.

Kocal:
Again, thanks @nikophil ! :)

Issue #319, opened 2022-11-02
Remove "bundleless" usage


nikophil:
Based on #110 and https://github.com/zenstruck/foundry/issues/311#issuecomment-1275669557 we would like to remove "bundleless" usage.

Here are some thoughts:
- [x] all env vars such as `FOUNDRY_RESET_CONNECTIONS` or `FOUNDRY_RESET_OBJECT_MANAGERS` should be passed as config https://github.com/zenstruck/foundry/pull/320
- [x] add something like `FOUNDRY_RESET_ODM_OBJECT_MANAGERS` https://github.com/zenstruck/foundry/pull/320
- [ ] "bundleless" usage should not require doctrine (there is presently an error when requiring foundry without doctrine) - the factories would automatically be with `withoutPersisting()`  behavior
- [x] we could deprecate using TestState with KernelTestCase - triggered if the bundle isn't enabled. Then we could say "TestState is only applicable to non-Kernel (unit) tests" https://github.com/zenstruck/foundry/pull/322

kbond:
> we need to keep the ability to add global state in

We do but I wonder if this could be configured?

```yaml
zenstruck_foundry:
    global_state:
        - invokable.service # just a service with ::invoke()
        - App\Stories\StoryThatIsAService
        - App\Stories\GlobalStory # if not a service, just run GlobalStory::load()
        - ...
```

Then global state could be used in your dev environment when loading fixtures. Stories as services could now be used as global state too.

If this can be done, maybe we can deprecate `TestState::addGlobalState()` too.

PR #320, opened 2022-11-07
[feat] [remove bundleless usge] use config instead of environment variables

### LINKED ISSUE #311: [feature] add a way to exclude database connection for mongo support on reset database
There's way to exclude database connection for the ORM with 
```
FOUNDRY_RESET_CONNECTIONS=connection1,connection2
FOUNDRY_RESET_OBJECT_MANAGERS=manager1,manager2
```
But we cannot exclude ODM connection for the moment.

It would be nice to add
```
FOUNDRY_RESET_ODM_CONNECTIONS
FOUNDRY_RESET_ODM_OBJECT_MANAGERS
```

nikophil:
fixes https://github.com/zenstruck/foundry/issues/311

also related to https://github.com/zenstruck/foundry/issues/319

I don't really know if/how we should test this:
- unit test of `DatabaseResetter` (and ODM/ORM schema resetters) would be mocking hell
- functional test would require several connections/objet managers, which really feels over-engineered...

Even testing the configuration would be tricky. I usually use a `KernelBuilder` for this.

WDYT?


src/Bundle/DependencyInjection/ZenstruckFoundryExtension.php
```diff
@@ -88,4 +91,42 @@ private function configureDefaultInstantiator(array $config, ContainerBuilder $c
...
+        }
+
+        $configurationDefinition->setArgument('$ormConnectionsToReset', $config['orm']['connections'] ?? []);
+        $configurationDefinition->setArgument('$ormObjectManagersToReset', $config['orm']['object_managers'] ?? []);
+        $configurationDefinition->setArgument('$ormResetMode', $config['orm']['reset_mode'] ?? 'schema');
+        $configurationDefinition->setArgument('$odmObjectManagersToReset', $config['odm']['object_managers'] ?? []);
+    }
+
+    /**
+     * @psalm-suppress UndefinedDocblockClass
```
nikophil, line 118: I don't even understand why this error is triggererd... 
```
ERROR: UndefinedDocblockClass - src/Bundle/DependencyInjection/ZenstruckFoundryExtension.php:126:34 - Docblock-defined class, interface or enum named UnitEnum does not exist (see https://psalm.dev/200)
            \is_array($bundles = $container->getParameter('kernel.bundles')) ? $bundles : [],
```

src/Configuration.php
```diff
@@ -35,12 +35,32 @@ final class Configuration
...
+    private $ormConnectionsToReset;
+
+    /** @var list<string> */
+    private $ormObjectManagersToReset;
+
+    /** @var string */
+    private $ormResetMode;
+
+    /** @var list<string> */
+    private $odmObjectManagersToReset;
```
nikophil, line 51: I'm wondering if this is the right place for these options...
here are few other options:
- declare a `DatabaseResetterConfiguration` service
- transform `DatabaseResetter` into a service, and it will be injected directly all these stuff
- transform `DatabaseResetter` and `ORMDatabaseResetter` and `ODMSchemaResetter` into a service, and each one will be injected with its related options

WDYT?

src/Test/AbstractSchemaResetter.php
```diff
@@ -24,4 +24,15 @@ protected function runCommand(Application $application, string $command, array $
             throw new \RuntimeException(\sprintf('Error running "%s": %s', $command, $output->fetch()));
         }
     }
+
+    /**
+     * @param list<string> $availableObjectsToReset
+     * @param list<string> $candidateObjectsToReset
+     */
+    protected static function validateObjectsToReset(string $objectsType, array $availableObjectsToReset, array $candidateObjectsToReset): void
```
nikophil, line 32: this was made to make the code DRY, but I don't like the wording :shrug:

nikophil, line 32: this check is made at runtime because it would require foundry bundle to read doctrine's configuration through a compiler pass, which is a little bit weird. WDYT?

src/Configuration.php
kbond: Let's leave here, the class is internal so we can move later w/o consequence.

src/Test/AbstractSchemaResetter.php
kbond: Again, internal, we can refactor later.

kbond: I'm fine with runtime.

src/Test/DatabaseResetter.php
```diff
@@ -93,4 +89,30 @@ private static function createApplication(KernelInterface $kernel): Application
 
         return $application;
     }
+
+    private static function createORMDatabaseResetter(Application $application, KernelInterface $kernel): ORMDatabaseResetter
+    {
+        $container = $kernel->getContainer();
+        $configuration = $container->get(Configuration::class);
```
nikophil: @kbond this breaks the tests which are not using the bundle. 

Should we maintain "bundleless" usage until 2.0 or do you think this could be acceptable?

EDIT: I've added some kind of deprecation layer around that. I had to change `SYMFONY_DEPRECATIONS_HELPER` for tests which not uses the bundle

nikophil:
about the test stuf, maybe a quick win would be to name our connections / managers with another name than "default" use this name in the config

kbond:
If not explicitly setting the connections/managers, does it use `default` or run on all available connections/managers?

nikophil:
we're using default names: `$this->registry->getDefaultConnectionName()`, `$this->registry->getDefaultManagerName()`

kbond:
Ah, ok, hah - if we add a non-default connection/manager to the config for the tests, we lose the test for choosing the defaults... Let's leave untested for now - I don't want yet another test permutation. Of course, if there's a reported issue, we can re-evaluate.

nikophil:
ok, hopefully we'll remove 5 or more test permutation after `USE_FOUNDRY_BUNDLE` usage will be removed

kbond:
_I honestly think I may the only person using foundry this way..._

src/Test/ResetDatabase.php
```diff
@@ -34,7 +33,9 @@ public static function _resetDatabase(): void
         $kernel = static::createKernel();
         $kernel->boot();
 
-        DatabaseResetter::resetDatabase($kernel);
+        if (self::shouldReset($kernel)) {
+            DatabaseResetter::resetDatabase($kernel);
+        }
```
nikophil, line 38: I've literally spent hours on this and I figured out that the kernel should not be booted before calling `StaticDriver::setKeepStaticConnections(false);`. 

Since the `shouldReset()` must access to the config, we need the kernel to be booted before calling it.

At first I was trying to do something like this.
```php
        $kernel = static::createKernel();
        $kernel->boot();

        if (!self::shouldReset($kernel)) {
            // $kernel->shutdown()
            return;
        }

        if ($isDAMADoctrineTestBundleEnabled = DatabaseResetter::isDAMADoctrineTestBundleEnabled()) {
            // disable static connections for this operation
            StaticDriver::setKeepStaticConnections(false);
        }

        DatabaseResetter::resetDatabase($kernel);
```
and all my tests resulted in errors such as:
> PDOException: SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT DOCTRINE2_SAVEPOINT_4 does not exist

src/Test/ResetDatabase.php
```diff
@@ -18,23 +21,22 @@ trait ResetDatabase
...
-            return;
-        }
-
         if (!\is_subclass_of(static::class, KernelTestCase::class)) {
             throw new \RuntimeException(\sprintf('The "%s" trait can only be used on TestCases that extend "%s".', __TRAIT__, KernelTestCase::class));
         }
 
         if ($isDAMADoctrineTestBundleEnabled = DatabaseResetter::isDAMADoctrineTestBundleEnabled()) {
             // disable static connections for this operation
+            // :warning: the kernel should not be booted before calling this!
```
nikophil, line 30: I've literally spent hours on this and I figured out that the kernel should not be booted before calling `StaticDriver::setKeepStaticConnections(false);`. 

Since the `shouldReset()` must access to the config, we need the kernel to be booted before calling it.

At first I was trying to do something like this.
```php
        $kernel = static::createKernel();
        $kernel->boot();

        if (!self::shouldReset($kernel)) {
            // $kernel->shutdown(); // this does not work
            return;
        }

        if ($isDAMADoctrineTestBundleEnabled = DatabaseResetter::isDAMADoctrineTestBundleEnabled()) {
            // disable static connections for this operation
            StaticDriver::setKeepStaticConnections(false);

           // $kernel->boot(); // this does not work
        }

        DatabaseResetter::resetDatabase($kernel);
```
and all my tests resulted in errors such as:
> PDOException: SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT DOCTRINE2_SAVEPOINT_4 does not exist

src/Test/ResetDatabase.php
```diff
@@ -18,23 +21,22 @@ trait ResetDatabase
...
             StaticDriver::setKeepStaticConnections(false);
         }
 
         $kernel = static::createKernel();
         $kernel->boot();
 
-        DatabaseResetter::resetDatabase($kernel);
+        if (self::shouldReset($kernel)) {
+            DatabaseResetter::resetDatabase($kernel);
+        }
```
nikophil, line 39: I've literally spent hours on this and I figured out that the kernel should not be booted before calling `StaticDriver::setKeepStaticConnections(false);`. 

Since the `shouldReset()` must access to the config, we need the kernel to be booted before calling it.

At first I was trying to do something like this.
```php
        $kernel = static::createKernel();
        $kernel->boot();

        if (!self::shouldReset($kernel)) {
            // $kernel->shutdown(); // this does not work
            return;
        }

        if ($isDAMADoctrineTestBundleEnabled = DatabaseResetter::isDAMADoctrineTestBundleEnabled()) {
            // disable static connections for this operation
            StaticDriver::setKeepStaticConnections(false);

           // $kernel->boot(); // this does not work
        }

        DatabaseResetter::resetDatabase($kernel);
```
and all my tests resulted in errors such as:
> PDOException: SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT DOCTRINE2_SAVEPOINT_4 does not exist

OR
> PDOException: SQLSTATE[HY000] [1049] Unknown database 'zenstruck_foundry'

(the latter only occurs for [php 7.2/lowest dependencies](https://github.com/zenstruck/foundry/actions/runs/3417471515/jobs/5688647666))

I'm wondering if we should not look in this direction for #49 and #73

tests/Fixtures/Kernel.php
```diff
@@ -43,7 +43,7 @@ public function registerBundles(): iterable
             yield new DAMADoctrineTestBundle();
         }
 
-        if ('migrate' === \getenv('FOUNDRY_RESET_MODE')) {
+        if (\getenv('USE_MIGRATIONS')) {
```
nikophil, line 46: I changed the name of the env var in order to remove deprecations triggers in the tests

src/Test/ResetDatabase.php
kbond: Yeah sheesh, I saw all the builds and had never seen that specific error before.

> I'm wondering if we should not look in this direction for #49 and #73

Interesting!

kbond (approved):

PR #323, opened 2022-11-09
[feat] Allow to use foundry without Doctrine

### LINKED ISSUE #319: Remove "bundleless" usage

nikophil:
fixes https://github.com/zenstruck/foundry/issues/319

### DIFF
--- a/src/Test/TestState.php
+++ b/src/Test/TestState.php
@@ -9,6 +9,7 @@ use Psr\Container\NotFoundExceptionInterface;
 use Zenstruck\Foundry\ChainManagerRegistry;
 use Zenstruck\Foundry\Configuration;
 use Zenstruck\Foundry\Factory;
+use Zenstruck\Foundry\Instantiator;
 use Zenstruck\Foundry\StoryManager;
 
 /**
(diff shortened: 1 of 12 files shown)


tests/Functional/WithoutDoctrineKernelTest.php
```diff
@@ -0,0 +1,38 @@
...
+use Zenstruck\Assert;
+use Zenstruck\Foundry\Test\Factories;
+use Zenstruck\Foundry\Test\ResetDatabase;
+use Zenstruck\Foundry\Tests\Fixtures\Entity\Address;
+use Zenstruck\Foundry\Tests\Fixtures\Factories\AddressFactory;
+use Zenstruck\Foundry\Tests\Fixtures\Kernel;
+
+final class WithoutDoctrineKernelTest extends KernelTestCase
+{
+    use Factories, ResetDatabase;
```
nikophil, line 18: ResetDatabase on purpose, just to check if no error is thrown

kbond (changes requested):
Code looks good, we'll need some doc changes though.

What about:
1. removing all reference to bundle-less usage.
2. add a unit testing section to document/explain `TestState::configure()`

tests/Functional/WithDoctrineDisbabledKernelTest.php
```diff
@@ -0,0 +1,48 @@
...
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\HttpKernel\KernelInterface;
+use Zenstruck\Assert;
+use Zenstruck\Foundry\Test\Factories;
+use Zenstruck\Foundry\Tests\Fixtures\Entity\Address;
+use Zenstruck\Foundry\Tests\Fixtures\Factories\AddressFactory;
+use Zenstruck\Foundry\Tests\Fixtures\Kernel;
+
+final class WithDoctrineDisbabledKernelTest extends KernelTestCase
```
kbond:
```suggestion
final class WithDoctrineDisabledKernelTest extends KernelTestCase
```

src/Test/TestState.php
```diff
@@ -91,9 +112,17 @@ public static function bootFoundry(?Configuration $configuration = null): void
             $configuration->disableDefaultProxyAutoRefresh();
         }
 
+        self::bootFoundry($configuration);
+    }
+
+    public static function bootFoundry(Configuration $configuration): void
```
nikophil, line 127: I was thinking to drop this method because it just call another static method, but maybe it could be useful to keep it to centralize "boot" step, if maybe we got something else than Factory to boot

kbond: Thinking we should mark everything in here except `configure()` as `@internal`/`@deprecated`, wdyt?

nikophil: absolutely!

kbond (approved):

PR #326, opened 2022-11-10
[feature] require php8+

kbond:


nikophil:
Hi @kbond 

I've found why the CI is red: https://github.com/doctrine/migrations/blob/3.5.x/docs/en/explanation/implicit-commits.rst

I've actually learnt something here.
The error is gone if I add this code to the migrations.
```php
    public function isTransactional(): bool
    {
        return false;
    }
```

Maybe we can enforce it with the help of a simple phpstan rule...

PR #327, opened 2022-11-11
[refactor] modernize code with rector

nikophil:
based on #326 

I played around with Rector today.

Mainly modernized the code to php8:
- added arrow function
- typed everything that should be (except for some edge case with BC problems)
- used property constructor promotion when possible

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [7.2, 7.4, 8.0, 8.1]
+        php: [8.0, 8.1]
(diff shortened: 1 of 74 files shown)


kbond (commented):
Nice! Should we switch to attributes for the test entities?

nikophil:
I made the mistake and... had to rollback since sf4.4 + lowest doctrine does not support it :disappointed:

kbond:
Would bumping the min doctrine req help?

What do you think about requiring 5.4+?

nikophil:
> What do you think about requiring 5.4+?

not a huge fan, since it's still maintained

kbond:
Ok, forgot it had another year

PR #328, opened 2022-11-11
[chore] use PhpStan

nikophil:
based on #326


phpstan-baseline.neon
```diff
@@ -0,0 +1,61 @@
+parameters:
+	ignoreErrors:
```
nikophil, line 2: these error are all kind of "normal" or hard to fix

phpstan.neon
```diff
@@ -0,0 +1,36 @@
+includes:
+	- phpstan-baseline.neon
+
+parameters:
+    inferPrivatePropertyTypeFromConstructor: true
+    checkMissingIterableValueType: false
+    checkUninitializedProperties: true
+    checkGenericClassInNonGenericObjectType: false
```
nikophil, line 8: this one is a pain to fix, and because we may change how the proxy stuff is done, the pain is not worth it

src/AnonymousFactory.php
```diff
@@ -23,12 +25,12 @@ public static function new(string $class, $defaultAttributes = []): self
...
      *
      * @return Proxy&TModel
-     * @psalm-return Proxy<TModel>
+     * @phpstan-return Proxy<TModel>
      */
     public function findOrCreate(array $attributes): Proxy
     {
         if ($found = $this->repository()->find($attributes)) {
-            return \is_array($found) ? $found[0] : $found;
+            return $found;
```
nikophil, line 33: don't understand the `is_array()`... find() would only return an object...

kbond (approved):

.github/workflows/ci.yml
```diff
@@ -204,11 +204,11 @@ jobs:
         with:
           composer-options: --prefer-dist
 
-      - name: Install Psalm
-        run: composer bin psalm install
+      - name: Install PhpStan
```
OskarStark, line 207: PhpStan -> PHPStan

PR #333, opened 2022-11-12
[docs] sort phpstan-method annotations

OskarStark:


nikophil:
Hi @OskarStark 

thank you for this.

We've recently moved from Psalm to PHPStan. So I think these annotation should be changed from `@phpstan-method` to `@psalm-method` and the text above should be adapted as well.

But I'm questioning the need of this now: all annotations with generics are duplicated: 
- 1 section for static analysis with generics
- 1 simpler section for auto-completion in the IDE without the generics

one out many example here: https://github.com/zenstruck/foundry/blob/1.x/src/AnonymousFactory.php#L27-L28

This was useful at the time when PHPStorm did not understand generics.

But now PHPStorm understands very well generics and I think we could remove all "IDE specific" annotation without generics in favor of 100% "regular annotation" (no more `@phpstan-*`) with generics. I'm just not aware of how other IDE understand generics and if the auto-completion would be degraded with this change :thinking: 

WDYT @kbond @OskarStark ?

EDIT: the `@method` annotations created by `maker:factory` should be updated as well

OskarStark:
> We've recently moved from Psalm to PHPStan. So I think these annotation should be changed from @phpstan-method to @psalm-method and the text above should be adapted as well.

I think it depends on the user, we in ur projectt are using PHPStan 🤷 Maybe both can be be added to the docs.

However, if I use the following in my Factory:
```php
 * @extends ModelFactory<User>
 *
 * @method static User[]|Proxy[]                 all()
 * @method        Proxy|User                     create(array|callable $attributes = [])
 * @method static User[]|Proxy[]                 createMany(int $number, array|callable $attributes = [])
 * @method static User|Proxy                     createOne(array $attributes = [])
 * @method static User[]|Proxy[]                 createSequence(array|callable $sequence)
 * @method static User|Proxy                     find(array|mixed|object $criteria)
 * @method static User[]|Proxy[]                 findBy(array $attributes)
 * @method static User|Proxy                     findOrCreate(array $attributes)
 * @method static User|Proxy                     first(string $sortedField = 'id')
 * @method static User|Proxy                     last(string $sortedField = 'id')
 * @method static User|Proxy                     random(array $attributes = [])
 * @method static User|Proxy                     randomOrCreate(array $attributes = [])
 * @method static User[]|Proxy[]                 randomRange(int $min, int $max, array $attributes = [])
 * @method static User[]|Proxy[]                 randomSet(int $number, array $attributes = [])
 * @method static UserRepository|RepositoryProxy repository()
 */
```

I have PHPStan errors in my code:
CleanShot 2022-11-13 at 15 41 22@2x

If I add the following:
```php
 * @phpstan-method static User&Proxy createOne(array $attributes = [])
 * @phpstan-method static User[]&Proxy[] createMany(int $number, array|callable $attributes = [])
 * @phpstan-method static User[]&Proxy[] createSequence(array|callable $sequence)
 * @phpstan-method static User&Proxy find(object|array|mixed $criteria)
 * @phpstan-method static User&Proxy findOrCreate(array $attributes)
 * @phpstan-method static User&Proxy first(string $sortedField = 'id')
 * @phpstan-method static User&Proxy last(string $sortedField = 'id')
 * @phpstan-method static User&Proxy random(array $attributes = [])
 * @phpstan-method static User&Proxy randomOrCreate(array $attributes = [])
 * @phpstan-method static User[]&Proxy[] all()
 * @phpstan-method static User[]&Proxy[] findBy(array $attributes)
 * @phpstan-method static User[]&Proxy[] randomSet(int $number, array $attributes = [])
 * @phpstan-method static User[]&Proxy[] randomRange(int $min, int $max, array $attributes = [])
 * @phpstan-method User&Proxy create(array|callable $attributes = [])
 */
```

the errors are gone, but PhpStorm is complaining:
CleanShot 2022-11-13 at 15 42 12@2x

I tried to use `&` instead of `|` in `@method` annotation, and it worked, no PHPStan errors anymore, but still PhpStorm is complaining 🤔

kbond:
I haven't checked myself in ~6 months but what @OskarStark found matched my findings. I want to prioritize autocompletion so it looks like for now, both are still required.

kbond:
@OskarStark is this how php-cs-fixer sorts them?

OskarStark:
> @OskarStark is this how php-cs-fixer sorts them?

Yes

OskarStark:
Not sure why it is not working to use `@method` and `&`.

[It should](https://phpstan.org/blog/union-types-vs-intersection-types) 🤔

nikophil:
I just tried to use generics in the phpdoc of a factory
```diff
- @method static User|Proxy createOne(array $attributes = [])
+ @method static Proxy<User> createOne(array $attributes = [])
```
and indeed PHPStorm does not auto-complete anymore.. This is because `Proxy` is a mixin of a template, and [PHPStorm still does not understand it](https://youtrack.jetbrains.com/issue/WI-64022/Support-mixin-with-template)

---

So indeed there is still a need to add these phpdoc specific to phpstan. But IMO the phpdoc we are suggesting are not accurate, we should suggest phpdoc with generics, because it is what the factories actually return:
```php
 * @phpstan-method static Proxy<Post> createOne(array $attributes = [])
 * @phpstan-method static list<Proxy<Post>> createMany(int $number, array|callable $attributes = [])
 * @phpstan-method static list<Proxy<Post>> createSequence(array|callable $sequence)
 * @phpstan-method static Proxy<Post> find(object|array|mixed $criteria)
 * @phpstan-method static Proxy<Post> findOrCreate(array $attributes)
 * @phpstan-method static Proxy<Post> first(string $sortedField = 'id')
 * @phpstan-method static Proxy<Post> last(string $sortedField = 'id')
 * @phpstan-method static Proxy<Post> random(array $attributes = [])
 * @phpstan-method static Proxy<Post> randomOrCreate(array $attributes = [])
 * @phpstan-method static list<Proxy<Post>> all()
 * @phpstan-method static list<Proxy<Post>> findBy(array $attributes)
 * @phpstan-method static list<Proxy<Post>> randomSet(int $number, array $attributes = [])
 * @phpstan-method static list<Proxy<Post>> randomRange(int $min, int $max, array $attributes = [])
 * @phpstan-method Proxy<Post> create(array|callable $attributes = [])
```

OskarStark:
Thanks I will try that 👍🏻

Edit:
But then we should be able to go with @method and not @phpstan-method, right?

nikophil:
> But then we should be able to go with @method and not @phpstan-method, right?

I don't think so, as I tried to explain in my previous message, we're stuck here: `Proxy` is a [mixin of its own template type](https://github.com/zenstruck/foundry/blob/1.x/src/Proxy.php#L13-L14) and [PHPStorm does not understand this yet](https://youtrack.jetbrains.com/issue/WI-64022/Support-mixin-with-template). And then I don't see any workaround except adding these `@phpstan-method` annotations.

nikophil:
BTW, I think we can do better here and auto-detect that phpstan is installed and then auto-add these annotations to the factory

I'm merging your PR here, and will use the new order (+ generics) in the `make:factory` command.

Thanks for this, sorry for the annoyance :sweat_smile: 

(see https://github.com/zenstruck/foundry/issues/336)

OskarStark:
> BTW, I think we can do better here and auto-detect that phpstan is installed and then auto-add these annotations to the factory

Yes 👍 

> Thanks for this, sorry for the annoyance 😅

No Problem

Issue #336, opened 2022-11-14
`make:factory` enhancements


nikophil:
- [x] auto-detect phpstan and add specific `@phpstan-method` (not needed for psalm) https://github.com/zenstruck/foundry/pull/338
- [x] availability to create factory for object not related to doctrine https://github.com/zenstruck/foundry/pull/343
- [x] Use field limit length when specified #286 #294
- [x] support default property for mongoDB #238 https://github.com/zenstruck/foundry/pull/340
- [x] auto defaults for non-nullable 1-1/n-1 relationships #175 https://github.com/zenstruck/foundry/pull/221
- [x] Foundry generates the factory with $faker->text() if its a string, i think $faker->sentence() oder $faker->word() would fit better. If doctrine type Text, I would recommend to use $faker->text()
- [x] sort the default values alphanumeric
- [x] indent the phpdoc properly (see below)
- [x] Move the generated comments to @see annotations above the methods. (see below)
- [x] Import non-nullable relationship factories (https://github.com/zenstruck/foundry/pull/351#discussion_r1029584697)
- [x] auto-complete FQCN's `bin/console make:entity User<tab>` (auto-complete classes that start with `User`) https://github.com/zenstruck/foundry/pull/383
- [x] handle enums default https://github.com/zenstruck/foundry/pull/393
- [x] list embeddables in `make:factory`
~- [ ] add docs on `--no-persistence` behavior~ _won'tdo_
- [x] handle when factory with same name already exists https://github.com/zenstruck/foundry/pull/402
- [x] match directory/namespace structure  https://github.com/zenstruck/foundry/issues/401 https://github.com/zenstruck/foundry/pull/411
- [x] add a bundle's configuration for base namespace where to create factories

Some more complex features that would be very nice to add:
- [x] based on `auto defaults for non-nullable`: if the factory is missing, we could also create the missing factory
- [x] handle previous feature with `--all`
~- [ ] "update" factories~ _this one is really tricky, and not sure people will actually use it...._
- [ ] add an option to `make:entity`: `--with-factory`

kbond:
Some additional suggestions by @OskarStark:
1. > Foundry generates the factory with $faker->text() if its a string, i think $faker->sentence() oder $faker->word() would fit better. If doctrine type Text, I would recommend to use $faker->text()
2. > Another proposal would be to sort the default values alphanumeric
3. > And we should indent the phpdoc properly

[...]

PR #338, opened 2022-11-14
[feat] auto add phpstan annotations in make:factory

nikophil:
part of https://github.com/zenstruck/foundry/issues/336

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -207,20 +207,21 @@ This command will generate a ``PostFactory`` class that looks like this:
         /**
          * ...
          *
-         * @phpstan-method Post&Proxy create(array|callable $attributes = [])
-         * @phpstan-method static Post&Proxy createOne(array $attributes = [])
-         * @phpstan-method static Post&Proxy find(object|array|mixed $criteria)
-         * @phpstan-method static Post&Proxy findOrCreate(array $attributes)
-         * @phpstan-method static Post&Proxy first(string $sortedField = 'id')
-         * @phpstan-method static Post&Proxy last(string $sortedField = 'id')
-         * @phpstan-method static Post&Proxy random(array $attributes = [])
-         * @phpstan-method static Post&Proxy randomOrCreate(array $attributes = [])
-         * @phpstan-method static Post[]&Proxy[] all()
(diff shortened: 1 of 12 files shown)


docs/index.rst
```diff
@@ -207,20 +207,21 @@ This command will generate a ``PostFactory`` class that looks like this:
...
-         * @phpstan-method static Post&Proxy last(string $sortedField = 'id')
-         * @phpstan-method static Post&Proxy random(array $attributes = [])
-         * @phpstan-method static Post&Proxy randomOrCreate(array $attributes = [])
-         * @phpstan-method static Post[]&Proxy[] all()
-         * @phpstan-method static Post[]&Proxy[] createMany(int $number, array|callable $attributes = [])
-         * @phpstan-method static Post[]&Proxy[] createSequence(array|callable $sequence)
-         * @phpstan-method static Post[]&Proxy[] findBy(array $attributes)
-         * @phpstan-method static Post[]&Proxy[] randomRange(int $min, int $max, array $attributes = [])
-         * @phpstan-method static Post[]&Proxy[] randomSet(int $number, array $attributes = [])
+         * @phpstan-method Proxy<Post> create(array|callable $attributes = [])
```
nikophil, line 210: I'm leaving this here, because if phpstan is used in a custom way (eg: like in this bundle :wink: ) we cannot detect it is used in the project

src/Bundle/Maker/MakeFactory.php
```diff
@@ -233,4 +234,9 @@ private function defaultPropertiesFor(string $class, bool $allFields): iterable
             yield $property['fieldName'] => $value;
         }
     }
+
+    private function phpstanEnabled(): bool
+    {
+        return \file_exists("{$this->projectDir}/vendor/phpstan/phpstan/phpstan");
```
nikophil, line 240: not sure if it is the best way...

docs/index.rst
kbond: 👍🏻

src/Bundle/Maker/MakeFactory.php
kbond: Good enough for now.

kbond (approved):

PR #343, opened 2022-11-15
feat(make:factory): create factory for not-persisted objects

nikophil:
related to #336

### DIFF
--- a/src/Bundle/Maker/MakeFactory.php
+++ b/src/Bundle/Maker/MakeFactory.php
@@ -17,6 +17,7 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\HttpKernel\KernelInterface;
 use Zenstruck\Foundry\ModelFactory;
 
 /**
@@ -24,7 +25,7 @@ use Zenstruck\Foundry\ModelFactory;
  */
 final class MakeFactory extends AbstractMaker
 {
-    private const DEFAULTS = [
+    private const DEFAULTS_FOR_PERSISTED = [
         'ARRAY' => '[],',
         'ASCII_STRING' => 'self::faker()->text({length}),',
         'BIGINT' => 'self::faker()->randomNumber(),',
@@ -50,10 +51,20 @@ final class MakeFactory extends AbstractMaker
         'TIME_IMMUTABLE' => '\DateTimeImmutable::createFromMutable(self::faker()->datetime()),',
     ];
 
-    /** @var string[] */
-    private array $entitiesWithFactories = [];
+    private const DEFAULTS_FOR_NOT_PERSISTED = [
+        'array' => '[],',
+        'string' => 'self::faker()->text(),',
+        'int' => 'self::faker()->randomNumber(),',
+        'float' => 'self::faker()->randomFloat(),',
+        'bool' => 'self::faker()->boolean(),',
+        \DateTime::class => 'self::faker()->dateTime(),',
+        \DateTimeImmutable::class => '\DateTimeImmutable::createFromMutable(self::faker()->dateTime()),',
+    ];
 
-    public function __construct(private ManagerRegistry $managerRegistry, \Traversable $factories, private string $projectDir)
+    /** @var string[] */
+    private array $entitiesWithFactories;
+
+    public function __construct(private ManagerRegistry $managerRegistry, \Traversable $factories, private string $projectDir, private KernelInterface $kernel)
     {
         $this->entitiesWithFactories = \array_map(
             static fn(ModelFactory $factory): string => $factory::getEntityClass(),
@@ -68,7 +79,7 @@ final class MakeFactory extends AbstractMaker
 
     public static function getCommandDescription(): string
     {
-        return 'Creates a Foundry model factory for a Doctrine entity class';
+        return 'Creates a Foundry model factory for a Doctrine entity class or a regular object';
     }
 
     public function configureDependencies(DependencyBuilder $dependencies): void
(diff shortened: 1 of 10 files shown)


src/Bundle/Maker/MakeFactory.php
```diff
@@ -80,18 +90,19 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
     {
         $command
             ->setDescription(self::getCommandDescription())
-            ->addArgument('entity', InputArgument::OPTIONAL, 'Entity class to create a factory for')
+            ->addArgument('class', InputArgument::OPTIONAL, 'Entity, Document or class to create a factory for')
             ->addOption('namespace', null, InputOption::VALUE_REQUIRED, 'Customize the namespace for generated factories', 'Factory')
             ->addOption('test', null, InputOption::VALUE_NONE, 'Create in <fg=yellow>tests/</> instead of <fg=yellow>src/</>')
             ->addOption('all-fields', null, InputOption::VALUE_NONE, 'Create defaults for all entity fields, not only required fields')
+            ->addOption('not-persisted', null, InputOption::VALUE_NONE, 'Create a factory for an object not managed by Doctrine')
```
nikophil, line 98: not sure of the option's name :shrug:

src/Bundle/Maker/MakeFactory.php
```diff
@@ -80,18 +90,19 @@ public function configureCommand(Command $command, InputConfiguration $inputConf
     {
         $command
             ->setDescription(self::getCommandDescription())
-            ->addArgument('entity', InputArgument::OPTIONAL, 'Entity class to create a factory for')
+            ->addArgument('class', InputArgument::OPTIONAL, 'Entity, Document or class to create a factory for')
```
nikophil, line 94: this seemed more accurate to name this argument `class`

I don't think this introduce BC problem

src/Bundle/Maker/MakeFactory.php
```diff
@@ -234,14 +264,43 @@ private function defaultPropertiesFor(string $class, bool $allFields): iterable
...
             }
 
             yield $property['fieldName'] => \str_replace('{length}', (string) $length, $value);
         }
     }
 
+    /**
+     * @param class-string $class
+     */
+    private function defaultPropertiesForNotPersistedObject(string $class, bool $allFields): iterable
```
nikophil, line 286: I think in near future this class will be more and more complex, so maybe we should externalise `defaultPropertiesForPersistedObject()` and `defaultPropertiesForNotPersistedObject()` in their own class (maybe 2 class and an interface, I don't know yet).

src/Bundle/Maker/MakeFactory.php
```diff
@@ -234,14 +264,43 @@ private function defaultPropertiesFor(string $class, bool $allFields): iterable
...
+    {
+        $object = new \ReflectionClass($class);
+
+        foreach ($object->getProperties() as $property) {
+            // ignore identifiers and nullable fields
+            if (!$allFields && ($property->hasDefaultValue() || !$property->hasType() || $property->getType()?->allowsNull())) {
```
nikophil, line 292: this could be a first draft as I'm not sure how far we should go here:
- should we look at the constructor to know which properties are mandatory?
- should we activate `alwaysForceProperties` if a property has type, no default value and is not present in construjctor?
- should we look at phpdoc?

src/Bundle/Resources/skeleton/Factory.tpl.php
```diff
@@ -2,49 +2,54 @@
...
- * @method static <?= $entity->getShortName() ?>|Proxy randomOrCreate(array $attributes = [])
- * @method static <?= $entity->getShortName() ?>[]|Proxy[] all()
- * @method static <?= $entity->getShortName() ?>[]|Proxy[] createMany(int $number, array|callable $attributes = [])
- * @method static <?= $entity->getShortName() ?>[]|Proxy[] createSequence(array|callable $sequence)
- * @method static <?= $entity->getShortName() ?>[]|Proxy[] findBy(array $attributes)
- * @method static <?= $entity->getShortName() ?>[]|Proxy[] randomRange(int $min, int $max, array $attributes = [])
- * @method static <?= $entity->getShortName() ?>[]|Proxy[] randomSet(int $number, array $attributes = [])
+ * @method <?= $object->getShortName() ?>|Proxy create(array|callable $attributes = [])
+ * @method static <?= $object->getShortName() ?>|Proxy createOne(array $attributes = [])
+<?php if ($persisted): ?> * @method static <?= $object->getShortName() ?>|Proxy find(object|array|mixed $criteria)
```
nikophil, line 17: I was wondering if we should not split into two template files, but I think we should not: otherwise we would need to add new factories' methods in both places.

src/Bundle/Maker/MakeFactory.php
nikophil: I'm wondering if we should not auto activated this option if doctrine is not present, WDYT?

kbond: > not sure of the option's name

Me either, lets go with this for now.

> I'm wondering if we should not auto activated this option if doctrine is not present, WDYT?

Yep, seems reasonable.

src/Bundle/Maker/MakeFactory.php
kbond: Let's leave it simple to start. We could look into using `symfony/property-info`.

src/Bundle/Resources/skeleton/Factory.tpl.php
kbond: Yeah, if it gets out of hand later we can split.

kbond (approved):
LGTM!

src/Bundle/Maker/MakeFactory.php
nikophil: not sure about `symfony/property-info`, it does not help to guess if the property has a default value

nikophil: Maybe I'd merge this PR like this if you're OK and I'll make something a little bit more evolved in another PR

kbond: Yep!

PR #345, opened 2022-11-16
minor(make:factory): misc enhancements of maker

nikophil:
- minor(make:factory): correctly align phpdoc
- minor(make:factory): order imports
- minor(make:factory): order default values
- minor(make:factory): use faker->sentence()


nikophil:
@OskarStark I tried to align phpdoc the same way shown [here](https://github.com/zenstruck/foundry/issues/336#issuecomment-1315135593)

but PHP-CS-Fixer aligns it in another way. WDYT is the best?

Capture d’écran du 2022-11-16 08-27-24
(left is the same as in your screenshot, right is after csfixer pass)

tests/Fixtures/Maker/expected/can-create-factory-for-not-persisted-class-interactively.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return [
+            // TODO add your default values here (https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#model-factories)
+            'dateTimeImmutableMandatory' => \DateTimeImmutable::createFromMutable(self::faker()->dateTime()),
+            'dateTimeMandatory' => self::faker()->dateTime(),
+            'floatMandatory' => self::faker()->randomFloat(),
+            'intMandatory' => self::faker()->randomNumber(),
+            'someMandatoryPropertyWithUnionType' => null, // TODO add value manually
+            'someOtherObjectMandatory' => null, // TODO add Zenstruck\Foundry\Tests\Fixtures\Object\SomeOtherObject value manually
+            'stringMandatory' => self::faker()->text(),
```
OskarStark:
```suggestion
            'stringMandatory' => self::faker()->text(),
```

`text()` could be too long, what about using `->sentence()` or just `word()` ?

nikophil: yes, I wanted to handle this point in a further commit

OskarStark:
> @OskarStark I tried to align phpdoc the same way shown but PHP-CS-Fixer aligns it in another way. WDYT is the best?

Lets use your version now, it anyway depends on the config of the user

kbond (commented):
Where's the change to use faker()->sentence()?

nikophil:
sorry, it's here now.

I think we cannot use `sentence()` for doctrine types, because we want to control the length, and that's not possible with `sentence()`

kbond (approved):
Agreed about keeping entity default as text().

PR #347, opened 2022-11-16
chore: Use PHP 8.1

OskarStark:


nikophil:
foundry still supports PHP 8.0, then we'd like to run the CI on PHP 8.0 - this will also be an incentive for developers to not use php 8.1 features

OskarStark:
Ok

PR #351, opened 2022-11-17
feat(make:factory): use factories to default non-nullable relationships

nikophil:
all work made by @benblub  :wink:

### DIFF
--- a/src/Bundle/Maker/MakeFactory.php
+++ b/src/Bundle/Maker/MakeFactory.php
@@ -61,14 +61,21 @@ final class MakeFactory extends AbstractMaker
         \DateTimeImmutable::class => '\DateTimeImmutable::createFromMutable(self::faker()->dateTime()),',
     ];
 
-    /** @var string[] */
+    /** @var array<string, string> */
     private array $entitiesWithFactories;
 
     public function __construct(private ManagerRegistry $managerRegistry, \Traversable $factories, private string $projectDir, private KernelInterface $kernel)
     {
-        $this->entitiesWithFactories = \array_map(
-            static fn(ModelFactory $factory): string => $factory::getEntityClass(),
-            \iterator_to_array($factories)
+        $this->entitiesWithFactories = \array_unique(
+            \array_reduce(
+                \iterator_to_array($factories),
+                static function(array $carry, ModelFactory $factory): array {
+                    $carry[\get_class($factory)] = $factory::getEntityClass();
+
+                    return $carry;
+                },
+                []
+            )
         );
     }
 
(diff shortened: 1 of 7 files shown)


src/Bundle/Maker/MakeFactory.php
```diff
@@ -268,6 +275,32 @@ private function defaultPropertiesForPersistedObject(string $class, bool $allFie
 
         $dbType = $em instanceof EntityManagerInterface ? 'ORM' : 'ODM';
 
+        // If Factory exist for related entities populate too with auto defaults
+        foreach ($metadata->associationMappings as $item) {
```
nikophil, line 279: I think I'll provide a more robust implementation in a further PR, it's time to split this class. I think we should handle `--all-fields` here as well

tests/Fixtures/Maker/expected/can_create_factory_with_relation_defaults.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * @see https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#model-factories
+     *
+     * @todo add your default values here
+     */
+    protected function getDefaults(): array
+    {
+        return [
+            'manyToOne' => \Zenstruck\Foundry\Tests\Fixtures\Factories\CategoryFactory::new(),
```
kbond, line 47: Can we have this imported? I think a previous PR of yours played with the imports - can we easily hook into that?

nikophil: yep that's possible, but a little bit tricky: we return a `\Generator<string, string>` from `defaultPropertiesForPersistedObject()`:
```php
// generateFactory()
'defaultProperties' => $this->defaultPropertiesForPersistedObject(),

// defaultPropertiesForPersistedObject()
yield \lcfirst($fieldName) => "\\{$factoryClass}::new(),";
```

I think the cleanest solution would be to introduce some kind of "view model":
```php
class FactoryData
{
    private string $objectShortName;
    private array $uses;
    private array $defaultProperties;
    // ...
}
```

this way we could pass this instance to each functions which would modify it, until we give this object to the template which would use it to display all the needed informations.

WDYT?

kbond: Yes, that makes sense to me. Want to do that in this PR or another?

nikophil: I think a new PR will be better

PR #354, opened 2022-11-23
refactor(make:factory): use value object to render template

nikophil:
- refactor(make:factory): use a value object as template variable
- refactor(make:factory): use a value object to render phpdoc


src/Bundle/Maker/MakeFactoryPHPDocMethod.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Bundle\Maker;
+
+final class MakeFactoryPHPDocMethod
```
nikophil, line 8: this is just an experiment, I wanted to see where it goes with a VO for that. I think I prefer this instead of the complex templating stuff. WDYT?

src/Bundle/Maker/MakeFactoryPHPDocMethod.php
```diff
@@ -0,0 +1,36 @@
...
+    public function __construct(private string $prototype, private bool $returnsCollection, private bool $isStatic = true, private bool $isRepository = false)
+    {
+    }
+
+    public function render(string $objectName, bool $withPHPStanEnabled): string
+    {
+        $annotation = $withPHPStanEnabled ? 'phpstan-method' : 'method';
+        $static = $this->isStatic ? 'static' : '      ';
+
+        $proxyType = $this->isRepository ? 'RepositoryProxy' : 'Proxy';
```
nikophil, line 51: I'm pretty sure there is a problem with repository here... and I've just realized there is no test which covers a case with an entity with repository. I'm gonna fix this in this PR

src/Bundle/Maker/MakeFactoryData.php
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    public function addUse(string $use): void
+    {
+        if (!\in_array($use, $this->uses, true)) {
+            $this->uses[] = $use;
+        }
+    }
+
+    public function renderUses(): string
```
nikophil: don't know if the clearer is to return a string or an array here (same for other `render*()` methods

src/Bundle/Maker/MakeFactoryPHPDocMethod.php
kbond: I'm not following.

src/Bundle/Maker/MakeFactoryData.php
kbond: Yeah, this is getting pretty _templatey_. Thinking the rendering should be done in the tpl file.

src/Bundle/Maker/MakeFactoryData.php
```diff
@@ -0,0 +1,167 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Bundle\Maker;
+
+use Zenstruck\Foundry\ModelFactory;
+use Zenstruck\Foundry\Proxy;
+
+final class MakeFactoryData
```
kbond, line 12: Let's be sure to mark as `@internal`.

src/Bundle/Maker/MakeFactoryPHPDocMethod.php
kbond, line 8: `@internal`

src/Bundle/Maker/MakeFactoryPHPDocMethod.php
nikophil: we're not testing to create a factory for an entity which uses a repository, and my implementation has a problem, I've figured out when reviewing, that's why the PR is still a draft

kbond: Ah ok, got it

src/Bundle/Maker/MakeFactory.php
```diff
@@ -376,10 +345,47 @@ private function doctrineEnabled(): bool
...
+     * @param class-string $class
+     */
+    private function createMakeFactoryData(Generator $generator, string $class, bool $persisted): MakeFactoryData
+    {
+        $object = new \ReflectionClass($class);
+
+        if ($persisted) {
+            $repository = new \ReflectionClass($this->managerRegistry->getRepository($object->getName()));
+
+            if (\str_starts_with($repository->getName(), 'Doctrine')) {
```
nikophil, line 366: I changed this condition, which used to be:
```php
if (0 !== \mb_strpos($repository->getName(), $generator->getRootNamespace())) {
```
because the condition was always true in out tests :shrug:

src/Bundle/Maker/MakeFactoryData.php
```diff
@@ -0,0 +1,119 @@
...
+        $methodsInPHPDoc = $this->methodsInPHPDoc;
+        \usort(
+            $methodsInPHPDoc,
+            static fn(MakeFactoryPHPDocMethod $m1, MakeFactoryPHPDocMethod $m2) => $m1->sortValue() <=> $m2->sortValue()
+        );
+
+        return $methodsInPHPDoc;
+    }
+
+    public function getMethodsPHPDocPaddingValue(bool $withPhpStanEnabled = false): int
```
nikophil: we need to get this value in order to apply the padding to return types. This way they get nicely aligned.

It gets tricky because this length can change depending on object name, repository name, whether or not phpstan is used...

kbond: I'm now thinking we should remove the padding entirely. I feel it adds unnecessary complexity and makes a blind assumption about a project's cs. For instance, in my projects, my php-cs config removes the padding.

WDYT? /cc @OskarStark

nikophil: I was thinking the same thing while doing this, it adds a lot of complexity when dealing with repositories

src/Bundle/Maker/MakeFactoryData.php
```diff
@@ -0,0 +1,119 @@
+<?php
+
+declare(strict_types=1);
```
kbond: Let's remove this (not part of this project's current cs).

src/Bundle/Maker/MakeFactoryPHPDocMethod.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
```
kbond: Let's remove

kbond (changes requested):

src/Bundle/Maker/MakeFactoryData.php
kbond: Ok, let's remove.

src/Bundle/Maker/MakeFactoryData.php
nikophil: oops this is in my phpstorm template :smile:

src/Bundle/Maker/MakeFactoryData.php
nikophil: done. I've kept the one on `static`

kbond (approved):

PR #357, opened 2022-11-24
refactor(make:factory): split command with DefaultPropertiesGuesser

nikophil:
just splitting the command, using `DefaultPropertiesGuesser`

from there I'd like to:
- handle `--all-fields` option for join columns
- use existing factories for embeddables
- go a little bit further in how we guess types for "not persisted" objects
- handle `x-to-many` relationships? we could use `Factory::new()->many()`

BTW, I'm still looking for a better name than `not-persisted` :innocent:

### DIFF
--- a/src/Bundle/Maker/Factory/FactoryFinder.php
+++ b/src/Bundle/Maker/Factory/FactoryFinder.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Bundle\Maker\Factory;
+
+use Zenstruck\Foundry\ModelFactory;
+
+/**
+ * @internal
+ */
+final class FactoryFinder
+{
+    /**
+     * @var array<class-string, class-string> factory classes as keys, object class as values
+     */
+    private array $classesWithFactories;
+
+    /** @param \Traversable<ModelFactory> $factories */
+    public function __construct(\Traversable $factories)
+    {
+        /** @phpstan-ignore-next-line */
+        $this->classesWithFactories = \array_unique(
+            \array_reduce(
+                \iterator_to_array($factories, preserve_keys: true),
+                static function(array $carry, ModelFactory $factory): array {
+                    $carry[\get_class($factory)] = $factory::getEntityClass();
+
+                    return $carry;
+                },
+                []
+            )
+        );
+    }
+
+    /** @param class-string $class */
+    public function classHasFactory(string $class): bool
+    {
+        return \in_array($class, $this->classesWithFactories, true);
+    }
+
+    /**
+     * @param class-string $class
+     *
+     * @return class-string|null
+     */
+    public function getFactoryForClass(string $class): ?string
+    {
+        $factories = \array_flip($this->classesWithFactories);
+
+        return $factories[$class] ?? null;
+    }
+}
--- a/src/Bundle/Maker/Factory/DefaultPropertiesGuesser.php
+++ b/src/Bundle/Maker/Factory/DefaultPropertiesGuesser.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Zenstruck\Foundry\Bundle\Maker\Factory;
+
+/**
+ * @internal
+ */
+interface DefaultPropertiesGuesser
+{
+    public function __invoke(MakeFactoryData $makeFactoryData, bool $allFields): void;
+
+    public function supports(bool $persisted): bool;
+}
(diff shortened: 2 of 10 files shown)


src/Bundle/Maker/MakeFactory/FactoryFinder.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Bundle\Maker\MakeFactory;
+
+use Zenstruck\Foundry\ModelFactory;
+
+final class FactoryFinder
```
nikophil: not sure of this name... it's more like an index of factory classes

kbond: `FactoryRegistry`?

src/Bundle/Maker/MakeFactory/DefaultPropertiesGuesserForDoctrineObject.php
```diff
@@ -0,0 +1,121 @@
+<?php
+
+namespace Zenstruck\Foundry\Bundle\Maker\MakeFactory;
```
kbond:
```suggestion
namespace Zenstruck\Foundry\Bundle\Maker\Factory;
```

src/Bundle/Maker/MakeFactory/DefaultPropertiesGuesserInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Zenstruck\Foundry\Bundle\Maker\MakeFactory;
+
+/**
+ * @internal
+ */
+interface DefaultPropertiesGuesserInterface
```
kbond:
```suggestion
interface DefaultPropertiesGuesser
```

I like to avoid suffixes

src/Bundle/Maker/MakeFactory/DefaultPropertiesGuesserForDoctrineObject.php
```diff
@@ -0,0 +1,121 @@
...
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata as ODMClassMetadata;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Mapping\ClassMetadataInfo as ORMClassMetadata;
+use Doctrine\Persistence\ManagerRegistry;
+use Doctrine\Persistence\ObjectManager;
+
+/**
+ * @internal
+ */
+class DefaultPropertiesGuesserForDoctrineObject implements DefaultPropertiesGuesserInterface
```
kbond:
```suggestion
class DoctrineObjectDefaultPropertiesGuesser implements DefaultPropertiesGuesserInterface
```

src/Bundle/Maker/MakeFactory/DefaultPropertiesGuesserForNotPersistedObject.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Zenstruck\Foundry\Bundle\Maker\MakeFactory;
+
+/**
+ * @internal
+ */
+class DefaultPropertiesGuesserForNotPersistedObject implements DefaultPropertiesGuesserInterface
```
kbond:
```suggestion
class ObjectDefaultPropertiesGuesser implements DefaultPropertiesGuesserInterface
```

kbond (changes requested):
Nice cleanup of the maker!

Just a few suggestions.

src/Bundle/Maker/MakeFactory/FactoryFinder.php
nikophil: I'm not sure... it does not even contain actual factory, but just a map with [factory class => object class]

src/Bundle/Maker/MakeFactory/DefaultPropertiesGuesserInterface.php
nikophil: cool, I don't like `Interface` neither, but people are hard to convince!

src/Bundle/Maker/MakeFactory/FactoryFinder.php
kbond: Fair, we can always rename later as it's internal

src/Bundle/Maker/Factory/FactoryFinder.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Bundle\Maker\Factory;
+
+use Zenstruck\Foundry\ModelFactory;
+
+final class FactoryFinder
```
kbond, line 12: Let's mark internal

kbond (changes requested):

src/Bundle/Maker/Factory/FactoryFinder.php
nikophil: oups

kbond (approved):

PR #363, opened 2022-11-24
chore: rename service ids

### LINKED ISSUE #360: Rename service IDs
[Ref](https://github.com/zenstruck/foundry/blob/20ac3499c224c7ad4f5fe3e5f4da04ad2330ea5d/src/Bundle/Resources/config/services.xml)

I recently learned that if you id a services with the class name, it makes it autowireable. We don't want this for foundry's services. Additionally, if you prefix the service id with `.`, it marks it as [_hidden_](https://symfony.com/blog/new-in-symfony-4-1-hidden-services) so users will not see them in `debug:container` by default.

kbond:
Fixes #360


wouterj:
Note that technically speaking, this is a BC break. You can add an alias from FQCN to the new service ID and deprecate that alias to avoid the break.

kbond:
I thought of this but since all these classes are marked `@internal` I figured we were ok.

PR #364, opened 2022-11-24
feat(make:factory): use existing factories for embeddables (ODM + ORM)

nikophil:
Use existing factories for embeddables


src/Bundle/Maker/Factory/DoctrineObjectDefaultPropertiesGuesser.php
```diff
@@ -56,66 +57,125 @@ public function __invoke(MakeFactoryData $makeFactoryData, bool $allFields): voi
...
+    /** @param class-string $fieldClass */
+    private function addDefaultValueUsingFactory(MakeFactoryData $makeFactoryData, string $fieldName, string $fieldClass, bool $isMultiple = false): void
+    {
+        if (!$factoryClass = $this->factoryFinder->getFactoryForClass($fieldClass)) {
+            $makeFactoryData->addDefaultProperty(\lcfirst($fieldName), "null, // TODO add {$fieldClass} type manually");
+
+            return;
+        }
+
+        $factoryMethod = $isMultiple ? 'new()->many(5)' : 'new()';
```
nikophil: `5` is a totally arbitrary int...

tests/Fixtures/Entity/EntityForRelations.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Zenstruck\Foundry\Tests\Fixtures\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="entity_for_relations")
+ */
+class EntityForRelations
```
nikophil, line 11: this one was needed to have a owning side for the `OneToMany`

tests/Fixtures/Kernel.php
```diff
@@ -34,14 +38,18 @@ class Kernel extends BaseKernel
 
     private string $ormResetMode = ORMDatabaseResetter::RESET_MODE_SCHEMA;
 
+    private array $factoriesRegistered = [];
```
nikophil, line 41: I had a case where sometimes we want the factory registered in container, and sometimes not. So now, we can chose in the test which factory we need to be registered.

tests/Fixtures/Maker/expected/can_create_factory_for_odm_with_data_set_document.php
```diff
@@ -45,11 +46,10 @@ protected function getDefaults(): array
     {
         return [
             'body' => self::faker()->text(),
-            'comments' => null, // TODO add MANY ODM type manually
             'createdAt' => self::faker()->dateTime(),
             'title' => self::faker()->text(),
-            'user' => null, // TODO add ONE ODM type manually
-            'viewCount' => null, // TODO add INT ODM type manually
```
nikophil, line 52: this was actually a bug :roll_eyes:

tests/Fixtures/Maker/expected/can_create_factory_with_embeddable_with_data_set_odm.php
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * @see https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#model-factories
+     *
+     * @todo add your default values here
+     */
+    protected function getDefaults(): array
+    {
+        return [
+            'body' => self::faker()->text(),
+            'comments' => CommentFactory::new()->many(5),
```
nikophil, line 50: I'm wondering something: if in `CommandFactory` we have
```
{
    return [
        'post' => PostFactory::new()
    ];
}
```
won't we have an infinite loop? or maybe you've already fixed that case?

src/Bundle/Maker/Factory/DoctrineObjectDefaultPropertiesGuesser.php
```diff
@@ -56,66 +57,125 @@ public function __invoke(MakeFactoryData $makeFactoryData, bool $allFields): voi
...
         $metadata = $em->getClassMetadata($class);
 
-        $dbType = $em instanceof EntityManagerInterface ? 'ORM' : 'ODM';
-
-        $this->guessDefaultValueForAssociativeFields($makeFactoryData, $metadata, $dbType, $allFields);
-        $this->guessDefaultValueForScalarFields($makeFactoryData, $metadata, $dbType, $allFields);
+        $metadata instanceof ORMClassMetadata && $this->guessDefaultValueForORMAssociativeFields($makeFactoryData, $metadata, $allFields);
+        $metadata instanceof ODMClassMetadata && $this->guessDefaultValueForODMAssociativeFields($makeFactoryData, $metadata, $allFields);
+        $metadata instanceof ORMClassMetadata && $this->guessDefaultValueForEmbedded($makeFactoryData, $metadata, $allFields);
+        $this->guessDefaultValueForScalarFields($makeFactoryData, $metadata, $allFields);
```
nikophil: maybe another clean strategy would have been to introduce `{ODM,ORM}DefaultGuesser`, WDYT? :thinking:

tests/Fixtures/Maker/expected/can_create_factory_with_embeddable_with_data_set_odm.php
kbond: I don't believe this is accounted for.

src/Bundle/Maker/Factory/DoctrineObjectDefaultPropertiesGuesser.php
kbond: I'm thinking we should not have `--all-fields` populate `OneToMany` relationships.

This can cause problems when testing _count_'s of entities in the db. You can get a greater count than you expect. @OskarStark ran into this problem just the other day.

WDYT?

OskarStark: Yes it can create side effects, because it can also nest many more entities downwards

tests/Fixtures/Maker/expected/can_create_factory_with_embeddable_with_data_set_odm.php
kbond: When we get to auto-creating factories for relationships this will be an issue as well (ie self referencing relationships)

nikophil: follow up [here](https://github.com/zenstruck/foundry/pull/364#pullrequestreview-1193542616) since it is the same problem

src/Bundle/Maker/Factory/DoctrineObjectDefaultPropertiesGuesser.php
nikophil: what about if we say `Factory::new(['propertyName' => $this])`?

nikophil: However, this is what we do in this test:
https://github.com/zenstruck/foundry/blob/f43b06708608f9f46691183d4417eebe967fbc94/tests/Functional/ORMModelFactoryTest.php#L65-L75

even if I add this code, the test still passes:
```php
        foreach (CommentFactory::all() as $comment) {
            self::assertSame($post->object(), $comment->getPost());
        }
```

in which ways this code could be problematic?

tests/Fixtures/Maker/expected/can_create_factory_with_embeddable_with_data_set_odm.php
nikophil: > I don't believe this is accounted for.

same as in https://github.com/zenstruck/foundry/pull/364#pullrequestreview-1193542616 we already test such case, which works fine. 

in `Factory::create()`, `$this->normalizeAttribute($value, $name)` returns null, but it adds an `afterPersist` callback which forces `Comment::$post` to be the current `$post`. This way, there is no infinite loop, and that's how the current object is set to the object owner of the relation.

It took me some time to understand _how_ this works :sweat_smile: - but now I understand things better :blush:

src/Bundle/Maker/Factory/DoctrineObjectDefaultPropertiesGuesser.php
kbond: It's totally legit to create them this way in a test. But when say 10 are added via the `getDefaults()`. In a test, you do something that creates 2 entities, then assert the count is 2, the actual count could be 12 (or more depending on your setup).

Sorry, I'm having a hard time explaining.

kbond: We are all on the same page now. We're going to restrict `--all-fields` to _nullable-non-relationships_.

tests/Fixtures/Maker/expected/can_create_factory_with_embeddable_with_data_set_odm.php
kbond: About self-referencing relationships, it isn't possible to have a non-nullable self-referencing relationship, is it?

nikophil: hmmm :thinking: 

nothing stops us to do it, even if I cannot find any valid usecase
we even have one in our tests : https://github.com/zenstruck/foundry/blob/1.x/tests/Fixtures/Entity/EntityWithRelations.php#L23

kbond: That example isn't self-referencing though

nikophil: :facepalm: 

you're right, sorry for that.
I've tried a case with self-referencing, non-nullable relationship, and nothing stops from doing this and doctrine validator says everything is OK.

do you want that we make something special for this case?

kbond: Nah, was just curious.

kbond (approved):
Looks good! I guess this opens the door to the final step: auto creating non-existent factories?

kbond:
Dep issue? Do we need to depend on the property info component?

tests/Functional/Bundle/Maker/MakeFactoryTest.php
```diff
@@ -298,12 +302,12 @@ public function can_customize_namespace_with_test_flag_with_root_namespace_prefi
     /**
      * @test
      */
-    public function can_create_factory_with_all_interactively(): void
+    public function can_create_all_factories_for_doctrine_objects(): void
     {
         $tester = new CommandTester((new Application(self::bootKernel()))->find('make:factory'));
 
+        $this->assertFileDoesNotExist(self::tempFile('src/Factory/BrandFactory.php'));
```
nikophil, line 309: we have a problem when different factories will have the same name: `ODM/Category.php` and `ORM/Category.php` will reasult in the same `Factory/CategoryFactory.php` file, then the second one will create an error. 

I've updated https://github.com/zenstruck/foundry/issues/336 to handle this case

tests/Functional/Bundle/Maker/MakeFactoryTest.php
```diff
@@ -298,12 +302,12 @@ public function can_customize_namespace_with_test_flag_with_root_namespace_prefi
...
      * @test
      */
-    public function can_create_factory_with_all_interactively(): void
+    public function can_create_all_factories_for_doctrine_objects(): void
     {
         $tester = new CommandTester((new Application(self::bootKernel()))->find('make:factory'));
 
         $this->assertFileDoesNotExist(self::tempFile('src/Factory/CategoryFactory.php'));
-        $this->assertFileDoesNotExist(self::tempFile('src/Factory/PostFactory.php'));
+        $this->assertFileDoesNotExist(self::tempFile('src/Factory/TagFactory.php'));
```
nikophil, line 310: we have a problem when different factories will have the same name: `ODM/Category.php` and `ORM/Category.php` will reasult in the same `Factory/CategoryFactory.php` file, then the second one will create an error. 

I've updated https://github.com/zenstruck/foundry/issues/336 to handle this case

nikophil:
PR is finally green :)

are you OK to keep `symfony/property-info` as a dependency?

kbond (approved):
> are you OK to keep symfony/property-info as a dependency?

Is there an alternative?

It's a bit annoying as it's only needed for the maker (and a user might not have the maker-bundle installed). Since foundry itself is a dev dep, I'm ok with keeping.

nikophil:
I'll have a try to find an alternative :)

kbond (approved):

PR #372, opened 2022-12-02
feat(make:factory): auto create missing factories

nikophil:
:warning: based on https://github.com/zenstruck/foundry/pull/374

Here is a first draft which auto creates "missing" factories. 
We only create a missing factory if its related field if `ManyToOne` or `OneToOne` and not nullable.

I don't compute any "graph" for the relationships, I don't think it is needed, even with the `All` option.

The CI only passes for "ODM only" tests because in `ORM` cases we have multiple entities called `Category`, resulting to the same file name, and we still don't handle this case. I'm afraid I'll have to fix this bug before finishing this PR.

I also need to manually test this new behavior in a project with some entities and relationships

### DIFF
--- a/src/Bundle/Maker/Factory/AbstractDoctrineDefaultPropertiesGuesser.php
+++ b/src/Bundle/Maker/Factory/AbstractDoctrineDefaultPropertiesGuesser.php
@@ -1,46 +1,54 @@
 <?php
 
-declare(strict_types=1);
-
 namespace Zenstruck\Foundry\Bundle\Maker\Factory;
 
 use Doctrine\Persistence\ManagerRegistry;
 use Doctrine\Persistence\Mapping\ClassMetadata;
-use Doctrine\Persistence\ObjectManager;
+use Symfony\Component\Console\Style\SymfonyStyle;
 
-/** @internal  */
+/**
+ * @internal
+ */
 abstract class AbstractDoctrineDefaultPropertiesGuesser implements DefaultPropertiesGuesser
 {
-    public function __construct(protected ManagerRegistry $managerRegistry, private FactoryFinder $factoryFinder)
+    public function __construct(protected ManagerRegistry $managerRegistry, private FactoryClassMap $factoryClassMap, private FactoryGenerator $factoryGenerator)
     {
     }
 
     /** @param class-string $fieldClass */
-    protected function addDefaultValueUsingFactory(MakeFactoryData $makeFactoryData, string $fieldName, string $fieldClass, bool $isMultiple = false): void
+    protected function addDefaultValueUsingFactory(SymfonyStyle $io, MakeFactoryData $makeFactoryData, MakeFactoryQuery $makeFactoryQuery, string $fieldName, string $fieldClass): void
     {
-        if (!$factoryClass = $this->factoryFinder->getFactoryForClass($fieldClass)) {
-            $makeFactoryData->addDefaultProperty(\lcfirst($fieldName), "null, // TODO add {$fieldClass} type manually");
+        if (!$factoryClass = $this->factoryClassMap->getFactoryForClass($fieldClass)) {
+            if ($makeFactoryQuery->generateAllFactories() || $io->confirm("A factory for class \"{$fieldClass}\" is missing for field {$makeFactoryData->getObjectShortName()}::\${$fieldName}. Do you want to create it?")) {
+                $factoryClass = $this->factoryGenerator->generateFactory($io, $makeFactoryQuery->withClass($fieldClass));
+            } else {
+                $makeFactoryData->addDefaultProperty(\lcfirst($fieldName), "null, // TODO add {$fieldClass} type manually");
 
-            return;
+                return;
+            }
         }
 
-        $factoryMethod = $isMultiple ? 'new()->many(5)' : 'new()';
+        $makeFactoryData->addUse($factoryClass);
 
-        $factory = new \ReflectionClass($factoryClass);
-        $makeFactoryData->addUse($factory->getName());
-        $makeFactoryData->addDefaultProperty(\lcfirst($fieldName), "{$factory->getShortName()}::{$factoryMethod},");
+        $factoryShortName = \mb_substr($factoryClass, \mb_strrpos($factoryClass, '\\') + 1);
+        $makeFactoryData->addDefaultProperty(\lcfirst($fieldName), "{$factoryShortName}::new(),");
     }
 
     protected function getClassMetadata(MakeFactoryData $makeFactoryData): ClassMetadata
     {
         $class = $makeFactoryData->getObjectFullyQualifiedClassName();
 
-        $em = $this->managerRegistry->getManagerForClass($class);
+        foreach ($this->managerRegistry->getManagers() as $manager) {
+            try {
+                $classMetadata = $manager->getClassMetadata($class);
+            } catch (\Throwable) {
+            }
+        }
 
-        if (!$em instanceof ObjectManager) {
+        if (!isset($classMetadata)) {
             throw new \InvalidArgumentException("\"{$class}\" is not a valid Doctrine class name.");
         }
 
-        return $em->getClassMetadata($class);
+        return $classMetadata;
     }
 }
--- a/src/Bundle/Maker/Factory/FactoryGenerator.php
+++ b/src/Bundle/Maker/Factory/FactoryGenerator.php
@@ -1,0 +1,113 @@
+<?php
+
+namespace Zenstruck\Foundry\Bundle\Maker\Factory;
+
+use Doctrine\Persistence\ManagerRegistry;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/**
+ * @internal
+ */
+final class FactoryGenerator
+{
+    /** @param \Traversable<int, DefaultPropertiesGuesser> $defaultPropertiesGuessers */
+    public function __construct(private ManagerRegistry $managerRegistry, private KernelInterface $kernel, private \Traversable $defaultPropertiesGuessers, private FactoryClassMap $factoryClassMap)
+    {
+    }
+
+    /**
+     * @return class-string The factory's FQCN
+     */
+    public function generateFactory(SymfonyStyle $io, MakeFactoryQuery $makeFactoryQuery): string
+    {
+        $class = $makeFactoryQuery->getClass();
+        $generator = $makeFactoryQuery->getGenerator();
+
+        if (!\class_exists($class)) {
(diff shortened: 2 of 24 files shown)


src/Bundle/Maker/Factory/AbstractDoctrineDefaultPropertiesGuesser.php
```diff
@@ -11,24 +11,23 @@
...
         if (!$factoryClass = $this->factoryFinder->getFactoryForClass($fieldClass)) {
-            $makeFactoryData->addDefaultProperty(\lcfirst($fieldName), "null, // TODO add {$fieldClass} type manually");
-
-            return;
+            $factoryClass = $this->factoryGenerator->generateFactory($fieldClass, $makeFactoryQuery);
         }
 
         $factoryMethod = $isMultiple ? 'new()->many(5)' : 'new()';
 
-        $factory = new \ReflectionClass($factoryClass);
```
nikophil, line 29: the class `$factoryClass` may not exist at this point, can't use reflection

src/Bundle/Maker/Factory/FactoryGenerator.php
```diff
@@ -0,0 +1,107 @@
...
+    public function __construct(private ManagerRegistry $managerRegistry, private KernelInterface $kernel, private \Traversable $defaultPropertiesGuessers, private FactoryFinder $factoryFinder)
+    {
+    }
+
+    /**
+     * @param class-string $class
+     *
+     * @return class-string The factory's FQCN
+     */
+    public function generateFactory(string $class, MakeFactoryQuery $makeFactoryQuery): string
```
nikophil: I didn't wanted to rely on `InputInterface` everywhere, that's why I've introduced this `MakeFactoryQuery` class

src/Bundle/Maker/Factory/FactoryGenerator.php
```diff
@@ -0,0 +1,107 @@
...
+            $this->guessNamespace($generator, $makeFactoryQuery->getNamespace(), $makeFactoryQuery->isTest()),
+            'Factory'
+        );
+
+        /** @var class-string $factoryClass */
+        $factoryClass = $factory->getFullName();
+
+        if (!$this->factoryFinder->classHasFactory($class)) {
+            $this->factoryFinder->addFactoryForClass($factoryClass, $class);
+        }
```
nikophil: here is the trick: we add the factory class to the finder before creating it. So we assume it exists for the rest of the execution.

I'm still looking for a good name for this class... It feels like we're adding a _real_ factory here, but we're only adding a factory name to a listing...
`FactoriesClassesList`? `ClassFactoryMap` ?

src/Bundle/Maker/Factory/FactoryGenerator.php
```diff
@@ -0,0 +1,107 @@
...
+
+namespace Zenstruck\Foundry\Bundle\Maker\Factory;
+
+use Doctrine\Persistence\ManagerRegistry;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/** @internal */
+final class FactoryGenerator
```
nikophil, line 14: 90% of this file was extracted from `MakeFactory.php`

src/Bundle/Maker/Factory/AbstractDoctrineDefaultPropertiesGuesser.php
```diff
@@ -11,24 +9,23 @@
...
 
     /** @param class-string $fieldClass */
-    protected function addDefaultValueUsingFactory(MakeFactoryData $makeFactoryData, string $fieldName, string $fieldClass, bool $isMultiple = false): void
+    protected function addDefaultValueUsingFactory(MakeFactoryData $makeFactoryData, MakeFactoryQuery $makeFactoryQuery, string $fieldName, string $fieldClass, bool $isMultiple = false): void
     {
         if (!$factoryClass = $this->factoryFinder->getFactoryForClass($fieldClass)) {
-            $makeFactoryData->addDefaultProperty(\lcfirst($fieldName), "null, // TODO add {$fieldClass} type manually");
-
-            return;
+            $factoryClass = $this->factoryGenerator->generateFactory($fieldClass, $makeFactoryQuery);
```
nikophil: so, here, when we don't know about a factory, we create it.

I'm wondering if we shouldn't ask the user if he/she really wants to create it?
This would mean we should bring the `$io` till here

src/Bundle/Maker/Factory/MakeFactoryQuery.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        return $this->allFields;
+    }
+
+    public function getClass(): string
+    {
+        return $this->class;
+    }
+
+    public function getGenerator(): Generator
```
nikophil, line 60: I'm not 100% happy with the generator in this class, I think I'll take it out, but it was easier when I created the feature

EDIT: it still helps to prevent method calls with too many parameters

src/Bundle/Maker/Factory/AbstractDoctrineDefaultPropertiesGuesser.php
kbond: Yeah, we should probably confirm (unless they passed `--all`)

nikophil: done :)

src/Bundle/Maker/Factory/AbstractDoctrineDefaultPropertiesGuesser.php
```diff
@@ -1,46 +1,54 @@
...
+        $factoryShortName = \mb_substr($factoryClass, \mb_strrpos($factoryClass, '\\') + 1);
+        $makeFactoryData->addDefaultProperty(\lcfirst($fieldName), "{$factoryShortName}::{$factoryMethod},");
     }
 
     protected function getClassMetadata(MakeFactoryData $makeFactoryData): ClassMetadata
     {
         $class = $makeFactoryData->getObjectFullyQualifiedClassName();
 
-        $em = $this->managerRegistry->getManagerForClass($class);
+        foreach ($this->managerRegistry->getManagers() as $manager) {
```
nikophil, line 41: I had to do this move because this code returns null:
```php
$class = $this->managerRegistry->getManagerForClass(OrmEmbedddableClass::class);
```

but this one works
```
$theRightManager->getClassMetadata(OrmEmbedddableClass::class);
```

src/Bundle/Maker/MakeFactory.php
```diff
@@ -80,6 +75,10 @@ public function interact(InputInterface $input, ConsoleStyle $io, Command $comma
             $io->newLine();
         }
 
+        if ($input->getArgument('class')) {
```
nikophil, line 80: I think previous notes should be displayed even if a class is passed as an argument to the command

src/Bundle/Maker/Factory/FactoryGenerator.php
kbond: _Registry_? I think I may have suggested this before and yeah, not really a registry...

src/Bundle/Maker/Factory/AbstractDoctrineDefaultPropertiesGuesser.php
```diff
@@ -1,46 +1,54 @@
...
+                $factoryClass = $this->factoryGenerator->generateFactory($io, $makeFactoryQuery->withClass($fieldClass));
+            } else {
+                $makeFactoryData->addDefaultProperty(\lcfirst($fieldName), "null, // TODO add {$fieldClass} type manually");
 
-            return;
+                return;
+            }
         }
 
         $factoryMethod = $isMultiple ? 'new()->many(5)' : 'new()';
```
kbond: Weren't we going to remove this?

src/Bundle/Maker/Factory/FactoryGenerator.php
nikophil: haha yes you've already suggested, but I don't want to call it _Registry_ because I think it would be confusing

if you don't think it's strange, I'm gonna go with `FactoryWithClassMap` or `FactoryClassMap`

src/Bundle/Maker/Factory/AbstractDoctrineDefaultPropertiesGuesser.php
nikophil: we removed it for "multiple" relationship (ie: `{One|Many}ToMany`)

but we could still have [multiple `EmbeddedDocument` for ODM](https://github.com/zenstruck/foundry/blob/1.x/tests/Fixtures/Document/Post.php#L49-L54). There is no risk to have circular dependencies here.

If you prefer, I can remove this one, since it's a collection, it is not mandatory

kbond: Ah ok, maybe we should remove just for consistency?

kbond: But don't worry about it for this PR.

src/Bundle/Maker/Factory/FactoryGenerator.php
kbond: Yep FactoryClassMap sounds fine

src/Bundle/Maker/Factory/FactoryGenerator.php
```diff
@@ -0,0 +1,111 @@
...
+
+namespace Zenstruck\Foundry\Bundle\Maker\Factory;
+
+use Doctrine\Persistence\ManagerRegistry;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Bundle\MakerBundle\Generator;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/** @internal */
```
kbond: Can you expand these docblocks?

src/Bundle/Maker/Factory/FactoryGenerator.php
nikophil: already removed :sweat_smile:

src/Bundle/Maker/Factory/FactoryGenerator.php
nikophil: I was thinking it was part of the cs-fixer rules.
done.

src/Bundle/Maker/Factory/AbstractDoctrineDefaultPropertiesGuesser.php
nikophil: already resolved :sweat_smile:

src/Bundle/Maker/Factory/FactoryGenerator.php
kbond: I thought so as well.

kbond (approved):

PR #374, opened 2022-12-04
refactor: set all fixtures class name unique

nikophil:
change class' name of some fixtures in order to prevent [this problem](https://github.com/zenstruck/foundry/blob/v1.24.1/tests/Functional/Bundle/Maker/MakeFactoryTest.php#L312-L315). 

In the future, in case of collision, I'd like to ask to the user the factory's name


kbond:
Is this one ready?

kbond (approved):

Issue #358, opened 2022-11-24
Makefile improvements


nikophil:
- [x] find a way to easily add `--stop-on-failure` (I'm always adding it to the `Makefile` itself, and commit the file by mistake) https://github.com/zenstruck/foundry/pull/376
- [x] provide an easy way to switch phpunit context (use dama, use orm, use odm)  https://github.com/zenstruck/foundry/pull/375
- [x] fix `make sca` and `make fixcs` which always try to update composer https://github.com/zenstruck/foundry/pull/376
- [x] change php version with an env var https://github.com/zenstruck/foundry/pull/376

kbond:
Is it possible to have `make test ...` pass all args to phpunit?

```
make test tests/foo/bar
make test --exclude-group foo
```

nikophil:
I would love this
I have to dig in a little bit but I'l afraid options won't work

nikophil:
:tada:

PR #375, opened 2022-12-06
chore: change test context with .env

nikophil:
related to https://github.com/zenstruck/foundry/issues/358 (but it does not fixes everything)

I wanted to make this move from a long time: now, when a problem occurs in a specific case, it is possible to easily reproduce the error in local with the same context.

I'd also like to provide a way to easily change the php version used :star_struck:

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,7 @@ Each target will build and start the docker stack and install composer only if n
 ```shell
 $ make help
 validate                       Run fixcs, sca, full test suite and validate migrations
-test-full                      Run full PHPunit (MySQL + Mongo)
-test-fast                      Run PHPUnit with SQLite
-test-mysql                     Run PHPUnit with MySQL
-test-postgresql                Run PHPUnit with PostgreSQL
-test-mongo                     Run PHPUnit with Mongo
+test                           Run PHPUnit tests suite
 fixcs                          Run PHP-CS-Fixer
 sca                            Run static analysis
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -92,11 +98,6 @@ create a server called `FOUNDRY` in your PHP Remote Debug, with the IDE key `xde
 
 ![PhpStorm with xdebug](docs/phpstorm-xdebug-config.png)
 
-### Run tests without docker
-
-If for any reason docker is not available on your computer, the target `make test-fast` will run tests with your local
-php version, and sqlite will be used as database. Results may differ from the CI!
-
```
nikophil, line 99: I removed the doc about this because the target `make test-fast` does not exist anymore.

I'd say I'll add it back if someone asks for it :innocent:

README.md
```diff
@@ -62,16 +58,26 @@ composer                       Run composer command
 clear                          Start from a fresh install (needed if vendors have already been installed with another php version)
 ```
 
-You can run each `test-*` target with a special argument `filter`:
+You can run each `make test` target with a special argument `filter`:
 ```shell
 $ make test-mysql filter=FactoryTest
```
kbond:
```suggestion
$ make test filter=FactoryTest
```

kbond (changes requested):
Very nice!

kbond (approved):

PR #376, opened 2022-12-07
chore: availability to chose php version

nikophil:


docker/build.sh
```diff
@@ -0,0 +1,49 @@
...
+        echo "Error: action \"${1}\" invalid. Allowed actions are push|load"
+        exit 1;
+    fi
+
+    # ensure default builder is used, it is needed to benefit from cache layering
+    docker buildx use default
+
+    build "${ACTION}" 8.0
+#    build "${ACTION}" 8.1
+#    build "${ACTION}" 8.2 8.2.0RC7
```
kbond: Don't need the RC anymore!

nikophil: still needed, the official image is not released
https://hub.docker.com/_/php/tags?page=1&name=8.2

nikophil: now it's good!

Makefile
```diff
@@ -27,99 +28,93 @@ endif
...
-ifeq ($(CI),'false')
-	@$(DOCKER_COMPOSE) build  --build-arg UID="${UID}" --build-arg XDEBUG_HOST="${XDEBUG_HOST}"
-	@$(DOCKER_COMPOSE) up --detach --remove-orphans
-else
-	@$(DOCKER_COMPOSE) up --detach --no-build
-endif
-	@$(DOCKER_COMPOSE) ps
-endif
+	@if [ -z "${DOCKER_IS_UP}" ]; then \
+		./docker/build.sh load "${PHP_VERSION}" ; \
```
nikophil, line 98: we target only the current PHP_VERSION, thus the php images are built/up on demand

Makefile
```diff
@@ -27,99 +28,93 @@ endif
...
-	@${DOCKER_PHP} bin/tools/phpstan/vendor/phpstan/phpstan/phpstan analyse
+sca: bin/tools/phpstan/vendor ### Run static analysis
+	@${DOCKER_PHP} php -d 'xdebug.mode=off' bin/tools/phpstan/vendor/phpstan/phpstan/phpstan analyse
 
-bin/tools/phpstan/vendor: vendor bin/tools/phpstan/composer.json $(wildcard bin/tools/phpstan/composer.lock)
-	@${DOCKER_PHP} composer bin phpstan install
+bin/tools/phpstan/vendor: bin/tools/phpstan/composer.json bin/tools/phpstan/composer.lock
+	@$(MAKE) --no-print-directory docker-start
+	@$(MAKE) --no-print-directory vendor
+	@$(MAKE) --no-print-directory composer bin phpstan install
```
nikophil, line 73: - removing `docker-start` and `vendor` as requirement of the make target + calling them _inside_ the target improve `make sca` and `make fixcs` which no more call `composer install` everytime
- I've manage to disable xdebug for composer / phpstan / php cs fixer targets

docker/build.sh
```diff
@@ -0,0 +1,53 @@
...
+        XDEBUG_HOST="172.17.0.1"
+    elif [ "${UNAME}" = "Darwin" ]; then
+        XDEBUG_HOST="host.docker.internal"
+    fi
+
+    docker pull ${DOCKER_IMAGE_TAG} || true
+
+    DOCKER_BUILDKIT=1 docker buildx build \
+        --tag "${DOCKER_IMAGE_TAG}" \
+        --cache-from "${DOCKER_IMAGE_TAG}" \
```
nikophil, line 26: using `docker buildx` make the magic of docker cache: first, we pull the image, then we take cache from it. Then we benefit from docker cache layering AND we also build if Dockerfile changed

PR #382, opened 2022-12-14
chore: makefile improvements

nikophil:
This is part of my quest for the "Ultimate Makefile"©  :smile: 

after a small chat with @theofidry I've realized we can rely on file system to know if docker is up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
-.PHONY: $(filter-out vendor bin/tools/phpstan/vendor %,$(MAKECMDGOALS))
-
 .DEFAULT_GOAL := help
+
+SHELL=/bin/bash
 
 # DB variables
 MYSQL_URL="mysql://root:root@mysql:3306/zenstruck_foundry?charset=utf8"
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -61,8 +61,10 @@ clear                          Start from a fresh install (needed if vendors hav
 
 Use double-dash to pass any PHPUnit options or arguments with `make`:
 ```shell
-$ make test -- --filter=FactoryTest
 $ make test -- --stop-on-failure
+$ make test -- --filter FactoryTest
+# don't use "=" options value. ie: don't do this:
+$ make test -- --filter=FactoryTest
```
nikophil, line 66: I've struggled with this, and discovered `make test -- --filter=FactoryTest` won't work although the other way works
sometimes Makefiles works in mysterious ways  :shrug:

Makefile
```diff
@@ -96,6 +98,7 @@ DOCKER_IS_UP = $(shell $(DOCKER_COMPOSE) ps | grep "${PHP}" | grep "Up")
 docker-start: ### Build and run containers
 	@if [ -z "${DOCKER_IS_UP}" ]; then \
 		./docker/build.sh load "${PHP_VERSION}" ; \
+		echo -e "\nStarting containers. This could take up to one minute.\n" ; \
```
nikophil: health check could take some time, which can be disturbing

kbond (approved):

PR #383, opened 2022-12-14
feat(make:factory): use autocompletion for no persistence classes

nikophil:
related to #336 

When no `--no-persistence` is passed we ask for a FQCN, which could be cumbersome to write.

Then this PRs proposes as autocomplete all classes found in compoer's PSR-4 config

Peek 14-12-2022 09-01

### DIFF
--- a/src/Bundle/Maker/Factory/NoPersistanceObjectsAutoCompleter.php
+++ b/src/Bundle/Maker/Factory/NoPersistanceObjectsAutoCompleter.php
@@ -1,0 +1,98 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Bundle\Maker\Factory;
+
+final class NoPersistanceObjectsAutoCompleter
+{
+    public function __construct(private string $kernelRootDir)
+    {
+    }
+
+    public function getAutocompleteValues(): array
+    {
+        $classes = [];
+
+        $excludedFiles = $this->excludedFiles();
+
+        foreach ($this->getDefinedNamespaces() as $namespacePrefix => $rootFragment) {
+            $allFiles = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($rootPath = "{$this->kernelRootDir}/{$rootFragment}"));
+
+            /** @var \SplFileInfo $phpFile */
+            foreach (new \RegexIterator($allFiles, '/\.php$/') as $phpFile) {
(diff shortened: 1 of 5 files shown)


src/Bundle/Maker/Factory/NoPersistanceObjectsAutoCompleter.php
```diff
@@ -0,0 +1,87 @@
...
+{
+    public function __construct(private string $kernelRootDir)
+    {
+    }
+
+    public function getAutocompleteValues(): array
+    {
+        $classes = [];
+
+        $excludedFiles = $this->excludedFiles();
```
nikophil, line 17: the file `src/functions.php` was problematic: when transforming this file into its PSR4 representation, if I call `class_exists()`, I got an error I can't even catch as a `\Throwable` :shrug: 

```
Fatal error: Cannot redeclare Zenstruck\Foundry\factory() (previously declared in /app/src/functions.php:16) in /app/src/functions.php on line 16
```

then I decided to exclude all files found in composer's config under `autoload.files` node

tests/Unit/Bundle/Maker/Factory/NoPersistanceObjectsAutoCompleterTest.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Tests\Unit\Bundle\Maker\Factory;
+
+use PHPUnit\Framework\TestCase;
+use Zenstruck\Foundry\Bundle\Maker\Factory\NoPersistanceObjectsAutoCompleter;
+
+final class NoPersistanceObjectsAutoCompleterTest extends TestCase
```
nikophil, line 10: I don't think (but I may be wrong) that it is possible to test this autocompletion behavior, even with `CommandCompletionTester`, this is why I used a unit test

kbond:
Neat! Is the autocompletion "fuzzy" (can it match any part of the class name or just "starts with")?

nikophil:
I would like to do this, but I couldn't: the autocompleter does not allow it. 

you can even pass a callback, but it does not take any argument
see https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Console/Question/Question.php#L134

nikophil:
Hi @kbond 

any chance to see this validated/merged at some point? I don't know where I can simplify the code here :disappointed:

kbond (approved):

PR #386, opened 2022-12-16
CI: sync meta files and automate php-cs-fixer

kbond:
This is part of an initiative to synchronize config files across zenstruck/* repositories to ease maintenance. Following our style guide is no longer a requirement for PRs - it's automatically fixed once merged.

e0944bb6d31ab07b2fc03beca006737260e359d6 is the relevant commit in this PR. It does the following:
- removes the php-cs-fixer from bin/tools and Makefile
- adds a `fixcs` gha job to automatically make cs-fixes once a PR is merged
- adjusts the `.php-cs-fixer.dist.php` config to pull from a [primary source](https://github.com/zenstruck/.github/blob/main/.php-cs-fixer.dist.php).
- adds a `sync-with-template` gha job to sync "meta files" (.editorconfig, .codecov.yml etc) from a primary source (`zenstruck/php-template`).


nikophil (approved):
:+1:

this is very cool :ok_hand: 

the only thing I'm not sure is that in some very rare occasions, csfixer could either make some unwanted updates or even break stuff, but nothing would warn us about that :thinking: 

another thing: I was a little bit surprised to see how your `.github` is made (with "standelone" branches). There is a way to host multiple GitHub actions/workflows in the same branch (ie: `uses: zenstruck/.github/actions/php-cs-fixer@v1` / `uses: zenstruck/.github/actions/sync-with-template@v1`)

kbond:
> the only thing I'm not sure is that in some very rare occasions, csfixer could either make some unwanted updates or even break stuff, but nothing would warn us about that 

Yeah, that is a concern. It could also, after an update, apply cs changes we don't want with no way to undo (without reverting). Let's see if this bites us.

> another thing: I was a little bit surprised to see how your .github is made (with "standelone" branches).

Are you able to use "composite steps" this way now? (Originally you couldn't which is why I have these ones on separate branches)

PR #390, opened 2022-12-29
feat: create  parameter to RepositoryAssertions::count() methods

nikophil:
Hello!

here is a simple feature I've been missing a lot lately!

### DIFF
--- a/src/RepositoryAssertions.php
+++ b/src/RepositoryAssertions.php
@@ -18,6 +18,8 @@ use Zenstruck\Assert;
  */
 final class RepositoryAssertions
 {
+    private static string $countWithoutCriteriaDeprecationMessagePattern = 'Passing the message to %s() as second parameter is deprecated. Use third parameter.';
+
     public function __construct(private RepositoryProxy $repository)
     {
     }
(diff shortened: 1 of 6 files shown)


src/RepositoryProxy.php
```diff
@@ -61,6 +61,16 @@ public function count(): int
         return \count($this->findAll());
     }
 
+    public function countWithCriteria(array $criteria): int
```
nikophil: at first, I wanted to add this `$criteria` parameter to `count()` but that's not possible because this class implements `\Countable`

src/RepositoryAssertions.php
```diff
@@ -18,6 +18,8 @@
  */
 final class RepositoryAssertions
 {
+    private static string $countWithoutCriteriaDeprecationMessagePattern = 'Passing the message to %s() as second parameter is deprecated. Use third parameter.';
```
nikophil, line 21: in all  PHPUnit assertions, "message" parameter is the last parameter, so I wanted to keep the same logic.

src/RepositoryProxy.php
kbond: I would have thought this possible with a default value for the parameter, no?

src/RepositoryAssertions.php
kbond: 👍🏻

src/RepositoryProxy.php
nikophil: hum, you're right! I'll fix this

nikophil: done

kbond (approved):
This is great!

PR #393, opened 2023-01-09
feat(make:factory): default value for enum types

nikophil:
related to #336

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
             CONFIGURATION="--configuration phpunit-dama-doctrine.xml.dist"
           fi
 
-          vendor/bin/simple-phpunit -v ${CONFIGURATION}
+          vendor/bin/simple-phpunit ${CONFIGURATION}
         env:
           USE_ORM: ${{ matrix.use-orm }}
           USE_ODM: ${{ matrix.use-odm }}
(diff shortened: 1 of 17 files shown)


.github/workflows/ci.yml
```diff
@@ -80,7 +80,7 @@ jobs:
             CONFIGURATION="--configuration phpunit-dama-doctrine.xml.dist"
           fi
 
-          vendor/bin/simple-phpunit -v ${CONFIGURATION}
+          vendor/bin/simple-phpunit ${CONFIGURATION}
```
nikophil, line 83: if you're OK, I'd like to keep it this way, the `-v` flag display all the skipped tests, which makes lot of noise in the CI results

src/ChainManagerRegistry.php
```diff
@@ -31,7 +32,7 @@ public function getRepository($persistentObject, $persistentManagerName = null):
         foreach ($this->managerRegistries as $managerRegistry) {
             try {
                 return $managerRegistry->getRepository($persistentObject, $persistentManagerName);
-            } catch (MappingException) {
+            } catch (MappingException|ORMMappingException) {
```
nikophil, line 35: not catching this exception throws an error in `MakeFactoryTest::can_create_factory_with_default_enum()` with data set `odm` (the latter does not extend the former)

not sure why we've not seen this before :thinking:

kbond (approved):
Nice!

PR #395, opened 2023-01-10
Split Instantiator

### LINKED ISSUE #388: Handling factory methods using `instantiateWith`
Hi, I have big entity with lots of attributes. This Entity has different instantiation logic based on the usage, so I use factory methods which have some validation logic. I would like to use these methods in foundry factories too, so I use 

```php
// ...
->instantiateWith(function(array $attributes, string $class): object {
    return MyEntity::factoryMethod($attributes['foo'], $attributes['bar']);
})
```
I'm wondering are there any other option? Because, in this case the other entity properties never initialized based on the attributes. I don't want to call all the setters on the entity manually. Using the `Instantiator` class is not possible because it will not call the factory method on the entity, but it always instantiates the subject by itself https://github.com/zenstruck/foundry/blob/321abbef052b485ddf3efeddd0c26f9dfcbb3091/src/Instantiator.php#L43

Maybe we should split this class into 2, one for instantiate and one for initialize the subject, or simply just allow to call factory methods instead of the constructor like:

```php
new Instantiator([MyEntity::class, `factoryMethod`])
// of if we want to keep it reusable 
public function __invoke(array $attributes, string $class, string $method = '__construct'): object
// or setting the instance
Instantiator::withObject($entityInstance)
```

kbond:
Closes #388.

This PR splits the Instantator into 2 separate objects:
- `Instantiator`: now only create the object (using constructor/no constructor/callback/named constructor)
- `Hydrator`: takes the instantiated object and hydrates it with the attributes not used to instantiate

The following is now possible:

```php
// customize the default instantiator (defined in your bundle config)
$factory->instantiateWith(Instantiator::noConstructor()); // trigger to create w/o constructor

$factory->instantiateWithoutConstructor(); // shortcut for above

$factory->instantiateWith('factoryMethod'); // creates via Entity::factoryMethod(...)

$factory->instantiateWith([SomeCustomInstantiator::class, 'method']);

$factory->instantiateWith(function(array $attributes) {
   return new ...;
});

// customize the default hydrator (defined in your bundle config)
$factory->hydrateWith(Hydrator::ALLOW_EXTRA_ATTRIBUTES);

$factory->hydrateWith(Hydrator::ALLOW_EXTRA_ATTRIBUTES|Hydrator::ALWAYS_FORCE_PROPERTIES);

$factory->hydrateWith(fn(Hydrator $h) => $h->alwaysForceProperties(['foo', 'bar']));

// fully customize
$factory->hydrateWith(function(object $object, array $attributes) {
    // ...

    return $object;
});
```

Note: if using a closure or non-Instantiator callable for the instantiator, the hydrator isn't used. This is to avoid the unset issue [discussed here](https://github.com/zenstruck/foundry/issues/388#issuecomment-1376020947). I think this is better for DX (not requiring you create a custom instantiator _and_ hydrator) but might cause problems when trying to solve the cascade: persist problem in #312.

Todo:

- [ ] Deprectations layer
- [ ] Tests
- [ ] Docs


kbond:
@nikophil, could you give this an early, high-level review?

src/Configuration.php
```diff
@@ -103,6 +111,17 @@ public function setInstantiator(callable $instantiator): self
         return $this;
     }
 
+    public function setHydrator(callable $hydrator): self
+    {
+        if (!$hydrator instanceof Hydrator) {
+            $hydrator = (new Hydrator())->using($hydrator);
+        }
```
nikophil, line 118: this is actually a nice technique :+1:

src/Factory.php
```diff
@@ -573,4 +616,32 @@ private function isPersisting(): bool
...
     }
+
+    /**
+     * @return TObject
+     */
+    private function instantiateAndHydrate(array $attributes): object
+    {
+        $instantiator = $this->instantiator ?? self::configuration()->instantiator();
+
+        if (!$instantiator instanceof Instantiator && !$this->hydrator) {
```
nikophil, line 627: > Note: if using a closure or non-Instantiator callable for the instantiator, the hydrator isn't used.

shouldn't we throw an error if we detect an hydrator which is not the default hydrator in this case?

kbond: Perhaps but then we'd break the possibility of fixing cascade persist with custom instantiators. One problem at a time though. I'll throw an exception for now.

nikophil: damn, that's quite a complex problem :exploding_head:

nikophil:
I'm closing this PR, even if we want to work on this, it was targetting 1.x, so we'd need to revamp it entirely

PR #398, opened 2023-01-20
feat: drop support of symfony 4, remove deprecation for console/command

Chris53897:
I am not sure if this bundle still wants to support symfony 4 (no longer LTS).

My main goal is to fix the deprecation for `protected static $defaultName`
But that of course should not be the only reason to drop support for symfony 4 right now.

There are other deprecation notices for `doctrine/mongodb-odm` why the tests are failing.


nikophil:
Hello @Chris53897 

we still want to support Symfony 4.4 until it reaches end of life support.

nikophil:
Hi @Chris53897 

I think we still want to support Symfony 4.4 one more year, until it reaches end of life

OskarStark:
IMHO Symfony 4.4 support should be dropped, it is in security only mode, so even bugs will not be fixed in 4.4

nikophil:
OK I just learnt that security only mode means EOL :sweat_smile: 

let's drop 4.4 then.

please @Chris53897 could you rebase your PR on `1.x`, then tests should pass

kbond:
https://github.com/zenstruck/foundry/blob/1.x/tests/Functional/ContainerBC.php can be removed too.

nikophil:
And we can use attributes instead of annotations in fixtures

Chris53897:
I have made the suggested changes from @kbond and @nikophil 
But i do not know how to fix the `--prefer-lowest `Testruns correctly. Without loseing to much older versions.
If i bump min version it too high, the 2 deprecation warnings from `doctrine/mongodb-odm` will rise again.

Any suggestions to solve this?

nikophil:
Hi @Chris53897 

wow! thanks to tackle the migration from annotations to attributes!

I managed to fix the problem, but it was quite hard.
Please modify the `require-dev` section of `composer.json` with the followings:
```diff
-        "dama/doctrine-test-bundle": "^6.0|^7.0",
-        "doctrine/doctrine-bundle": "^2.0",
+        "dama/doctrine-test-bundle": "^7.0",
+        "doctrine/doctrine-bundle": "^2.5",
         "doctrine/doctrine-migrations-bundle": "^2.2|^3.0",
-        "doctrine/mongodb-odm-bundle": "^3.1|^4.2",
+        "doctrine/mongodb-odm-bundle": "^4.4.0",
         "doctrine/mongodb-odm": "^2.0",
         "doctrine/orm": "^2.9",
         "matthiasnoback/symfony-dependency-injection-test": "^4.1",
+        "symfony/doctrine-bridge": "^5.4|^6.0",
```

nikophil:
:tada:

nikophil (approved):

Chris53897:
@nikophil Thanks for the help

kbond:
@Chris53897, can you rebase and fix the conflicts?

Chris53897:
@kbond Can you please check if i have done this correct?

kbond:
Thanks @Chris53897!

Issue #397, opened 2023-01-16
Can't access foundry's faker if not booted (PHPUnit data-provider) + error in symfony foundry doc


Methraen:
There is a problem in symfony foundry's doc...
https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#phpunit-data-providers
This case is not possible :
```
    public static function postDataProvider(): iterable
    {
        yield [PostFactory::new()];
        yield [PostFactory::new()->published()]; // <- this one
    }
```

You cant add a state to a factory when the state uses Faker because foundry's Faker is accessible only if foundry is booted
(Factory.php)

```
    final public static function faker(): Faker\Generator
    {
        return self::configuration()->faker();
    }
[...]
    final public static function configuration(): Configuration
    {
        if (!self::isBooted()) {
            throw new \RuntimeException('Foundry is not yet booted. Using in a test: is your Test case using the Factories trait? Using in a fixture: is ZenstruckFoundryBundle enabled for this environment?');
        }

        return self::$configuration; // @phpstan-ignore-line
    }
```

I have experienced it myself,
my code :
```
class CustomerRepositoryTest extends KernelTestCase
{
    use ResetDatabase, Factories;
    private CustomerRepository $repository;

    /**
     * @throws Exception
     */
    protected function setUp(): void
    {
        self::bootKernel();

        /** @var CustomerRepository $repository */
        $repository       = static::getContainer()->get(CustomerRepository::class);
        $this->repository = $repository;
    }

    /**
     * @param array<string, mixed> $expectedValues
     *
     * @see CustomerRepository::getCustomerQueryBuilder
     *
     * @dataProvider provideGetCustomersQueryBuilderDatasets
     */
    public function testGetCustomersQueryBuilder(FactoryCollection $factory, array $expectedValues, CustomerSearchCommand $customerSearchCommand): void
    {
        $factory->create();

        /** @var int $numberOfEntities */
        $numberOfEntities = $expectedValues['numberOfEntities'];

        /** @var Customer[] $items */
        $items = $this->repository->getCustomersQueryBuilder($customerSearchCommand)->getQuery()->getResult();

        $this->assertCount($numberOfEntities, $items);
        $itemsCount = count($items);
    }

    /**
     * @throws Exception
     */
    public function provideGetCustomersQueryBuilderDatasets(): Generator
    {
        $customerFactory = CustomerFactory::new()->sequence(
            static function (): array {
                $items = [];
[...]

PR #399, opened 2023-01-20
minor: meaningful error for faker in data provider

nikophil:
related to #397 

a more opinionated solution would be to deprecate the usage of `array` parameters in `withAttributes()`
but I don't know if it is worth it

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
         "doctrine/doctrine-bundle": "^2.0",
         "doctrine/doctrine-migrations-bundle": "^2.2|^3.0",
         "doctrine/mongodb-odm-bundle": "^3.1|^4.2",
+        "doctrine/mongodb-odm": "^2.0",
         "doctrine/orm": "^2.7",
         "matthiasnoback/symfony-dependency-injection-test": "^4.1",
         "symfony/framework-bundle": "^4.4|^5.0|^6.0",
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1472,6 +1472,28 @@ It is possible to use factories in
     For the same reason as above, it is not possible to use `Factories as Services`_ with required
     constructor arguments (the container is not yet available).
 
+.. note::
+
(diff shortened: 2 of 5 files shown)


src/Factory.php
```diff
@@ -290,7 +292,13 @@ final public static function isBooted(): bool
 
     final public static function faker(): Faker\Generator
     {
-        return self::configuration()->faker();
+        try {
+            return self::configuration()->faker();
+        } catch (FoundryNotBootedException $exception) {
+            throw new \RuntimeException(
+                "Cannot get Foundry's configuration. If using faker in a data provider, consider passing attributes as a callable."
```
nikophil, line 299: I'm wondering if this error could happen in an other case than in a data provider? :thinking:

composer.json
```diff
@@ -27,6 +27,7 @@
         "doctrine/doctrine-bundle": "^2.0",
         "doctrine/doctrine-migrations-bundle": "^2.2|^3.0",
         "doctrine/mongodb-odm-bundle": "^3.1|^4.2",
+        "doctrine/mongodb-odm": "^2.0",
```
nikophil, line 30: this prevents some weird deprecations triggered by a `dev-*` version of `doctrine/mongo-odm`

kbond (approved):

docs/index.rst
```diff
@@ -1472,6 +1472,28 @@ It is possible to use factories in
...
+    to pass attributes as a *callable*.
+
+    Given the data provider of the previous example, here is ``PostFactory::published()``
+
+    .. code-block:: php
+
+        public function published(): self
+        {
+            // This won't work in a data provider!
+            // return $this->wtiAttributes(['published_at' => self::faker()->dateTime()]);
```
```suggestion
            // return $this->withAttributes(['published_at' => self::faker()->dateTime()]);
```

nikophil: :oups:!

nikophil: > // This won't work in a data provider!

this is really true! :smile:

PR #400, opened 2023-01-20
feat(make:factory): list embeddables

nikophil:
This PR fixes multiple things:
- we now list ORM embeddables in the listing in `make:factory`
- then I discovered some problems with ODM embeddables which I fixed
- which leads me to improve default property guesser for "not persisted" objects


src/Bundle/Maker/Factory/AbstractDefaultPropertyGuesser.php
```diff
@@ -0,0 +1,36 @@
...
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+abstract class AbstractDefaultPropertyGuesser implements DefaultPropertiesGuesser
+{
+    public function __construct(private FactoryClassMap $factoryClassMap, private FactoryGenerator $factoryGenerator)
+    {
+    }
+
+    /** @param class-string $fieldClass */
+    protected function addDefaultValueUsingFactory(SymfonyStyle $io, MakeFactoryData $makeFactoryData, MakeFactoryQuery $makeFactoryQuery, string $fieldName, string $fieldClass): void
```
nikophil, line 18: just moved this method from `AbstractDoctrineDefaultPropertiesGuesser` to here

src/Bundle/Maker/Factory/ObjectDefaultPropertiesGuesser.php
```diff
@@ -62,4 +63,53 @@ public function supports(MakeFactoryData $makeFactoryData): bool
...
+        if ($constructor = $property->getDeclaringClass()->getConstructor()) {
+            foreach ($constructor->getParameters() as $parameter) {
+                if ($parameter->getName() === $property->getName()) {
+                    return $parameter;
+                }
+            }
+        }
+
+        return null;
+    }
```
nikophil, line 114: this method it a little bit biased because we rely on the fact the constructor's parameters have the same name than the property.

I think this could be accepted because it is a pattern widely used

tests/Fixtures/Maker/expected/can_create_factory_for_not_persisted_class.php
```diff
@@ -41,7 +41,7 @@ protected function getDefaults(): array
             'intMandatory' => self::faker()->randomNumber(),
             'propertyWithoutType' => null, // TODO add value manually
             'someMandatoryPropertyWithUnionType' => null, // TODO add value manually
-            'someOtherObjectMandatory' => null, // TODO add Zenstruck\Foundry\Tests\Fixtures\Object\SomeOtherObject value manually
+            'someOtherObjectMandatory' => SomeOtherObjectFactory::new(),
```
nikophil, line 44: at first I thought "there is a bug, the factory class is not imported", but both `SomeOtherObjectFactory` and this one are in the same namespace :sweat_smile:

kbond (approved):
LGTM!

Issue #401, opened 2023-01-20
Factory Maker: match directory/namespace structure


jmsche:
Hi, it would be nice if the `make:factory` command could match the entity namespace/directory structure?

Eg. for an entity with the `App\Entity\Franchise\Shop` FQCN, it would create a factory in the `src/Factory/Franchise/Shop.php` file, in the `App\Factory\Franchise` namespace.

nikophil:
Hi @jmsche 

this actually makes sense, let's do it!

nikophil:
this is actually harder than it seems:

I'm actually wondering from where I should stop to read the original namespace... in your example I should strip `App\Entity` in favor of `App\Factory` (or `App\Test\Factory` if `--test` option is passed)

but what should we do if the entity class is something like `App\Domain\Shop\Model\Franchise\Shop`?

I'm wondering if the answer would not be to access doctrine's configuration. WDYT @kbond?

And if we're asked to create a factory for an object not managed by doctrine, we'd place the factory at the root of the `App\Factory` namespace, as we're doing now

EDIT: after some digging, I think I'll go with this solution

nikophil:
Hi @jmsche 

I've finished the PR

would you play around to check if everything is OK for you?

jmsche:
Seems to work fine for my current use case, thanks :)

PR #402, opened 2023-01-20
feat(make:factory): handle name collision

nikophil:
If two factories will have the same name. ie: we use `All` and multiple entities have the same "short name", but in different namespace, it currently results as an error

Now, we ask for a replacement "short name" (because full namespace are cumbersome to write)


kbond:
#401 is related. As discussed with @nikophil in slack, this PR might not be needed depending on how we solve that issue.

nikophil:
This would still be relevant for non-doctrine related factories, if I use the strategy I suggested in the issue

PR #406, opened 2023-01-22
tests: improve tests for ide

Chris53897:
This helps the IDE (in my case PhpStorm) to read the code.
Please have a look if you find it useful and is correct.

### DIFF
--- a/src/Test/TestState.php
+++ b/src/Test/TestState.php
@@ -78,6 +78,7 @@ final class TestState
 
     /**
      * @deprecated Use TestState::enableDefaultProxyAutoRefresh()
+     * @noinspection PhpDeprecationInspection
      */
     public static function alwaysAutoRefreshProxies(): void
     {
--- a/src/Test/TestState.php
+++ b/src/Test/TestState.php
@@ -78,7 +78,6 @@ final class TestState
 
     /**
      * @deprecated Use TestState::enableDefaultProxyAutoRefresh()
-     * @noinspection PhpDeprecationInspection
      */
     public static function alwaysAutoRefreshProxies(): void
     {
--- a/tests/Functional/ProxyTest.php
+++ b/tests/Functional/ProxyTest.php
@@ -290,7 +290,7 @@ abstract class ProxyTest extends KernelTestCase
         $this->assertSame('another new body', $post->getBody());
     }
 
-    /** @phpstan-return \Zenstruck\Foundry\Tests\Fixtures\Factories\PostFactory|\Zenstruck\Foundry\Tests\Fixtures\Factories\ODM\PostFactory */
+    /** @return class-string<ODMPostFactory|ORMPostFactory> */
     abstract protected function postFactoryClass(): string;
 
     abstract protected function postClass(): string;
--- a/tests/Functional/ProxyTest.php
+++ b/tests/Functional/ProxyTest.php
@@ -290,6 +290,7 @@ abstract class ProxyTest extends KernelTestCase
         $this->assertSame('another new body', $post->getBody());
     }
 
+    /** @phpstan-return \Zenstruck\Foundry\Tests\Fixtures\Factories\PostFactory|\Zenstruck\Foundry\Tests\Fixtures\Factories\ODM\PostFactory */
     abstract protected function postFactoryClass(): string;
 
     abstract protected function postClass(): string;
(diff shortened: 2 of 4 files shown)


tests/Functional/RepositoryProxyTest.php
```diff
@@ -342,7 +344,9 @@ public function can_use_get_count(): void
         $this->assertSame(4, $categoryFactoryClass::repository()->getCount());
     }
 
+    /** @phpstan-return \Zenstruck\Foundry\Tests\Fixtures\Entity\Category|\Zenstruck\Foundry\Tests\Fixtures\Document\ODMCategory */
     abstract protected function categoryClass(): string;
 
+    /** @phpstan-return \Zenstruck\Foundry\Tests\Fixtures\Factories\CategoryFactory|\Zenstruck\Foundry\Tests\Fixtures\Factories\ODM\CategoryFactory */
     abstract protected function categoryFactoryClass(): string;
```
nikophil, line 351: - these methods does not return actual objects but their class, so we can use `class-string` 
- I think there is no need for the `phpstan-` prefix?

```suggestion
    /** @return class-string<ODMCategory|ORMCategory> */
    abstract protected function categoryClass(): string;
    
    /** @return class-string<ODMCategoryFactory|ORMCategoryFactory> */
    abstract protected function categoryFactoryClass(): string;
```

nikophil: - these methods does not return actual objects but their class, so we can use `class-string` 
- I think there is no need for the `phpstan-` prefix?

```suggestion
    /** @return class-string<ODMCategory|ORMCategory> */
    abstract protected function categoryClass(): string;
    
    /** @return class-string<ODMCategoryFactory|ORMCategoryFactory> */
    abstract protected function categoryFactoryClass(): string;
```

tests/Functional/RepositoryProxyTest.php
```diff
@@ -93,6 +93,7 @@ public function assertions(): void
     /**
      * @test
      * @group legacy
+     * @noinspection PhpDeprecationInspection
```
nikophil: not sure we should add PHPStorm specific annotations? :thinking: 

I'm not even sure we should hide that these methods use deprecated methods...

tests/Functional/ProxyTest.php
```diff
@@ -290,6 +290,7 @@ public function without_auto_refresh_keeps_disabled_if_originally_disabled(): vo
         $this->assertSame('another new body', $post->getBody());
     }
 
+    /** @phpstan-return \Zenstruck\Foundry\Tests\Fixtures\Factories\PostFactory|\Zenstruck\Foundry\Tests\Fixtures\Factories\ODM\PostFactory */
     abstract protected function postFactoryClass(): string;
```
nikophil:
```suggestion
    /** @return class-string<ODMPostFactory|ORMPostFactory> */
    abstract protected function postFactoryClass(): string;
```

src/Test/TestState.php
```diff
@@ -78,6 +78,7 @@ public static function disableDefaultProxyAutoRefresh(): void
 
     /**
      * @deprecated Use TestState::enableDefaultProxyAutoRefresh()
+     * @noinspection PhpDeprecationInspection
```
nikophil: I don't really understand why we should hide that we're using a deprecated method?

nikophil (changes requested):
hi @Chris53897 thanks for your PR

IMO IDE-specific annotations should not land in the code.

src/Test/TestState.php
Chris53897: I have reverted it

tests/Functional/RepositoryProxyTest.php
Chris53897: I have reverted it

nikophil (approved):

Chris53897:
I changed it and add the use statements.

PhpStorm does not yet understand it like this https://youtrack.jetbrains.com/issue/WI-55925.
I guess it can only understand it like this.
https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-release/#class-string

But maybe other ide's can understand it now. And hopefully phpstorm in the future

nikophil:
> PhpStorm does not yet understand it

I've tried your PR, and trying to "ctrl + click" `createMany()` in this code, and it works pretty well 
```php
// \Zenstruck\Foundry\Tests\Functional\RepositoryProxyTest

    public function can_find_random_object(): void
    {
        $this->categoryFactoryClass()::createMany(5);
```

nikophil:
if PHPStorm does not understand it, we could change like this, maybe?

```php

    /** @return class-string */
    abstract protected function categoryClass(): string;

    /** @return class-string<ModelFactory> */
    abstract protected function categoryFactoryClass(): string;
```

PR #411, opened 2023-01-24
feat(make:factory): match directory/namespace structure

### LINKED ISSUE #401: Factory Maker: match directory/namespace structure

nikophil:
fix #401 

I've also added a configuration to chose the default namespace that will be used

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -204,6 +204,23 @@ This command will generate a ``PostFactory`` class that looks like this:
 
     Using ``make:factory --test`` will generate the factory in ``tests/Factory``.
 
+.. tip::
+
+    You can globally configure which namespace the factories will be generated in:
+
+    .. configuration-block::
+
+        .. code-block:: yaml
+
+            # config/packages/zenstruck_foundry.yaml
+            when@dev: # see Bundle Configuration section about sharing this in the test environment
+                zenstruck_foundry:
+                    make_factory:
+                        default_namespace: 'App\\MyFactories'
+
+    You can override this configuration by using the ``--namespace`` option.
+
+
 .. note::
 
     The generated ``@method`` docblocks above enable autocompletion with PhpStorm but
@@ -1956,6 +1973,12 @@ Full Default Bundle Configuration
             # Add global state.
             global_state: []
 
+            # Configure Foundry's make:factory command
+            make_factory:
+
+                # Namespace to use for make:factory. Is overridden by --namespace option
+                default_namespace: 'Factory'
+
     .. code-block:: php
 
         $config->extension('zenstruck_foundry', [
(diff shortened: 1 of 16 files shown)


docs/index.rst
```diff
@@ -204,6 +204,24 @@ This command will generate a ``PostFactory`` class that looks like this:
...
+    .. configuration-block::
+
+        .. code-block:: yaml
+
+            # config/packages/zenstruck_foundry.yaml
+            when@dev: # see Bundle Configuration section about sharing this in the test environment
+                zenstruck_foundry:
+                    maker:
+                        factory:
+                            default_namespace: 'App\\MyFactories'
```
nikophil: I'm wondering if this is not too "deep" just for one conf, but it seemed semantically more correct

docs/index.rst
```diff
@@ -2015,8 +2042,19 @@ Full Default Bundle Configuration
                     'object_managers' => false,
                 ],
 
-                // Add global state
-                'global_state' => []
```
nikophil, line 2019: there was an error in the doc: `global_state` should be a direct child of `zenstruck_foundry`

docs/index.rst
```diff
@@ -204,6 +204,24 @@ This command will generate a ``PostFactory`` class that looks like this:
 
     Using ``make:factory --test`` will generate the factory in ``tests/Factory``.
 
+.. tip::
+
+    You can configure globally it which namespace the factories will be generated:
```
```suggestion
    You can globally configure which namespace the factories will be generated in:
```

docs/index.rst
```diff
@@ -204,6 +204,24 @@ This command will generate a ``PostFactory`` class that looks like this:
...
+        .. code-block:: yaml
+
+            # config/packages/zenstruck_foundry.yaml
+            when@dev: # see Bundle Configuration section about sharing this in the test environment
+                zenstruck_foundry:
+                    maker:
+                        factory:
+                            default_namespace: 'App\\MyFactories'
+
+    You can override this configuration by using the option ``--namespace``.
```
```suggestion
    You can override this configuration by using the ``--namespace`` option.
```

docs/index.rst
kbond: I don't have a strong opinion but what about:

```yaml
zenstruck_foundry:
    make_factory:
        default_namespace: ...
```

kbond (changes requested):
Looks good, just a few minor things.

kbond (approved):
Beautiful!

PR #417, opened 2023-02-02
Drop testing on unsupported symfony versions

### LINKED ISSUE #405: Improve CI for Forks (Sync meta files)
It would be great to improve the CI, to get green testresults in forks.
`Sync meta files` fails in lack of token

Example [Sync meta files](https://github.com/Chris53897/foundry/actions/runs/3978621994/jobs/6820511009#logs)

I am not sure if it is possible and how much work that would be.

kbond:
Also fixes #405... I think, we'll need to test once this is merged.


nikophil (approved):
LGTM

Issue #414, opened 2023-01-25
PHPStan issues


benblub:
I have two different problems with phpstan and foundry.

**1. load state from a story and using it on typed setter.**

Without the docblock where I specify the var creator is from type User i get stan error `Parameter #1 $creator of method App\Tests\Factory\SessionFactory::creator() expects App\Entity\User, object given.`. Using the docblock is just a workaround and error prone. If we use ->object() from a factory proxy object than its correct unwrapped and we have the correct return type in stan. In both cases Proxy.php Line 114 public function object(): object is called. No clue what's the difference. Another solution would be to replace type User with object in the factory `public function creator(object $creator): self`. But that's also a workaround and less type safe as i wanted^^ 

```php
        /** @var User $creator */
        $creator = WondrousUsersStory::load()::get('rolf')->object();

        $this->addState('restaurant', SessionFactory::new()
            ->creator($creator)
            ->create()
        );
```

In the factory
```php
    public function creator(User $creator): self
    {
        return $this->addState(['creator' => $creator]);
    }
```

**2. The second prob occurs with using pools in combination with createMany**

Stan error: `Parameter #2 $objects of method Zenstruck\Foundry\Story::addToPool() expects object, array<App\Entity\Alternative|Zenstruck\Foundry\Proxy> given.`

Currently I using the stan ignore error as workaround 🤦 
```php
        /** @phpstan-ignore-next-line */
        $this->addToPool('alternatives', AlternativeFactory::createMany(5,
            function () use ($session, &$titles) {
                return [
                    'session' => $session,
                    'title' => array_shift($titles),
                ];
            }
        ));
```

Any ideas how to solve this? I'am also not sure the problem is myCode, foundry or stan related.

nikophil:
Hi @benblub 

**For your first question:** the problem is that stories are not tied to only one type and thus cannot be generic, so `SomeStory::get('something')` returns just a `Proxy` in the type system.

On the other hand, factories are tied to a specific type, and can then be generic: PHPStan knows that `UserFactory::createOne()` returns a `Proxy<User>` thanks to its `@extends ModelFactory<User>` annotation.

The only solution here is to use `WondrousUsersStory::rolf()` (this methods automatically exists thanks [to this](https://github.com/zenstruck/foundry/blob/1.x/src/Story.php#L41)) and to document the method in your story:
```php
/**
 * @method rolf(): Proxy<User>
 */
class WondrousUsersStory extends Story
```
[...]

PR #420, opened 2023-02-13
fix return type syntax

benblub:
prevents `PHPDoc tag @method has invalid value (rolf(): User): Unexpected token "\n ", expected type at offset  27` (from the closed issue https://github.com/zenstruck/foundry/issues/414)

the errors occurs from the colon `:`
https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/method.html

Maybe also hint in the docs that you need to install `phpstan/phpdoc-parser`

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1843,7 +1843,7 @@ Later, you can access the story's state when creating other fixtures:
         use Zenstruck\Foundry\Story;
 
         /**
-         * @method php(): Category
+         * @method php() Category
          */
         final class CategoryStory extends Story
         {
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -1843,7 +1843,7 @@ Later, you can access the story's state when creating other fixtures:
         use Zenstruck\Foundry\Story;
 
         /**
-         * @method php(): Category
+         * @method php() Category
```
OskarStark:
```suggestion
         * @method Category php()
```

?

benblub: [It looks](https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/method.html) like the method name are coming after the return type but its not working. The proposed change is tested successful @OskarStark

benblub: ok tested again, you are right 👍 Dunno why both cases works -> phpstan doesn't complain but like written in the docs return before method name..

kbond (approved):

PR #423, opened 2023-02-23
refactor: simplify atrtibutes normalization

nikophil:


src/Factory.php
```diff
@@ -350,33 +352,40 @@ private function normalizeAttribute(mixed $value, ?string $name = null)
...
+
+        if ($relationshipMetadata['isOwningSide']) {
+            $cascadePersist = $relationshipMetadata['cascade'];
+        } else {
+            $isCollection = $relationshipMetadata['inverseIsCollection'];
+            $relationshipField = $relationshipMetadata['inversedField'];
+            $cascadePersist = $relationshipMetadata['cascade'];
+
+            if ($this->isPersisting() && null !== $relationshipField && false === $cascadePersist) {
+                $this->afterPersist[] = static function(Proxy $proxy) use ($value, $relationshipField, $isCollection): void {
```
nikophil: I'm wondering how we could remove this "mutable" stuff (same thing in `normalizeCollection()`) :thinking: 

here is an idea: instead of returning `return self::NULL_VALUE;`, what about returning directly the callback OR an invokable object? they could be handled differently afterwards, and be called in the "post persist" phase.

Not sure how we could handle this, more over if we extract the "attributes normalization" phase afterwards, but I'll give it a try!

src/Factory.php
```diff
@@ -418,144 +434,6 @@ function(self $factory): self {
         );
     }
 
-    private function ownedRelationshipField(?string $relationName, self $factory): ?string
```
nikophil, line 422: all of these methods were just actually digging into doctrine's metadata... I've replaced them with `getRelationshipMetadata()`

src/Factory.php
```diff
@@ -389,20 +398,27 @@ private static function normalizeObject(object $object): object
         }
     }
 
-    private function normalizeCollection(?string $relationName, FactoryCollection $collection): array
+    private function normalizeCollection(string $relationName, FactoryCollection $collection): array
```
nikophil: something I've noticed here: if we replace this whole method with `return $collection->all();` the tests are not failing :shrug: 

so we'll need to understand what's really happening here, why (if?) it is needed, and how we could create a failing test case for this.

src/Factory.php
```diff
@@ -350,33 +352,40 @@ private function normalizeAttribute(mixed $value, ?string $name = null)
...
-                    };
+        if (!$objectManager instanceof EntityManagerInterface || $objectManager->getClassMetadata($value->class)->isEmbeddedClass) {
+            return $value->create()->object();
+        }
 
-                    // creation delegated to afterPersist event - return null here
-                    return self::NULL_VALUE;
-                }
-            }
+        $relationshipMetadata = self::getRelationshipMetadata($objectManager, $this->class, $name);
```
nikophil, line 380: "fun" fact: I've actually refactored all this without REALLY knowing **WHY** we are doing this

(I don't know if it is really a "fun" fact :sweat_smile: )

src/Factory.php
```diff
@@ -25,8 +28,6 @@
  */
 class Factory
 {
-    private const NULL_VALUE = '__null_value';
```
nikophil, line 28: I'm glad I manage to get rid of this thing

src/Factory.php
```diff
@@ -103,10 +104,22 @@ final public function create(array|callable $attributes = []): Proxy
...
-                $normalizedAttribute = null;
+
+            if (is_array($normalizedAttribute)) {
+                $postPersistCallbacks = [
+                    ...$postPersistCallbacks,
+                    ...array_values(array_filter($normalizedAttribute, static fn(mixed $attribute): bool => $attribute instanceof PostPersistCallback))
+                ];
+
+                $normalizedAttribute = array_values(array_filter($normalizedAttribute, static fn(mixed $attribute): bool => !$attribute instanceof PostPersistCallback));
+            }
```
nikophil, line 121: ok, there is a trick here, sometimes `normalizeAttribute()` will return arrays with some `PostPersistCallback` AND with some proxyfied objects.
So, we need to extract the callbacks on one side, and the regular attributes on the other side.

Maybe to remove this awful thing, we could make `normalizeAttribute()` return some `AttributesCollection`?

src/Factory.php
```diff
@@ -103,10 +104,22 @@ final public function create(array|callable $attributes = []): Proxy
...
+
+            if ($normalizedAttribute instanceof PostPersistCallback) {
+                $postPersistCallbacks[] = $normalizedAttribute;
+                $normalizedAttribute = $value instanceof FactoryCollection ? [] : null;
```
nikophil, line 125: this replaces the `NULL_VALUE` stuff

src/Factory.php
kbond: So the return type is: `array<string,PostPersistCallback|mixed>`?

nikophil: return type of what? `normalizeAttribute()`?

nikophil: it's `mixed|PostPersistCallback|list<mixed|PostPersistCallback>` I think

nikophil: but... then it's `mixed`  :smile:

nikophil: we return a `list<>` when we have a `X-To-Many`

kbond: Ok, yeah, that's more complex than I thought...

nikophil: actually here is the full list of what we can return:
- any scalar and any "normal" property
- objects created by a "sub" factory (`X-To-One`  with cascade:persist)
- a `PostPersistCallback` in the case of a `X-To-One` without cascade:persist
- a list of objects (`X-To-Many` with cascade:persist)
- a list of `PostPersistCallback` in the case of a `X-To-Many` without cascade:persist

and this test creates a list of `PostPersistCallback` AND objects created by the factory:
```php
        $category = create(Category::class, [
            'name' => 'bar',
            'posts' => [
                factory(Post::class, ['title' => 'Post A', 'body' => 'body']),
                create(Post::class, ['title' => 'Post B', 'body' => 'body']),
            ],
        ]);
```

kbond: Damn... so you're thinking an `AttributesCollection` could encapsulate this complexity?

nikophil: The only thing an AttributCollection will do is to hide the filtering process here I guess
But what we'd need is an abstraction and a `PersistenceAttributeCollection`
Don't know if we want such a complexity 🤔

kbond: Perhaps let's leave out for the context of this PR?

nikophil: For sure, we'll see in the next step. 

But if we create a `PersistenceAtrributesNormalizer` with a method `getPostPersistCallbacks()` (not really sure we willbe able to do this, but that would be useful) we won't need any `AttributesCollection`at all

kbond (approved):

PR #422, opened 2023-02-19
minor: remove `Factory::$cascadePersist`

nikophil:
This reduces a little bit `Factory` class state

PR #424, opened 2023-02-24
fix: restore Factory::$cascadePersist

nikophil:
sorry for that, I just discovered a case where `Factory::$cascadePersist` is needed. The solution in #422 could not work because of this: https://github.com/zenstruck/foundry/blob/1.x/src/Factory.php#L350

I've provided in the PR a test which ensures this case is covered


tests/Functional/FactoryDoctrineCascadeTest.php
```diff
@@ -119,17 +125,17 @@ public function many_to_many_reverse_relationship(): void
      */
     public function one_to_one_relationship(): void
     {
-        $variant = factory(Variant::class, [
+        $product = factory(Product::class, [
```
nikophil, line 128: had to change this test, because it was not anymore relying on a `OneToOne` + cascade persist

kbond (approved):

Issue #428, opened 2023-02-28
Problem with `Factory::findOrCreate()` and `Embeddable`


nikophil:
There is actually no way to use embeddables with `Factory::findOrCreate()`:
```php
#[ORM\Embeddable]
final class Address
{
    #[ORM\Column()]
    public string $value;
}

#[ORM\Entity]
final class Contact
{
    #[ORM\Embedded(Address::class)]
    public Address $address;
}

// 💥 Error "Unrecognized field: address" when doctrine tries to get the contact:
// for doctrine, this field is called "address.value" 
ContactFactory::findOrCreate(['address' => 'some address']); 

// 💥 "Class "Contact" does not have property "address.value"" when Foundry tries to create the address
ContactFactory::findOrCreate(['address.value' => 'some address']); 
```
Actually, I don't think we will be able to fix this :thinking:

KDederichs:
Ah, bummer cause I came here cause I ran into EXACTLY the same issue :)

One solution could be that you require the Address to be initialized (via constructor) and then if you find an Embedded in create/find it'll call the getter/setter on that maybe?

Like `address.value` is mapped to `address->getValue()` or `address->setValue()` internally?

nikophil:
> Ah, bummer cause I came here cause I ran into EXACTLY the same issue :)

this may give some more motivation to fix this!

I'm wondering if we couldn't use it like this:
```php
ContactFactory::findOrCreate(['address' => new Address('some address')]); 
```
this way Foundry is happy when it tries to create the address, and we could transform all embeddable objects into DQL criteria understandable by doctrine. 

Actually that won't be that complex to fix.

Any thoughts @kbond?

kbond:
Yes that makes sense, since it's an embeddable, I think we could also accept:

```php
ContactFactory::findOrCreate(['address' => AddressFactory::new('postalCode' => '90210')]);
```

PR #432, opened 2023-03-15
feat: findOrCreate() can use embeddable

nikophil:
related to #428

This PR has some limitations:
- still not possible to use factories as suggested in https://github.com/zenstruck/foundry/issues/428#issuecomment-1466957360 but I'm wondering if wan can do this, because we need to extract the attributes from the factory. Thus `Factory::normalizeAttributes()` should be opened, and I'm not sure it is a good idea (maybe we can make it internal?)
- not possible to deal with nested embeddable: this is not impossible, but would need some recursive method that would complicate the code

not really sure where we should document this?

/cc @KDederichs

### DIFF
--- a/src/RepositoryProxy.php
+++ b/src/RepositoryProxy.php
@@ -12,9 +12,14 @@
 namespace Zenstruck\Foundry;
 
 use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata as ODMClassMetadata;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\EntityRepository;
+use Doctrine\ORM\Mapping\ClassMetadata as ORMClassMetadata;
+use Doctrine\Persistence\Mapping\MappingException;
+use Doctrine\Persistence\ObjectManager;
 use Doctrine\Persistence\ObjectRepository;
+use Symfony\Component\PropertyAccess\PropertyAccess;
 
 /**
  * @mixin EntityRepository<TProxiedObject>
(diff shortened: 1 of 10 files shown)


src/RepositoryProxy.php
```diff
@@ -311,7 +315,42 @@ public function find($criteria)
...
+            };
+
+            if (!$isEmbedded) {
+                throw new \InvalidArgumentException('Only embeddable objects can be passed as attributes for "findOrCreate()" method.');
+            }
+
+            foreach ($metadataForAttribute->getFieldNames() as $field) {
+                $embeddableFieldValue = $propertyAccessor->getValue($attributeValue, $field);
+                if (is_object($embeddableFieldValue)) {
+                    throw new \InvalidArgumentException('Nested embeddable objects are still not supported in "findOrCreate()" method.');
```
kbond, line 350: Does doctrine allow nested embeddables?

kbond (commented):
This adds quite a bit of complexity 😬

For accepting a factory, what about just creating the factory and getting the created object?

src/RepositoryProxy.php
nikophil: yes, it does

kbond: TIL

nikophil:
> This adds quite a bit of complexity

I know :disappointed: the best would have been to put it in its own class maybe?

> For accepting a factory, what about just creating the factory and getting the created object?

yes good idea! I'm just wondering if we should do something if a user wrongly adds a "persisting" factory here (he/she would have used this totally wrong!)?

kbond:
> I know 😞 the best would have been to put it in its own class maybe?

Let's leave it for now but keep this in mind when doing our refactor.

> something if a user wrongly adds a "persisting" factory here

Let's throw a withoutPersisting call to be safe but they'd hit an exception because it wouldn't be embeddable.

nikophil:
> Let's throw a withoutPersisting call

ok cool, that was what I had in mind

nikophil:
ok done :heavy_check_mark: 

FYI the failures in the CI come from a bug in doctrine/mongo-odm which has been fixed [here](https://github.com/doctrine/mongodb-odm/pull/2511) but still not released

kbond (approved):

PR #435, opened 2023-03-17
fix: running `array_values()` on array attribute

kbond:
#423 lost the keys for attributes that were arrays.


src/Factory.php
```diff
@@ -116,7 +116,7 @@ final public function create(array|callable $attributes = []): Proxy
                     ...\array_values(\array_filter($normalizedAttribute, static fn(mixed $attribute): bool => $attribute instanceof PostPersistCallback)),
```
kbond, line 116: Pretty sure we want to keep the `array_values` here.

nikophil: indeed!

nikophil (approved):

PR #436, opened 2023-03-19
refactor: deprecate AnonymousFactory class and factory() helper

nikophil:


src/functions.php
```diff
@@ -20,13 +20,31 @@
...
+/**
+ * @see Factory::__construct()
+ *
+ * @template TObject of object
+ *
+ * @param class-string<TObject> $class
+ *
+ * @return Factory<TObject>
+ */
+function anon(string $class, array|callable $defaultAttributes = []): Factory
```
nikophil: not 100% sure about this method name :shrug:

src/functions.php
```diff
@@ -20,13 +20,31 @@
...
+{
+    return new class($class, $defaultAttributes) extends Factory{};
```
nikophil, line 45: finally could use anonymous class, because we only use `Factory` and not `ModelFactory`

src/functions.php
kbond: Maybe `anonymous` to be explicit?

docs/index.rst
```diff
@@ -937,8 +937,7 @@ Foundry can be used to create factories for entities that you don't have model f
     use function Zenstruck\Foundry\create;
     use function Zenstruck\Foundry\create_many;
 
-    $factory = AnonymousFactory::new(Post::class);
-    $factory = factory(Post::class); // alternative to above
+    $factory = anon(Post::class);
```
kbond: Above:

```diff
-use Zenstruck\Foundry\AnonymousFactory;
-use function Zenstruck\Foundry\factory;
+use function Zenstruck\Foundry\anon;
+use function Zenstruck\Foundry\repository;
```

docs/index.rst
```diff
@@ -948,46 +947,43 @@ Foundry can be used to create factories for entities that you don't have model f
...
+    $repository->randomRange(0, 5); // array containing 0-5 "Post|Proxy" objects
+    $repository->randomRange(0, 5, ['author' => 'kevin']); // filter by the passed attributes
 
     // convenience functions
     $entity = create(Post::class, ['field' => 'value']);
     $entities = create_many(Post::class, 5, ['field' => 'value']);
 
+.. note::
+
+    If you need to do more with an anonymous factory, you need to create an explicit model factory class.
```
```suggestion
    If your anonymous factory code is getting too complex, this could be a sign you need an explicit model factory class.
```

src/AnonymousFactory.php
```diff
@@ -16,9 +16,18 @@
...
  *
  * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @deprecated
  */
 final class AnonymousFactory extends Factory implements \Countable, \IteratorAggregate
 {
+    public function __construct(string $class, array|callable $defaultAttributes = [])
+    {
+        trigger_deprecation('zenstruck\foundry', '1.30', 'Class "AnonymousFactory" is deprecated and will be removed in 2.0. Use functions "create()", "instantiate()" or "repository()" instead.');
```
kbond:
```suggestion
        trigger_deprecation('zenstruck\foundry', '1.30', 'Class "AnonymousFactory" is deprecated and will be removed in 2.0. Use the "anon()" or "repository()" functions instead.');
```

src/functions.php
```diff
@@ -20,13 +20,31 @@
  *
  * @param class-string<TObject> $class
  *
+ * @deprecated
+ *
  * @return AnonymousFactory<TObject>
  */
 function factory(string $class, array|callable $defaultAttributes = []): AnonymousFactory
 {
+    trigger_deprecation('zenstruck\foundry', '1.30', 'Usage of "factory()" function is deprecated and will be removed in 2.0. Use functions "create()", "instantiate()" or "repository()" instead.');
```
kbond:
```suggestion
    trigger_deprecation('zenstruck\foundry', '1.30', 'Usage of "factory()" function is deprecated and will be removed in 2.0. Use the "anon()" or "repository()" functions instead.');
```

docs/index.rst
nikophil: nice catch!

kbond (approved):

PR #438, opened 2023-03-21
fix: regression from embedded object in findOrCreate function

### LINKED ISSUE #437: findOrCreate doesn't allow entity in criteria, only embeddable object
The following PR https://github.com/zenstruck/foundry/pull/432 allow embeddable objects in criteria, but introduce a regression when using entities that aren't embeddable.

This exception is raised `InvalidArgumentException: Only embeddable objects can be passed as attributes for "findOrCreate()" method.` but the previous behavior allow to use entity in criteria.

For example:

```php
#[ORM\Entity]
class User 
{
   #[ORM\OneToOne(targetEntity: Device::class)]
    private Device $device;
}


// Foundy code that works before
$device = DeviceFactory::createOne();
UserFactory::findOrCreate(['device' => $device]); // now InvalidArgumentException is raised

// workaround that work with the merged PR
$device = DeviceFactory::createOne();
$user = UserFactory::findBy(['device' => $device]);

if(\count($user) === 0){
  UserFactor::createOne(['device' => $device]);
}
```

nicolasne:
Fix the issue https://github.com/zenstruck/foundry/issues/437.

I added:

- in `src/RepositoryProxy.php` when `isEmbedded` is false, that means this is a regular entity
- in `src/RepositoryProxy.php` I added a test to allow `$attributeValue` to be a `Proxy` object (maybe it's an hidden feat, WDYT ?)
- add minimal tests to prevent futur regressions

### DIFF
--- a/src/RepositoryProxy.php
+++ b/src/RepositoryProxy.php
@@ -326,6 +326,8 @@ final class RepositoryProxy implements ObjectRepository, \IteratorAggregate, \Co
 
             if ($attributeValue instanceof Factory) {
                 $attributeValue = $attributeValue->withoutPersisting()->create()->object();
+            } else if ($attributeValue instanceof Proxy) {
+                $attributeValue = $attributeValue->object();
             }
 
             try {
@@ -340,8 +342,11 @@ final class RepositoryProxy implements ObjectRepository, \IteratorAggregate, \Co
                 default => throw new \LogicException(\sprintf('Metadata class %s is not supported.', $metadataForAttribute::class))
             };
 
+            // it's a regular entity
             if (!$isEmbedded) {
-                throw new \InvalidArgumentException('Only embeddable objects can be passed as attributes for "findOrCreate()" method.');
+                $normalizedCriteria[$attributeName] = $attributeValue;
+
+                continue;
             }
 
             foreach ($metadataForAttribute->getFieldNames() as $field) {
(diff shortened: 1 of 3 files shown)


src/RepositoryProxy.php
```diff
@@ -326,6 +326,8 @@ public function find($criteria)
 
             if ($attributeValue instanceof Factory) {
                 $attributeValue = $attributeValue->withoutPersisting()->create()->object();
+            } else if ($attributeValue instanceof Proxy) {
+                $attributeValue = $attributeValue->object();
```
nikophil, line 330: nice catch :+1:

nikophil (changes requested):
thanks to tackle this!

please, could you update the exception message at line 336 in `RepositoryProxy`? (I cannot comment this line)

the rest of the PR is ok for me

nicolasne:
> thanks to tackle this!
> 
> please, could you update the exception message at line 336 in `RepositoryProxy`? (I cannot comment this line)
> 
> the rest of the PR is ok for me

Phpstorm indicates that exception isn't raised in the try block. Also what do you think to remove the try-catch block as the exception will be trown only if the class isn't managed by Doctrine?  Or do you have a suggestion for the message? 🤔

kbond:
I pushed a quick fix for the exception message.

`ObjectManager::getClassMetadata()` doesn't indicate that a `MappingException` is thrown but it is by implementations.

nikophil (approved):

kbond:
Thanks @nicolasne!

PR #440, opened 2023-03-22
fix: allow any object passed as find() criteria

### LINKED ISSUE #439: findOrCreate doesn't allow `\DateTime`or `\DateTimeImmutable`
The following PR https://github.com/zenstruck/foundry/pull/432 allow embeddable objects in criteria, but introduce a regression when using `\DateTime` or `\DateTimeImmutable` property.

This exception is raised `InvalidArgumentException: Only managed objects can be passed as attributes for "findOrCreate()" method.` and `Doctrine\Persistence\Mapping\MappingException: The class 'DateTimeImmutable' was not found in the chain configured namespaces App\Entity`, but the previous behavior allow to use type object in criteria.

```php
#[ORM\Entity]
class User 
{
   #[ORM\Column(type: 'datetime_immutable')]
    private \DateTimeImmutable $createdAt;
}


// Foundy code that works before
UserFactory::findOrCreate([
    'createdAt' => \DateTimeImmutable::createFromFormat('Y-m-d H:i:s', '2012-10-17 00:00:00'),
]); // now InvalidArgumentException is raised

// workaround that work with the merged PR
$createdAt =  \DateTimeImmutable::createFromFormat('Y-m-d H:i:s', '2012-10-17 00:00:00');
$user = UserFactory::findBy(['createdAt' => $createdAt]);

if(\count($user) === 0){
  UserFactor::createOne(['createdAt' => $createdAt]);
}
```

I didn't check but i think it will be the same for custom DBAL type having object value.

It's related with https://github.com/zenstruck/foundry/issues/437

nikophil:
fixes #439

/cc @sad270

### DIFF
--- a/src/RepositoryProxy.php
+++ b/src/RepositoryProxy.php
@@ -332,8 +332,10 @@ final class RepositoryProxy implements ObjectRepository, \IteratorAggregate, \Co
 
             try {
                 $metadataForAttribute = $this->getObjectManager()->getClassMetadata($attributeValue::class);
-            } catch (MappingException $e) {
-                throw new \InvalidArgumentException('Only managed objects can be passed as attributes for "findOrCreate()" method.', previous: $e);
+            } catch (MappingException) {
+                $normalizedCriteria[$attributeName] = $attributeValue;
+
+                continue;
             }
 
             $isEmbedded = match ($metadataForAttribute::class) {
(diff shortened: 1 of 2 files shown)


sad270:
Works for me ! Thanks for the fix and for the reactivity 😄

nikophil:
> Thanks for the fix and for the reactivity

sorry for the annoyance :wink: 

The fix is released in 1.30.2

PR #441, opened 2023-03-22
fix(RepositoryProxy::find()): allow not entity object in

nikophil:
want a good joke? `Doctrine\ORM\Mapping\MappingException` does not extend `Doctrine\Persistence\Mapping\MappingException`! (but `\Doctrine\ODM\MongoDB\Mapping\MappingException` does) :facepalm: :clown_face:


OskarStark:
Argh, this should be reported to doctrine

kbond (approved):

nikophil:
@OskarStark https://github.com/doctrine/orm/issues/10591

kbond:
Looks like this won't be fixed until orm 3.0

PR #442, opened 2023-03-24
Have `delayFlush()` return the callback's return

HypeMC:
A minor DX improvement, instead of having to do eg:

```php
$tags = [];
Factory::delayFlush(function() use (&$tags) {
    // ...
    $tags = TagFactory::createMany(200);
});
```

the needed fixtures can now be returned instead:

```php
$tags = Factory::delayFlush(function() {
    // ...
    return TagFactory::createMany(200);
});
```

or multiple:

```php
[$categories, $tags] = Factory::delayFlush(function() {
    return [
        CategoryFactory::createMany(100),
        TagFactory::createMany(200),
    ];
});
```

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -155,17 +155,19 @@ final class Configuration
         return $this->databaseResetEnabled;
     }
 
-    public function delayFlush(callable $callback): void
+    public function delayFlush(callable $callback): mixed
     {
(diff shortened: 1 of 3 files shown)


kbond (approved):

kbond:
Nice, thanks @HypeMC!

PR #444, opened 2023-03-27
fix: nested factories not persisting should not throw error

### LINKED ISSUE #443: Class was not found in the chain confnigured namespaces since 1.30.0
Loving Foundry, we're rolling it out across our very heavily testing Symfony 5.4 codebase with plans to completely replace Alice fixtures. 

I ran into an issue when upgrading from 1.29.0 to 1.30.3.

PR #423 appears to have broken Foundry when creating objects which are not Doctrine entities. We're using Foundry to create both Entities and DTO's in the same test and now we're receiving the following exception:

```
Doctrine\Persistence\Mapping\MappingException : The class 'App\Integration\FooApiResponse' was not found in the chain configured namespaces ...
vendor/doctrine/persistence/src/Persistence/Mapping/MappingException.php:26
vendor/doctrine/persistence/src/Persistence/Mapping/Driver/MappingDriverChain.php:89
vendor/doctrine/doctrine-bundle/Mapping/MappingDriver.php:42
vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:132
vendor/doctrine/doctrine-bundle/Mapping/ClassMetadataFactory.php:18
vendor/doctrine/persistence/src/Persistence/Mapping/AbstractClassMetadataFactory.php:343
vendor/doctrine/persistence/src/Persistence/Mapping/AbstractClassMetadataFactory.php:207
vendor/doctrine/orm/lib/Doctrine/ORM/EntityManager.php:318
vendor/zenstruck/foundry/src/Factory.php:375
vendor/zenstruck/foundry/src/Factory.php:111
src/Factories/VendorFactory.php:101
tests/Functional/Api/VendorUpdaterTest.php:60
 ```

Stepping in with Xdebug I can identify the simplification in `Factory::normalizeAttribute()` which fetches the object manager for class. It appears that Doctrine is returning the default object manager when the given class is not a configured entity, and then calling `->getClassMetadata()` for a non-entity throws an exception. See: https://github.com/zenstruck/foundry/pull/423/files#diff-f1158fbec0bb446d87dbd5346fba683c58ccec1a98b1787511ec7786585e8c31R376

Whereas previously, `$this->ownedRelationshipField($name, $value);` returned `null` on line 360 and `$this->isPersisting()` returned `false` on line 369.

I think the `$objectManager instanceof EntityManagerInterface` check on line 376 was intended to catch this. I'm unsure if there's a better way to ask Doctrine if a gives class is a configured entity.

nikophil:
fixes #443 

/cc @ndench

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -14,6 +14,7 @@ namespace Zenstruck\Foundry;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata as ODMClassMetadata;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata as ORMClassMetadata;
+use Doctrine\Persistence\Mapping\MappingException as ORMMappingException;
 use Faker;
 use Zenstruck\Foundry\Exception\FoundryNotBootedException;
 use Zenstruck\Foundry\Persistence\InversedRelationshipCascadePersistCallback;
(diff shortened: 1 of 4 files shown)


kbond (approved):

kbond:
@ndench, can you confirm this fixes your issue in #443 and I'll tag a bugfix.

ndench:
@kbond thanks for fixing this so quickly! I have installed `1.x-dev` and can confirm it resolves by issue :smile:

PR #448, opened 2023-04-13
docs: add static ide hint to stories

adrianrudnik:
The story docs mention a way to hint the return value of story state references. It uses static method calls to retrieve them, but does not declare them as static, leading to IDE warnings.

It is also not mentioned that the magic getters, through `\Zenstruck\Foundry\Story::__callStatic` already ensure the story is loaded correctly, so calling `::load` does not add anything to the test.


nikophil (approved):

nikophil:
thanks @adrianrudnik

PR #449, opened 2023-04-15
minor: improve makefile

nikophil:
One thing I don't like with the current makefile is that when the containers are not up, something like `make test` will create an error.

This PR fixes this behavior, for such scenario:
- `make test` for the first time => the images will be built, vendor installed, etc
- turn of computer
- `make test` another time, containers are down, but everything else is OK => containers will be created

I think this covers 95% of the usages...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,7 @@ validate: sca psalm test database-validate-mapping ### Run sca, full test suite 
 
 .PHONY: test
 test: vendor ### Run PHPUnit tests suite
+	@$(MAKE) --no-print-directory docker-start-if-not-running
 	@${DC_EXEC} -e USE_ORM=${USE_ORM} -e USE_ODM=${USE_ODM} ${PHP} vendor/bin/simple-phpunit --configuration ${PHPUNIT_CONFIG_FILE} $(ARGS)
 
 .PHONY: sca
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -105,23 +110,39 @@ database-validate-mapping: database-drop-schema ### Validate mapping in Zenstruc
...
 	@${DOCKER_PHP_WITHOUT_XDEBUG} /usr/bin/composer $(ARGS)
 
 vendor: $(DOCKER_PHP_CONTAINER_FLAG) composer.json $(wildcard composer.lock) $(wildcard .env)
+	@$(MAKE) --no-print-directory docker-start-if-not-running
 	@${DC_EXEC} -e SYMFONY_REQUIRE=${SYMFONY_REQUIRE} ${PHP} php -d 'xdebug.mode=off' /usr/bin/composer update ${COMPOSER_UPDATE_OPTIONS}
 	@touch -c $@ composer.json .env composer.lock
 
+.PHONY: docker-start-if-not-running
+docker-start-if-not-running: ### some xxx
+	@if [ -f "$(DOCKER_PHP_CONTAINER_FLAG)" ] ; then \
```
nikophil, line 130: this line ensures the containers have been built at lease one time, so that the `--no-build` flag will not create an error.

otherwise, if `DOCKER_PHP_CONTAINER_FLAG` a docker build will be triggered

kbond (approved):

Issue #445, opened 2023-04-04
Usage of `repository()` + PHPStan makes unfixable error


er1z:
Suppose:
```
$repo = MyFactory::repository();
$subscription = $repo->customRepoMethod($user);
```

Will make PHPStan screaming. So tried:
```
/** @var RepositoryProxy<My>|MyRepository $repo */
$repo = MyFactory::repository();
$subscription = $repo->customRepoMethod($user);
```
but it's no longer possible: https://phpstan.org/blog/phpstan-1-10-comes-with-lie-detector

Any idea how to solve this?

nikophil:
hello @er1z 

do you have in your factory the `phpstan-method` phpdoc suggested [in the doc](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#generate)?

there is a specific annotation for the `repository()` method:
```php
/**
 * @phpstan-method static RepositoryProxy<Post> repository()
 */
```

er1z:
How does it solve the problem that I need to have signatures of proxied repository?

nikophil:
the phpdoc I gave you above should fix it: it gives some hints to phpstan that `repository()` method returns a `RepositoryProxy<YourEntity>`
and the `RepositoryProxy` is a mixin of `EntityRepository` so this should work
https://github.com/zenstruck/foundry/blob/1.x/src/RepositoryProxy.php#L26

Edit: for custom repo methods, maybe you can change the annotatio on phpstan-method to the one you used with `@var`. I've still never used phpstan with this new option, so not sure if this will work 🤔

er1z:
Yes, all this issue is about custom repository methods. And it won't work anymore due to the reason described on PHPStan's website.

Now it tells that (don't remember exact error) that `MyRepository` is not a derivative of returned type `RepositoryProxy`. And the only solution is to put it into ignores.

Gonna check if annotation like this work via `phpstan-method` tomorrow.

kbond:
Sounds like, in addition to the new phpstan mode not trusting vars, it doesn't trust mixins either.

nikophil:
I've made some tests here, using phpstan 1.10 + bleeding edge and everything is OK if `@phpstan-method` is updated to:
```php
@phpstan-method static MyRepository&RepositoryProxy<MyEntity> repository()
```
this might be a lie, because that's not the real type, but phpstan does not complains about it.

(and BTW, it seems to me that `@mixin` is still well understood)

[...]

PR #450, opened 2023-04-15
minor: add $persistentManagerName param

### LINKED ISSUE #447: Can I specify $persistentManagerName for repository?
Hi! Thanks for awesome library!

Is there any possibility to specify $persistentManagerName here https://github.com/zenstruck/foundry/blob/1.x/src/Configuration.php#L189?

nikophil:
fixes #447 

I've not provided tests, because the code is trivial, but this would need to add a new document manager in the test :shrug: 

/cc @kerstvo

EDIT: I just realized that adding a param to `ModelFactory::repository()` will create an error in psalm in userland, because of the `@phpstan-method` annotation :disappointed: 

https://github.com/zenstruck/foundry/actions/runs/4707671150/jobs/8349671474?pr=450#step:8:18

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -175,7 +175,7 @@ final class Configuration
      * @phpstan-param Proxy<TObject>|TObject|class-string<TObject> $objectOrClass
      * @phpstan-return RepositoryProxy<TObject>
      */
-    public function repositoryFor(object|string $objectOrClass): RepositoryProxy
+    public function repositoryFor(object|string $objectOrClass, string|null $persistentManagerName = null): RepositoryProxy
     {
         if ($objectOrClass instanceof Proxy) {
             $objectOrClass = $objectOrClass->object();
@@ -186,7 +186,7 @@ final class Configuration
         }
 
         /** @var EntityRepository<TObject>|null $repository */
-        $repository = $this->managerRegistry()?->getRepository($objectOrClass);
+        $repository = $this->managerRegistry()?->getRepository($objectOrClass, $persistentManagerName);
 
         if (!$repository) {
             throw new \RuntimeException(\sprintf('No repository registered for "%s".', $objectOrClass));
--- a/src/ModelFactory.php
+++ b/src/ModelFactory.php
@@ -260,9 +260,9 @@ abstract class ModelFactory extends Factory
     /**
      * @phpstan-return RepositoryProxy<TModel>
      */
-    final public static function repository(): RepositoryProxy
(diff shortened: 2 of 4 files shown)


kbond:
> I just realized that adding a param to ModelFactory::repository() will create an error in psalm in userland

You mean on existing factories? I guess we should we update the maker/docs?

nikophil:
> You mean on existing factories?

yes, all existing factories will create an error with psalm if we make this move... We can update update maker and docs, but still, this could be pretty annoying...

kbond:
What about phpstan?

nikophil:
even in max level, phpstan does not complain. But psalm is actually true: one param is missing in the methods prototype in the phpdoc

kbond:
Huh, well I don't want to _not_ add a feature because it breaks static analysis, but that's just my opinion.

kbond:
What about starting a UPGRADE.md with a note for the next feature release?

nikophil:
ok, this seems reasonable :)

I feel the same, otherwise it would restrict too much further updates... see this issue as well https://github.com/zenstruck/foundry/issues/445#issuecomment-1500440459

kerstvo:
It seems that problem is more complicated.
I apply your changes in my project and got error
image

kbond:
I think we need to have a `Factory::withManagerName()` where you actually configure the alternate manager name per factory.

This would remove the need for `$persistentManagerName` in `ModelFactory::repository()` (and not break static analysis for existing factories).

nikophil:
this seems not possible because `ModelFactory::repository()` is static...

I think this needs to be hardcoded directly in the class, and then have a factory class which is bound to a given manager. ex:
```php
class SomeFactoryForSpecificManager
{
    {
        return SomeEntity::class;
    }

    protected static function getManagerName(): ?string
    {
        return 'specific-manager;
    }
}
```
@kerstvo would this fix your problem?

kbond:
So `ModelFactory` would look like?

```
abstract class ModelFactory extends Factory
{
    // ...
    
    protected static function getManagerName(): ?string
    {
        return null;
    }
}
```

nikophil:
That's what I'm suggesting :)

kbond:
Since we're deprecating ModelFactory soon, maybe we should just add this feature to #452?

We'll need to update some function parameters and the anon. proxy generator.

nikophil:
I'm closing this PR, which targets 1.x and is full of conflicts. The issue stays opens, though

PR #451, opened 2023-04-17
fix: makefile minor fixes

nikophil:


Makefile
```diff
@@ -126,14 +126,14 @@ vendor: $(DOCKER_PHP_CONTAINER_FLAG) composer.json $(wildcard composer.lock) $(w
 	@touch -c $@ composer.json .env composer.lock
 
 .PHONY: docker-start-if-not-running
-docker-start-if-not-running: ### some xxx
```
nikophil, line 129: oups :innocent: 

BTW I remove the doc because I don't think this target will be directly used by users...

PR #452, opened 2023-04-17
refactor: split Factory class

nikophil:
sorry for the :t-rex: PR :sweat_smile: 

see some observations in the PR comments. I tried to prevent BC breaks at maximum, I cannot gurantee static analysis will not break in user land (I still have to test this branch on my projects which use Foundry)

some works is still needed (maybe in other PRs):
- create some `AttributeNormalizer` classes, and extract all this logic from factories
- update maker, in order to ask if the user (or automatically decide?) wants to extend from `ObjectFactory` or `PeristenceObjectFactory`
- totally remove everything related to persistence from `Configuration`
- create a class `\Zenstruck\Foundry\Persistence\ObjectManager`


src/AnonymousFactory.php
```diff
@@ -11,29 +11,33 @@
...
 /**
  * @template TModel of object
- * @template-extends Factory<TModel>
  *
  * @author Kevin Bond <kevinbond@gmail.com>
  *
  * @deprecated
  */
-final class AnonymousFactory extends Factory implements \Countable, \IteratorAggregate
+final class AnonymousFactory implements \Countable, \IteratorAggregate
```
nikophil, line 25: we cannot extend `Factory` anymore because of the method static `new` which now exists in `BaseFactory` and does not have the same prototype

src/AnonymousFactory.php
```diff
@@ -188,10 +198,42 @@ public function assert(): RepositoryAssertions
...
     {
-        return self::configuration()->repositoryFor($this->class());
+        return PersistentObjectFactory::persistenceManager()->repositoryFor($this->class);
+    }
+
+    /**
+     * @return Proxy&TModel
+     * @phpstan-return Proxy<TModel>
+     */
+    public function create(array|callable $attributes = []): mixed
```
nikophil, line 214: I've added all these methods in order to keep BC

src/BaseFactory.php
```diff
@@ -0,0 +1,262 @@
...
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @immutable
+ *
+ * @template T
+ *
+ * @method static list<T> createMany(int $number, Attributes $attributes = [])
+ */
+abstract class BaseFactory
```
nikophil, line 30: this "main" factory class is called `BaseFactory` in order to keep BC

src/BaseFactory.php
```diff
@@ -0,0 +1,262 @@
...
+            throw new \RuntimeException('Factories with dependencies (Factory services) cannot be created before foundry is booted.', 0, $e);
+        }
+
+        $factory = $factory
+            ->withAttributes($attributes)
+            ->initialize();
+
+        if (!$factory instanceof static) {
+            throw new \TypeError(\sprintf('"%1$s::initialize()" must return an instance of "%1$s".', static::class));
+        }
```
nikophil, line 69: this shall be kept, because we cannot add `: static` return type to `BaseFactory::initialize()` without BC. I've added the attribute `#[ReturnTypeWillChange]` on the method

src/BaseFactory.php
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * @param Attributes|string $attributes
+     */
+    final public static function new(array|callable|string $attributes = [], string ...$states): static
+    {
+        if (\is_string($attributes)) {
+            $states = [$attributes, ...$states];
+            $attributes = [];
+        }
```
nikophil, line 55: I've discovered this thing... never used these string states. This was not in your foundry-next skeleton, would you like that we add a deprecation?

src/Factory.php
```diff
@@ -11,461 +11,28 @@
...
  * @template TObject of object
- * @abstract
+ * @template-extends PersistentObjectFactory<TObject>
  *
  * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @deprecated since 1.32, use "Zenstruck\Foundry\Persistence\PersistentObjectFactory" instead
  */
-class Factory
+abstract class Factory extends PersistentObjectFactory
```
nikophil, line 24: our current factories are all `PersistentObjectFactory`

src/Factory.php
nikophil, line 1: this class is now deprecated and extends `PersistentObjectFactory`

I've made a BC break here: with the new system, we cannot implement `Factory::class()` which is static, but in this class we can only access to the class name on an instance of the class. So I made it abstract... It breaks the initial system which has been deprecated for 2 years ago... I think this is OK, because people had 2 years to move on the new system. Maybe we would have to keep maintenance on a `1.31.x` branch...

src/FactoryCollection.php
```diff
@@ -46,28 +42,28 @@ public function __construct(private Factory $factory, ?int $min = null, ?int $ma
...
     {
         return new self($factory, 0, null, $sequence, true);
     }
 
     /**
-     * @return list<TObject&Proxy<TObject>>
+     * @param Attributes $attributes
      *
-     * @phpstan-return list<Proxy<TObject>>
+     * @return list<T>
```
nikophil, line 69: this actually sometimes returns `Proxy<T>`, when used along with `PersistenceObjectFactory`, but i've hardcoded this in some `@method` phpdoc in `PersistenceObjectFactory`. Still, this creates errors with `anonymous()` helper in phpstan (see the phpstan baseline)

src/FactoryManager.php
```diff
@@ -0,0 +1,77 @@
...
+
+    /** @var CallableInstantiator */
+    private Instantiator|\Closure $instantiator;
+    private Faker\Generator $faker;
+
+    /**
+     * @param BaseFactory[] $factories
+     * @param CallableInstantiator $instantiator
+     */
+    public function __construct(iterable $factories = [], Instantiator|\Closure $instantiator = null, Faker\Generator $faker = null)
```
nikophil, line 36: for now, all the methods here are not static, in opposite to your foundry-next skeleton

but this can change, now everything is well split, this might be easier

src/Persistence/PersistenceManager.php
```diff
@@ -0,0 +1,98 @@
...
+use Doctrine\Persistence\ManagerRegistry;
+use Doctrine\Persistence\ObjectManager;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Zenstruck\Foundry\Proxy;
+use Zenstruck\Foundry\RepositoryProxy;
+
+/**
+ * @internal
+ */
+final class PersistenceManager
```
nikophil, line 17: I'd like to handle everything related to persistence here

src/Persistence/PersistentObjectFactory.php
```diff
@@ -0,0 +1,471 @@
...
+    /** @var array<callable(T&Proxy,Parameters):void> */
+    private array $afterPersist = [];
+
+    private bool $persist = true;
+    private bool $cascadePersist = false;
+
+    /**
+     * @internal
+     */
+    final public static function bootPersistentObjectFactory(PersistenceManager $persistenceManager): void
```
nikophil, line 56: not sure about all of these new methods, but we need to initialize separately  the `PersistenceManager` from `Configuration` and `FactoryManager`
and I cannot override `BaseFactory::boot()` method here, because the parameters would be different

src/Persistence/PersistentObjectFactory.php
```diff
@@ -0,0 +1,471 @@
...
+                if (!$callbacks) {
+                    return;
+                }
+
+                foreach ($callbacks as $callback) {
+                    $proxy->executeCallback($callback, $parameters);
+                }
+
+                $proxy->save(); // save again as afterPersist events may have modified
+            });
```
nikophil, line 280: I kept the "old" way to do this... I saw you've changed that in foundry-next, we still can do this in another PR

src/Persistence/PersistentObjectFactory.php
```diff
@@ -0,0 +1,471 @@
...
+            // ensure attribute Factories' are also not persisted
+            $value = $value->withoutPersisting();
+        }
+
+        if (!self::persistenceManager()->hasManagerRegistry()) {
+            return $value->create()->object();
+        }
+
+        try {
+            $objectManager = self::persistenceManager()->objectManagerFor(static::class());
```
nikophil, line 321: all the following code should go in a class dedicated to the ORM

src/Test/TestState.php
```diff
@@ -109,7 +113,7 @@ public static function addGlobalState(callable $callback): void
      */
     public static function bootFoundryForUnitTest(): void
```
nikophil, line 113: I'm wondering why all this code related to "boot" foundry is in this class, `TestState`? I'm thinking of moving all of this in a class which is dedicated for the "boot" logic, and keep here only which is related to the global state. WDYT?
BTW, shouldn't this class be `@internal`, now it only has internal and deprecated methods?

src/Test/TestState.php
```diff
@@ -125,23 +129,34 @@ public static function bootFoundryForUnitTest(): void
...
     }
 
     /**
      * @internal
      */
     public static function shutdownFoundry(): void
     {
-        Factory::shutdown();
+        BaseFactory::shutdown();
+        if (class_exists(PersistenceManager::class)) {
```
nikophil, line 156: I'm anticipating the split of the repositories :smile:

src/functions.php
```diff
@@ -34,25 +36,53 @@ function factory(string $class, array|callable $defaultAttributes = []): Anonymo
...
+ * @template T of object
  *
- * @param class-string<TObject> $class
+ * @param class-string<T> $class
  *
- * @return Factory<TObject>
+ * @return Factory<T>
  */
-function anonymous(string $class, array|callable $defaultAttributes = []): Factory
+function anonymous(string $class, array|callable $defaultAttributes = []): PersistentObjectFactory
```
nikophil, line 42: here I return a `PersistentObjectFactory` but how should we do if we only want to have an anonymous object factory?

that's the same for all other functions here... should they return `T` or `Proxy<T>`?

tests/Unit/FactoryTest.php
```diff
@@ -35,22 +41,31 @@ final class FactoryTest extends TestCase
 
     /**
      * @test
+     * @dataProvider attributesProvider
      */
-    public function can_instantiate_object(): void
```
nikophil, line 39: now that `Factory` is abstract this could no longer work

tests/Unit/FactoryTest.php
```diff
@@ -85,58 +97,6 @@ public function lazy_values_are_only_calculated_if_needed(): void
         $this->assertSame(1, $count);
     }
 
-    /**
-     * @test
-     * @group legacy
-     */
-    public function can_instantiate_many_objects_legacy(): void
```
nikophil, line 92: same

tests/Unit/FactoryTest.php
```diff
@@ -215,23 +175,12 @@ public function can_add_after_instantiate_events(): void
         $this->assertSame(2, $object->getViewCount());
     }
 
-    /**
-     * @test
-     */
-    public function can_register_custom_faker(): void
```
nikophil, line 221: there is no more setter for faker, this test is not useful anymore

tests/bootstrap.php
```diff
@@ -13,4 +13,4 @@
 
 require \dirname(__DIR__).'/vendor/autoload.php';
 
-(new Filesystem())->remove(__DIR__.'/../var');
+(new Filesystem())->remove(__DIR__.'/../var/cache');
```
nikophil, line 16: I use to keep some files in `./var` this is useful sometimes...

kbond:
Thanks for working on this @nikophil - I'll try and carve out some time to review soon.

/cc @OskarStark this should open the door for an `ArrayFactory`

src/Object/ObjectFactory.php
```diff
@@ -0,0 +1,124 @@
...
+    /** @var array<callable(Parameters,class-string<T>):Parameters> */
+    private array $beforeInstantiate = [];
+
+    /** @var array<callable(T,Parameters):void> */
+    private array $afterInstantiate = [];
+
+    /**
+     * @return class-string<T>
+     */
+    abstract protected static function getClass(): string;
```
nikophil, line 41: For BC sake, I kept the same name, and did not changed for `class()`, even if I do prefer this name

src/Object/ObjectFactory.php
```diff
@@ -0,0 +1,132 @@
...
+     */
+    public static function class(): string
+    {
+        return static::getClass();
+    }
+
+    /**
+     * @return class-string<T>
+     */
+    abstract protected static function getClass(): string;
```
nikophil: not sure what to do here: this is made here to prevent BC breaks (a public method is need to list factories in `\Zenstruck\Foundry\Bundle\Maker\Factory\FactoryClassMap`)

src/BaseFactory.php
kbond: Neither have I, yes, let's deprecate - ~is it even documented?~ It is https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#reusable-model-factory-states

src/BaseFactory.php
```diff
@@ -0,0 +1,294 @@
...
+
+        self::factoryManager()->faker()->unique(true); // reset unique
+        self::$factoryManager = null;
+        self::$configuration = null;
+    }
+
+    /**
+     * @throws FoundryNotBootedException
+     * @internal
```
kbond:
```suggestion
     * @internal
     * @throws FoundryNotBootedException
```

src/Factory.php
kbond: The idea is we'd rename `BaseFactory` to `Factory` at some point post 2.0?

src/Object/ObjectFactory.php
kbond: Why can't we use `::class()` above in the `FactoryClassMap`?

src/Object/ObjectFactory.php
```diff
@@ -0,0 +1,132 @@
...
+    /** @var array<callable(Parameters,class-string<T>):Parameters> */
+    private array $beforeInstantiate = [];
+
+    /** @var array<callable(T,Parameters):void> */
+    private array $afterInstantiate = [];
+
+    /**
+     * @return class-string<T>
+     */
+    public static function class(): string
```
kbond: Can we make this abstract and remove `getClass()` below?

src/Persistence/PersistentObjectFactory.php
kbond: Ok to leave for now I think - it's internal so we can change later.

src/Persistence/PersistentObjectFactory.php
```diff
@@ -0,0 +1,474 @@
...
+        if (!self::isPersistentObjectFactoryBooted()) {
+            return;
+        }
+
+        self::$persistenceManager = null;
+    }
+
+    /**
+     * @throws FoundryNotBootedException
+     * @internal
```
kbond:
```suggestion
     * @internal
     * @throws FoundryNotBootedException
```

src/functions.php
kbond: Can we have this return `PersistentObjectFactory<T>|ObjectFactory<T>` depending on if the class is managed by doctrine?

kbond: And the others: `T|Proxy<T>`?

src/Object/ObjectFactory.php
kbond: Since this is a new class and we're deprecating `ModelFactory`, I think this is a good time to do this.

kbond (commented):
Initial high-level review - looking really great!

OskarStark:
I am super exited 😎

src/Factory.php
nikophil: yes: in 2.0, the current `Factory` will disappear, and because `BaseFactory` is internal we will be able to rename it in the first version of `2.0`

src/functions.php
nikophil: I think I'll return the union type for PHPStorm, and static analysis will understand it better if https://github.com/zenstruck/foundry/pull/456 gets merged

src/Object/ObjectFactory.php
nikophil: :thinking: hmmm OK what  I did was wrong! I've overridden `self::class()` in `ModelFactory` now. When users will move from `ModelFactory` to one of `PersistentObjectFactory` or `ObjectFactory`, they'll also need to mode from `protected function getClass()` to `public function class()`

src/Object/ObjectFactory.php
nikophil: done

src/BaseFactory.php
nikophil: done. I've also removed the related doc

src/FactoryCollection.php
nikophil: this will be fixed with https://github.com/zenstruck/foundry/pull/456

src/Persistence/PersistentObjectFactory.php
nikophil: will be done in another PR

src/functions.php
nikophil: will be done in a further PR

PR #456, opened 2023-04-23
feat: create PhpStan type resolvers

nikophil:
Here is a draft suggestion to add PHPStan type resolvers for factories' methods.

This would override the `@phpstan-method` stuff in the factories, and it has the big advantage to give us back the control over the type system, we could then update these as long as we're updating the code.

It has the drawback that we may need to support two versions of this: one for PHPStan, and another one for Psalm (and I'll have to learn how to make this for Psalm :sweat_smile:) For now we can only drop the `@phpstan-method` generation with maker, and keep the `@psalm-method` annotation.

Some stuff still needs to be done:
- [ ] ~I would like to merge https://github.com/zenstruck/foundry/pull/452 before~ _we will merge it in the branch `feat/split-factories`_
- [ ] We will be able to fix all type problems in https://github.com/zenstruck/foundry/pull/452 (see [here](https://github.com/zenstruck/foundry/pull/452#discussion_r1168659984) and [here](https://github.com/zenstruck/foundry/pull/452#discussion_r1168687668))
- [x] add tests
- [x] provide a `foundry-phpstan.neon` file
- [ ] update doc: remove references to `@phpstan-method` and show how it should be configured
- [ ] update `make:factory`
- [ ] create a `CHANGELOG.md` to explain existing `@phpstan-method` can be dropped in userland

we can even imagine that once we've made the subtree split, we'd create a `zenstruck/foundry-phpstan` package which will be a phpstan extension and be auto installed with `phpstan/extension-installer`

WDYT of this?

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,6 +1,21 @@
 parameters:
 	ignoreErrors:
 		-
+			message: "#^Method Zenstruck\\\\Foundry\\\\AnonymousFactory\\:\\:create\\(\\) should return Zenstruck\\\\Foundry\\\\Proxy\\<TModel of object\\> but returns \\(TModel of object\\)\\|Zenstruck\\\\Foundry\\\\Proxy\\<TModel of object\\>\\.$#"
+			count: 1
+			path: src/AnonymousFactory.php
+
+		-
+			message: "#^Method Zenstruck\\\\Foundry\\\\AnonymousFactory\\:\\:many\\(\\) should return Zenstruck\\\\Foundry\\\\FactoryCollection\\<TModel of object\\> but returns Zenstruck\\\\Foundry\\\\FactoryCollection\\<object\\>\\.$#"
+			count: 1
+			path: src/AnonymousFactory.php
+
+		-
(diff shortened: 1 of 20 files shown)


kbond (commented):
Neat! Yeah, I agree to wait for #452

phpstan-baseline.neon
nikophil, line 1: I couldn't get rid of all these phpstan errors :shrug:

phpstan-foundry.neon
nikophil, line 1: In a further PR, I'll remove the `@phpstan-method` annotations added by `make:factory` and encourage people to include this file in there `phpstan.neon` 

maybe when I'll find time I'll create a real phpstan extension in a dedicated repo

phpstan.neon
nikophil, line 1: I'm more and more thinking to analyse `./tests` folder with phpstan... this will be useful to battle test the phpstan extension

phpstan-foundry.neon
kbond: Should we document the process of adding to your phpstan config? Can it be autoconfigured?

kbond (approved):
I'm good to merge this into feat/split-factories

What about the todos though?

phpstan-foundry.neon
nikophil: as I told you, not sure how to handle the autoconfiguration... I'm afraid it's made thanks to the type of the repository of the repository of the extension, they are all tagged `"type": "phpstan-extension"`

kbond: https://github.com/phpstan/extension-installer#instructions-for-extension-developers

kbond (approved):

PR #457, opened 2023-04-25
feat: allow setters as factory attribute

### LINKED ISSUE #454: Since 1.30.0 getDefaults array key must be an Entity field
Hi everyone,
Since the 1.30.0, if we define in getDefaults a key that not refer to an Entity property, an error is trigger.

In my case, an entity as a `$documents` property, but we use:
 - `addInitialDocument`
 - `getInitialDocuments`
 - `addAdditionalDocument`
 - `getAdditionalDocuments`

These one set a Discriminator (adder) or use filter (getter), this one work well in Factory:
```php
    {
        return [
            //...
            'additionalDocuments' => DocumentFactory::new()->asAdditional()->many(1, 3),
            //...
        ];
    }
```

But since the 1.3.0, this one trigger an error:

> In Factory.php line 447:
>                                                                                                           
> Field App\Entity\ChangeRequest::$additionalDocuments does not exist or is not a relationship field.

nikophil:
fixes #454

### DIFF
--- a/tests/Functional/ORMModelFactoryTest.php
+++ b/tests/Functional/ORMModelFactoryTest.php
@@ -105,6 +105,22 @@ final class ORMModelFactoryTest extends ModelFactoryTest
     /**
      * @test
      */
+    public function one_to_many_with_two_relationships_same_entity_and_adders(): void
+    {
+        $category = CategoryFactory::createOne([
+            'addPost' => PostFactory::new(),
+            'addSecondaryPost' => PostFactory::new(),
+        ]);
+
+        $this->assertCount(1, $category->getPosts());
+        $this->assertCount(1, $category->getSecondaryPosts());
+        PostFactory::assert()->count(2);
+        CategoryFactory::assert()->count(1);
+    }
+
+    /**
+     * @test
+     */
     public function inverse_one_to_many_relationship_without_cascade(): void
     {
         UserFactory::createOne([
@@ -620,6 +636,44 @@ final class ORMModelFactoryTest extends ModelFactoryTest
         EntityWithEnumFactory::assert()->count(1);
 
         self::assertSame($entityWithEnum->object(), $entityWithEnum2->object());
+    }
+
+    /**
+     * @test
+     * @dataProvider addManyToOneWithExtraAttributes
+     */
+    public function it_adds_many_to_one_with_extra_attributes(string $extraAttributeName): void
+    {
+        $post = PostFactory::createOne([$extraAttributeName => CategoryFactory::new(['name' => 'category name'])]);
+
+        self::assertSame('category name', $post->getCategory()->getName());
+        CategoryFactory::assert()->count(1);
+        PostFactory::assert()->count(1);
+    }
+
+    public static function addManyToOneWithExtraAttributes(): iterable
+    {
+        yield ['extraCategoryBeforeInstantiate'];
+        yield ['extraCategoryAfterInstantiate'];
+    }
+
+    /**
+     * @test
+     * @dataProvider addOneToManyWithExtraAttributes
+     */
+    public function it_adds_one_to_many_with_extra_attributes(string $extraAttributeName): void
+    {
+        $category = CategoryFactory::createOne([$extraAttributeName => PostFactory::createMany(2)]);
+
+        self::assertCount(2, $category->getPosts());
+        CategoryFactory::assert()->count(1);
+        PostFactory::assert()->count(2);
+    }
+
+    public static function addOneToManyWithExtraAttributes(): iterable
+    {
+        yield ['extraPostsBeforeInstantiate'];
+        yield ['extraPostsAfterInstantiate'];
     }
 
     protected function categoryClass(): string
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1893,7 +1893,7 @@ Later, you can access the story's state when creating other fixtures:
         }
 
     Now your IDE will know ``CategoryStory::php()`` returns an object of type ``Category``.
-    
+
     Using a magic method also does not require a prior ``::load()`` call on the story, it will initialize itself.
 
 .. note::
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -15,8 +15,9 @@ use Doctrine\ODM\MongoDB\Mapping\ClassMetadata as ODMClassMetadata;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata as ORMClassMetadata;
 use Faker;
+use Symfony\Component\PropertyInfo\Extractor\ReflectionExtractor;
 use Zenstruck\Foundry\Exception\FoundryBootException;
-use Zenstruck\Foundry\Persistence\InversedRelationshipCascadePersistCallback;
+use Zenstruck\Foundry\Persistence\InversedRelationshipPostPersistCallback;
 use Zenstruck\Foundry\Persistence\PostPersistCallback;
 
 /**
@@ -384,6 +385,10 @@ class Factory
 
         $relationshipMetadata = self::getRelationshipMetadata($objectManager, $this->class, $name);
 
+        if (!$relationshipMetadata) {
+            return $value->create()->object();
+        }
+
         if ($relationshipMetadata['isOwningSide']) {
             $cascadePersist = $relationshipMetadata['cascade'];
         } else {
@@ -392,7 +397,7 @@ class Factory
             $cascadePersist = $relationshipMetadata['cascade'];
 
             if ($this->isPersisting() && null !== $relationshipField && false === $cascadePersist) {
-                return new InversedRelationshipCascadePersistCallback($value, $relationshipField, $isCollection);
+                return new InversedRelationshipPostPersistCallback($value, $relationshipField, $isCollection);
             }
         }
 
@@ -439,18 +444,18 @@ class Factory
     }
 
     /**
-     * @param class-string $class
+     * @param class-string $factoryClass
      *
-     * @return array{cascade: bool, inversedField: ?string, inverseIsCollection: bool, isOwningSide: bool}
+     * @return array{cascade: bool, inversedField: ?string, inverseIsCollection: bool, isOwningSide: bool}|null
      */
-    private static function getRelationshipMetadata(EntityManagerInterface $entityManager, string $class, string $relationshipName): array
+    private static function getRelationshipMetadata(EntityManagerInterface $entityManager, string $factoryClass, string $relationshipName): array|null
     {
-        $metadata = $entityManager->getClassMetadata($class);
+        $metadata = $entityManager->getClassMetadata($factoryClass);
 
         $relationshipMetadata = $metadata->associationMappings[$relationshipName] ?? null;
 
         if (!$relationshipMetadata) {
-            throw new \RuntimeException("Field {$class}::\${$relationshipName} does not exist or is not a relationship field.");
+            return null;
         }
 
         return [
--- a/src/Persistence/InversedRelationshipPostPersistCallback.php
+++ b/src/Persistence/InversedRelationshipPostPersistCallback.php
@@ -19,7 +19,7 @@ use Zenstruck\Foundry\Proxy;
 /**
  * @internal
  */
-final class InversedRelationshipCascadePersistCallback implements PostPersistCallback
+final class InversedRelationshipPostPersistCallback implements PostPersistCallback
 {
     public function __construct(
         private Factory $factory,
--- a/tests/Fixtures/Document/ODMCategory.php
+++ b/tests/Fixtures/Document/ODMCategory.php
@@ -32,4 +32,9 @@ class ODMCategory
     {
         $this->name = $name;
     }
+
+    public function updateName(?string $name): void
+    {
+        $this->name = $name;
+    }
 }
--- a/tests/Fixtures/Entity/Category.php
+++ b/tests/Fixtures/Entity/Category.php
@@ -48,6 +48,11 @@ class Category
         $this->name = $name;
     }
 
+    public function updateName(?string $name): void
+    {
+        $this->name = $name;
+    }
+
     public function getPosts(): Collection
     {
         return $this->posts;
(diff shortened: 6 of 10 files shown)


kbond:
@mpiot are you able to confirm this fixes #454?

OskarStark:
cc @silasjoisten

mpiot:
@kbond nope, it trigger an error that say the id is not set:

> An exception occurred while executing a query: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "change_request_id" of relation "change_request_document" violates not-null constraint

```php
class ChangeRequest
{
    #[ORM\OneToMany(mappedBy: 'changeRequest', targetEntity:ChangeRequestDocument::class, cascade: ['persist', 'remove'], orphanRemoval: true)]
    #[ORM\OrderBy(['id' => 'asc'])]
    private Collection $documents;

    public function addAdditionalDocument(ChangeRequestDocument $additionalDocument): self
    {
        if (true === $additionalDocument->isInitial()) {
            throw new \Exception('The addAdditionalDocument must receive a ChangeRequestDocument with initial as false.');
        }

        if (!$this->documents->contains($additionalDocument)) {
            $this->documents->add($additionalDocument);
            $additionalDocument->setChangeRequest($this);
        }

        return $this;
    }
}
```

```php
class ChangeRequestDocument
{
    #[Assert\DisableAutoMapping]
    #[ORM\ManyToOne(inversedBy: 'documents')]
    private ChangeRequest $changeRequest = null;

    public function getChangeRequest(): ChangeRequest
    {
        return $this->changeRequest;
    }

    public function setChangeRequest(?ChangeRequest $changeRequest): self
    {
        $this->changeRequest = $changeRequest;

        return $this;
    }
}
```

```php
final class ChangeRequestFactory extends ModelFactory
{
    {
        return [
            'additionalDocument' => ChangeRequestDocumentFactory::new()->asAdditional()->many(1, 3),
            'createdBy' => $user = UserFactory::randomOrCreate(),
            'updatedBy' => $user,
            'createdAt' => $date = \DateTimeImmutable::createFromMutable(self::faker()->dateTimeThisDecade()),
            'updatedAt' => $date,
        ];
    }
}
```

```php
final class ChangeRequestDocumentFactory extends ModelFactory
{
    {
        return [
            'category' => self::faker()->text(),
            'updated' => self::faker()->boolean(),
            'documentReferences' => self::faker()->optional()->text(),
        ];
    }
}
```

kbond:
@mpiot, do you have a `ChangeRequest::removeAdditionalDocument()` method? I believe property-accessor requires both an _adder_ and a _remover_ to work.

mpiot:
@kbond yes I have a `removeAdditionalDocument`, just not put it in the example :)

In this case the error is kind of strange, it find the documents because try to persist them, but the change request id is not set in the document. Like if the adder was not called.

It cause by: src/Factory.php line 387-390, if the property is not found, it not create the related entities.

kbond:
@nikophil, let's add a test case that's as similar to @mpiot's scenario as possible to figure this out.

Something like `Product::addAdditionalCategory(Category $category)` (in our fixtures).

nikophil:
Hi @mpiot 

Maybe I' missing something, but I don't understand how your code could work:
```php
final class ChangeRequestFactory extends ModelFactory
{
    {
        return [
            'additionalDocument' => ChangeRequestDocumentFactory::new()->asAdditional()->many(1, 3),
        ];
    }
}

class ChangeRequest
{
    public function addAdditionalDocument(ChangeRequestDocument $additionalDocument): self;
}
```

you're passing an array to a method which expects on object. I've even tested on foundry v1.29 and still get the same error:
> Expected argument of type "Zenstruck\Foundry\Tests\Fixtures\Entity\Post", "array" given at property path "addPost".

nikophil:
Hi!

ok, I think I've misunderstood the problem at first.

What I was fixing and testing: the ability to pass random method (it does not have to be an adder or a setter) of the model as an attribute. ex:
```php
PostFactory::createOne(['updateName' => 'some name']);
```

The original issue, was the ability to pass a "virtual" property as an attribute:
```php
class ChangeRequest
{
    private Collection $documents;

    // it only work if the virtual property has adder/remover/getter, this is by design in property access component
    public function addAdditionalDocument(ChangeRequestDocument $additionalDocument): self;
    public function removeAdditionalDocument(ChangeRequestDocument $additionalDocument): self;
    public function getAdditionalDocuments(): Collection;
}

ChangeRequestFactory::createOne([
    // notice the plural form. Passing "additionalDocument" never worked and still does not work
    'additionalDocuments' => ChangeRequestDocumentFactory::createMany(2)
]);
```

It happens that both of the problems are fixed by not throwing an error [here](https://github.com/zenstruck/foundry/blob/1.x/src/Factory.php#L453).

I noticed a glitch with the second use case: this code was only creating 2 posts:
```php
        $category = CategoryFactory::createOne([
            'additionalPosts' => PostFactory::new(['title' => 'additional post'])->many(2),
            'addPost' => PostFactory::new(['title' => 'regular post added']),
            'posts' => PostFactory::new(['title' => 'regular post'])->many(2),
        ]);
```
This was because,  in our tests, we don't have a `cascade: persist` on the relationship `Category::$posts`, and then the normalization [first sets `null` on the property](https://github.com/zenstruck/foundry/blob/1.x/src/Factory.php#L124), before it is hydrated through post persist callback. I've fixed this in this PR.

This bug could occur because now we have the possibility to pass two attributes to a factory to modify the same property. I'm afraid it would open the door for some subtle new bugs, that's what I was saying [here](https://github.com/zenstruck/foundry/issues/454#issuecomment-1514226291). 

I've also added a test on virtual properties in `FactoryDoctrineCascadeTest` in order to assert the behavior is working with and without cascade:persist.

Last but not least, I've not managed to reproduce your error @mpiot, everything seems to work correctly in our tests
> It cause by: src/Factory.php line 387-390, if the property is not found, it not create the related entities.

passing here seems legit to me

tests/Functional/ORMModelFactoryTest.php
```diff
@@ -622,6 +622,26 @@ public function can_find_or_create_entity_with_enum(): void
...
+    {
+        $category = CategoryFactory::createOne([
+            'posts' => PostFactory::new(['virtualTitle' => 'regular post'])->many(2),
+            'additionalPosts' => PostFactory::new(['virtualTitle' => 'additional post'])->many(2),
+            'addPost' => PostFactory::new(['virtualTitle' => 'regular post added']),
+        ]);
+
+        self::assertCount(5, $category->getPosts());
+
+        self::assertSame('additional post', $category->getPosts()[0]->getTitle());
```
nikophil: there is a small glitch here: `posts` will be hydrated with post persist callback, then they will be added at the last position in the collection

mpiot:
I've try to add a `dd()` in the adder method, before the PR (at least in 1.29.0) the `dd()` is triggered, but with this PR (and in 1.30.0) the `dd()` is not triggered. Apparently the property-accessor is not used or at least the adder is not called.

In my case, it fail, because the adder call a setter to define the relation:
```php
    public function addAdditionalDocument(ChangeRequestDocument $additionalDocument): self
    {
        if (true === $additionalDocument->isInitial()) {
            throw new \Exception('The addAdditionalDocument must receive a ChangeRequestDocument with initial as false.');
        }

        if (!$this->documents->contains($additionalDocument)) {
            $this->documents->add($additionalDocument);
            $additionalDocument->setChangeRequest($this);
        }

        return $this;
    }
```
If the adder is not called, that not execute the `$additionalDocument->setChangeRequest($this);` in my case, because of that, the ChangeRequestDocument$changeRequest is null.

nikophil:
we have the exact same behavior in the tests: [a setter is called in the adder](https://github.com/zenstruck/foundry/blob/1.x/tests/Fixtures/Entity/Category.php#L60)

in you [code sample](https://github.com/zenstruck/foundry/pull/457#issuecomment-1522891962) you use `additionalDocument` (singular) but in the issue it is `additionalDocuments` (plural), which one are you actually using? I guess it is the plural form (it should be the plural form).

Would you mind to make a public reproducer to your problem? 

We need to know why the adder is not called in your case: digging into what's happening in the property access component ([around here](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/PropertyAccess/PropertyAccessor.php#L525)) will help us
You can reach me on Symfony's slack if you want

mpiot:
@nikophil I've just created a simple example: https://github.com/mpiot/test-foundry

It contains:
 - 2 entities and repositories
 - 2 factories
 - 1 data fixture
 - use sqlite db

If I use Foundry 1.29, it wotks, 1.30 fail because not found relation, PR fail because the 'change_request_id' is null. I hope it could help :)

nikophil:
Super nice, thanks, I'll check later this week

mpiot:
> Super nice, thanks, I'll check later this week

I thank you for taking the time to look at this issue :)

nikophil:
ok, I understand better now the difference between your case and the tests: `ChangeRequestDocument::$changeRequest` is a mandatory field whereas `Post::$category` is not, if I change this, I got the same error.

we need to improve `Factory::getRelationshipMetadata()` in order to better grasp the relationship involved. But I really don't like how it was done prior to 1.30.0: in this case, in the [old code](https://github.com/zenstruck/foundry/blob/v1.29.0/src/Factory.php#L483), we were looping over the fields of the "parent" class and take the first one which has the right class, which could lead to weird behaviors...

nikophil:
Hi!

@mpiot could you please have a look to this new version? I've tested it with your reproducer and it seems to work.

I must say, I don't really like this feature, or at least how it is implemented: I had no other choice to reproduce what was done prior to v1.30: if we don't find the relationship, we loop over all relationships, until we find one where types match. I tried to use `symfony/property-info` but it does not work. I still really think if this used to work, it was a fluke...

Moreover it will only work with cascade persist relationships because in the other case, we don't call the adders (this problem was already existing prior to v1.30)

mpiot:
@nikophil it works for me.

But, if the way to do is really "dirty" and hinders the maintainability of the bundle, I can check if I find a way to do it on the v1.30+. I think this problem impact 2-3 entities on 220+ in my project.

nikophil:
> But, if the way to do is really "dirty" and hinders the maintainability of the bundle, I can check if I find a way to do it on the v1.30+. I think this problem impact 2-3 entities on 220+ in my project.

cool, we were actually talking yesterday with @kbond about not merging this PR...
or at least, not merging the "dirty" part (ie: the new method `Factory::fallbackOnFirstMatchingField()` I've introduced)

mpiot:
For the moment I do not find a way to do... I'm looking about it.
Because in the app, the only way to populate the $documents for exemple is to use adders, and the $documents property is private.

Edit: Try to use default getter and adder (getDocuments, addDocument, removeDocument): but dangerous in code, no dev must use it, only there for Fixtures and test, and cause a problem when need to merge 2 Factory::new()->as...->many(), because they return `FactoryCollection` and not `array`. 

Edit 2: Try to use `->instantiateWith(function (array $attributes) {` to create empty object, then loop over the attributes and use the property accessor to define them. If we match one of specific attributes, set the field we want  (later, can use Reflection to avoid create public adder). But the attributes array contains the field name, and the exception is throw. 

How it works actually when define: `->instantiateWith((new Instantiator())->allowExtraAttributes(['foo', 'bar']))` for exemple ? Why they don't throw the following exception ? 

>  Field App\Entity\MyClass::$foo does not exist or is not a relationship field.
>  Field App\Entity\MyClass::$bar does not exist or is not a relationship field.

mpiot:
Re :) As shown with @nikophil 

I can use the tips bellow:
```php
        ->instantiateWith((new Instantiator())->allowExtraAttributes(['additionalDocuments']))
        ->afterInstantiate(function (ChangeRequest $changeRequest, array $attributes): void {
            foreach ($attributes['additionalDocuments'] as $additionalDocument) {
                $changeRequest->addAdditionalDocument($additionalDocument);
            }
        })
```

The problem actually are: 
 - the extra `additionalDocuments` is catch by the exception that not found the relation (even if annoted as extra)
 - the extra documents are persisted before the `ChangeRequest`, but add `->witthoutPersisting()` on the `Document` solve the issue

nikophil:
Hi!

I think this PR is finally OK.

I've added a bunch of tests which validates different scenarios:
- possibility to add relationships OneToMany/ManytoOne through extra attributes and before/after instantiate, which permits the usage of "virtual" relationships properties, as needed by @mpiot 
- possibility to use a method as a property. However this behavior is not documented, since we don't want to encourage it.

@mpiot could you validate the PR is OK for you?
@kbond a review would ba welcomed :pray:

mpiot:
@nikophil I'll try the PR Tuesday morning 🙂

If you prefer avoid the usage of a method in place of the property, in my case that's not a problem. Because, we can do that by using: 
 - extra property
 - after instantiate event 

If needed, I can do a little PR to help user with this kind of usage ? 

Maybe, that can avoid other BC ? I don't really know 🤷‍♂️ 

Thanks a lot for your help 😀

nikophil:
> If you prefer avoid the usage of a method in place of the property, in my case that's not a problem

I know but it was my original intention in this PR

> If needed, I can do a little PR to help user with this kind of usage ?
> 
> Maybe, that can avoid other BC ? I don't really know man_shrugging

Not sure what you mean exactly?

mpiot:
> > Maybe, that can avoid other BC ? I don't really know man_shrugging
> 
> 
> 
> Not sure what you mean exactly?

Documentation :)
Explain, how to deal with other adder or setter (my case and maybe all other case where we want use a specific setter ?)

Depend if you prefer keep the way you implement: can call setter but without documentation ? 

Or explain, we can do it via extra parameter then call the function we want ?

nikophil:
Actually we'd prefer not documenting the methods as attributes.

For the other case, the callbacks are already documented, but maybe this can be improved!

kbond:
What I'd like is to keep the docs around `$attributes` purely for object properties. Ideally, I'd like to enforce this...

It does get into a bit of a gray area with property-access and adders but by and large, I'd really like to make it clear that attributes are only for properties.

mpiot:
@nikophil I've test the PR on my project and it's OK :)

nikophil:
nice!

ok for merging this PR @kbond ?

kbond (approved):
👍🏻

PR #458, opened 2023-04-28
feat: drop all connections before dropping db for postgres

### LINKED ISSUE #455: PostgreSQL and ResetDatabase trait
Since im working with a new project which is working with a postgreSQL Database i have issues with the ResetDatabase trait. This is maybe a problem which needs to be fixed on DB level or on the depending libary from `doctrine:database:drop` in AbstractSchemaResetter.php

The error
```
RuntimeException: Error running "doctrine:database:drop": Could not drop database "test" for connection named default                    
   ├ An exception occurred while executing a query: SQLSTATE[55006]: Object in use: 7 ERROR:  database "test" is being accessed by other users
   ├ DETAIL:  There is 1 other session using the database.
``` 

This happens in difference cases for me. 
1. When i run the tests while i have open and connected my DB Tool.
2. More worst: The project is using https://github.com/liip/LiipTestFixturesBundle/blob/2.x/doc/database.md and in some tests it uses the https://github.com/liip/LiipTestFixturesBundle/blob/2.x/src/Services/DatabaseTools/ORMDatabaseTool.php This leads to the same error. Im unsure why its a different user. Both "resetter" use the entityManager default connection (needs some more debugging)

Looking for a solution :)

1. Dunno something like this has to be added to doctrine/dbal?
https://www.postgresql.org/docs/current/sql-dropdatabase.html
https://stackoverflow.com/questions/55747282/dropping-db-when-there-are-active-connections
Postgres 13 or higher
```sql
DROP DATABASE database_name WITH (FORCE);
```

nikophil:
fixes #455

we cannot test this, since we use mysql in docker images, I'll test it in a project of mine
EDIT: it does not work on my project, but I don't know why... `SELECT current_database()` gives me the good database name :man_shrugging: 
@benblub  would you mind to test it on a project of yours? :pray: maybe it's coming from my setup...


src/Test/ORMDatabaseResetter.php
```diff
@@ -103,11 +105,22 @@ private function dropAndResetDatabase(): void
         foreach ($this->connectionsToReset() as $connection) {
             $dropParams = ['--connection' => $connection, '--force' => true];
 
-            if ('sqlite' !== $this->registry->getConnection($connection)->getDatabasePlatform()->getName()) {
+            if (!$this->registry->getConnection($connection)->getDatabasePlatform() instanceof SqlitePlatform) {
```
nikophil, line 108: fixes a deprecation

src/Test/ORMDatabaseResetter.php
```diff
@@ -101,9 +103,21 @@ private function dropSchema(): void
...
+                        '--connection' => $connection,
+                        'sql' => 'SELECT pid, pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname = current_database() AND pid <> pg_backend_pid()',
+                    ]
+                );
+            }
+
             $dropParams = ['--connection' => $connection, '--force' => true];
 
-            if ('sqlite' !== $this->registry->getConnection($connection)->getDatabasePlatform()->getName()) {
+            if (!$this->registry->getConnection($connection)->getDatabasePlatform() instanceof SqlitePlatform) {
```
nikophil: fixes a deprecation

benblub:
> fixes #455
> 
> we cannot test this, since we use mysql in docker images, I'll test it in a project of mine EDIT: it does not work on my project, but I don't know why... `SELECT current_database()` gives me the good database name 🤷‍♂️ @benblub would you mind to test it on a project of yours? 🙏 maybe it's coming from my setup...

Yes I w'll test

src/Test/ORMDatabaseResetter.php
```diff
@@ -101,9 +103,21 @@ private function dropSchema(): void
     private function dropAndResetDatabase(): void
     {
         foreach ($this->connectionsToReset() as $connection) {
+            if ($this->registry->getConnection($connection)->getDatabasePlatform() instanceof PostgreSQLPlatform) {
```
benblub: working for me with the class `PostgreSQL100Platform`.

the class is deprecated but the newer class isn't working yet.
```
Deprecated:
This class will be merged with PostgreSQLPlatform in 4.0 because support for Postgres releases prior to 10.0 will be dropped.
```

Suggest: inplement with an OR clause ..

@nikophil

benblub (commented):
working for me with PostgreSQL100Platform.

the class is deprecated but the newer class isn't working yet.

Deprecated:
This class will be merged with PostgreSQLPlatform in 4.0 because support for Postgres releases prior to 10.0 will be dropped.
Suggest: inplement with an OR clause ..

@nikophil

src/Test/ORMDatabaseResetter.php
nikophil: hi @benblub 

thanks for having tested that!

> Suggest: inplement with an OR clause ..

not sure to understand what you mean?

benblub:
```php
if ($this->registry->getConnection($connection)->getDatabasePlatform() instanceof PostgreSQLPlatform || $this->registry->getConnection($connection)->getDatabasePlatform() instanceof PostgreSQL100Platform)
```

@nikophil

nikophil: I've not done this because `PostgreSQL100Platform` extends `PostgreSQLPlatform`

benblub: ok then just replace PostgreSQLPlatform with PostgreSQL100Platform. In my test PostgreSQL100Platform is working and PostgreSQLPlatform not.

benblub: Bildschirmfoto 2023-05-10 um 15 03 19

nikophil: very strang it does not enter in the second if since `PostgreSQL100Platform` extends `PostgreSQLPlatform` :thinking: 

actually even when I enter the `if` and the query is played, it seems it does not kill other connections and the drop still fails... I really don't know why

benblub: For me: if its enter the if, the other connection is killed.

benblub: Bildschirmfoto 2023-05-10 um 15 40 14


Bildschirmfoto 2023-05-10 um 15 41 02

nikophil: I don't understand... `$x instanceof ChildClass` and `$x instanceof ParentClass` should return true if `$x` is a `ChildClass` :thinking: 

if the version with the `OR` works for you, I'll use this! but I'd like to figure out why it is not working on my project before merging

benblub: ok i tested now a little bit more. 

instance of is working on all 3 classes like expected.. not sure this didn't entered the if last days 🤔 
PostgreSQL100Platform
PostgreSQL94Platform
PostgreSQLPlatform

**so killing the connection is working for me with all 3 classes**

From the postgreSQL docs
```
however only superusers can terminate superuser backends.
```

**on your test you are maybe not a superuser???**

So basically we need to figure out if it was possible to drop the other connections?
if I'm not `SELECT rolsuper FROM pg_roles WHERE rolname = current_user;` a superuser we do not need to try go further.

It's possible to check after if connections was dropped `SELECT COUNT(*) AS active_connections FROM pg_stat_activity WHERE state IS NOT NULL` or `SELECT COUNT(*) AS active_connections FROM pg_stat_activity WHERE state = 'active'`

also we can setup a timeout and pg_terminate_backend will then return a false.

> pg_terminate_backend ( pid integer, timeout bigint DEFAULT 0 ) → boolean
> Terminates the session whose backend process has the specified process ID. This is also allowed if the calling role is a member of the role whose backend is being terminated or the calling role has privileges of pg_signal_backend, however only superusers can terminate superuser backends.
> If timeout is not specified or zero, this function returns true whether the process actually terminates or not, indicating only that the sending of the signal was successful. If the timeout is specified (in milliseconds) and greater than zero, the function waits until the process is actually terminated or until the given time has passed. If the process is terminated, the function returns true. On timeout, a warning is emitted and false is returned.

https://www.postgresql.org/docs/15/functions-admin.html

@nikophil

benblub: If we add posgreSQL to docker we can write tests.. not sure its worth at all 🙈

kbond: @nikophil and I have talked about doing that. Since postgres is the default for Symfony, I think it makes sense.

nikophil: I'm gonna change the image from mysql to postgres in this PR ;)
I've found why the connexion was not killed: I was using a persistant connexion 🤦‍♂️

kbond: Maybe we should add the db platform to the matrix? I'm fine keeping it limited to 2 MySQL jobs: one with dama and one without. Just for a sanity check.

nikophil: Yes, that was planned 🙂

nikophil: I'm struggling with the CI :disappointed: 

I used to have this error (see [this job](https://github.com/zenstruck/foundry/actions/runs/4967955407/jobs/8890273755)) - it seems the parallel jobs in the matrix drop connection of each other
```
SQLSTATE[HY000]: General error: 7 FATAL:  terminating connection due to administrator command
server closed the connection unexpectedly
```

then I added a condition where I say "dont kill connections if in the CI"
but I still get [some errors](https://github.com/zenstruck/foundry/actions/runs/4968012115/jobs/8890366127) in cases without dama:
```
1) Zenstruck\Foundry\Tests\Functional\WithMigrationTest::it_generates_a_valid_schema
RuntimeException: Error running "doctrine:database:drop": Could not drop database "zenstruck_foundry" for connection named default
An exception occurred while executing a query: SQLSTATE[55006]: Object in use: 7 ERROR:  database "zenstruck_foundry" is being accessed by other users
DETAIL:  There is 1 other session using the database.
```

I would appreciate some ideas, otherwise, I think we won't use postgres in the CI :innocent:

kbond: Is this the same problem as I had in https://github.com/zenstruck/foundry/pull/73?

nikophil: don't know, the logs have expired

nikophil: don't know, the logs have expired

nikophil: I really don't understand what's going on: https://github.com/zenstruck/foundry/actions/runs/5000161439/jobs/8957237306

I changed the CI to have only ONE job with postgres, and the db is named `zenstruck_foundry_${{ matrix.use-dama }}_${{ matrix.orm-db }}`, ie: `zenstruck_foundry_1_postgres` so it is unique in the matrix...

what I've noticed is also that the CI with postgres works without dama... :thinking:

kbond: > what I've noticed is also that the CI with postgres works without dama...

Ok, that was exactly the issue in #73. Worked without dama but not with.

nikophil: I managed to make dama work with postgres in GH CI: we must add `?serverVersion=15` in the dsn

kbond: Sheesh, ok that matches the recipe

nikophil: ok then, the CI is green with postgres as the main db engine!

this was... painful :sweat_smile:

tests/Functional/WithMigrationTest.php
```diff
@@ -28,22 +28,19 @@ final class WithMigrationTest extends KernelTestCase
     /**
      * @test
      */
-    public function it_generates_a_valid_schema(): void
+    public function it_can_use_schema_reset_with_migration(): void
```
nikophil, line 31: I don't know why, but in some case, when this test class used to have 2 tests, the second one was failing. Hence why I merged them

kbond (approved):
good to go after the mysql jobs are added as discussed!

PR #461, opened 2023-05-10
feat: allow ModelFactory::findOrCreate() in unit tests

nikophil:
sometimes, in order to ease creation of multiple nested elements and prevent unique constraint errors, I need to add a bunch of `SomeFactory::findOrCreate()` but it breaks my tests where I'm not using `KernelTestCase`


kbond (commented):
I don't have a strong opinion here. I guess my only concern is with swallowing legit runtime exceptions (when using with KernelTestCase)

kbond (approved):

Issue #465, opened 2023-05-22
Impossible to create object when using different name for field in constructor vs property


norkunas:
For example I have this entity:
```php
class EntityWithConstructorUnionType {
  private EntityType1 $field1;
  private EntityType2 $field2;
  public function __construct(EntityType1|EntityType2 $object) {
    if ($object instanceof EntityType2) { $this->type2 = $object; $this->type1 = null; }
    else {  $this->type2 = null; $this->type1 = $object; }
  }
}
```
and three foundry factories for all these entities.
Now in the `EntityWithConstructorUnionType` factory i declare:
```php
    protected function getDefaults(): array
    {
        return [
            'object' => EntityType1Factory::new(),
        ];
    }
```
And an error is thrown:
```
RuntimeException: Field App\Entity\EntityWithConstructorUnionType::$object does not exist or is not a relationship field.

/var/www/html/vendor/zenstruck/foundry/src/Factory.php:453
/var/www/html/vendor/zenstruck/foundry/src/Factory.php:385
/var/www/html/vendor/zenstruck/foundry/src/Factory.php:111
/var/www/html/vendor/zenstruck/foundry/src/ModelFactory.php:86
```
If i rename from `object` to `type1` in `getDefaults()` then other error is thrown:
```
InvalidArgumentException: Missing constructor argument "object" for "App\Entity\EntityWithConstructorUnionType".

/var/www/html/vendor/zenstruck/foundry/src/Instantiator.php:260
/var/www/html/vendor/zenstruck/foundry/src/Instantiator.php:43
/var/www/html/vendor/zenstruck/foundry/src/Factory.php:134
/var/www/html/vendor/zenstruck/foundry/src/ModelFactory.php:86
```

nikophil:
hello,

I think what you should do is to configure the instantiator as following:
```php
// EntityWithConstructorUnionTypeFactory

    public function initialize(): self
    {
        return $this->instantiateWith(
            (new Instantiator())->alwaysForceProperties()->withoutConstructor()
        );
    }
```

this way, your constructor won't be called and you will be able to rely on the properties' name

norkunas:
But my point that the constructor logic should run,i dont want to duplicate it :)

nikophil:
If you set the `$fieldX` nullable by defaults, this logic is not needed anymore 😇

norkunas:
Lib should not force to change entities imho:)

nikophil:
[...]

PR #466, opened 2023-05-23
tests: can create object with fields name different from construct

nikophil:
add a test to validate #465


docker/Dockerfile
```diff
@@ -58,4 +56,7 @@ RUN addgroup --system --gid ${UID} ${USER} ; \
 
 USER ${USER}
 
+RUN composer global require --no-progress --no-scripts --no-plugins symfony/flex ; \
+    composer global config --no-plugins allow-plugins.symfony/flex true
```
kbond, line 60: I actually think this isn't required. Same with the flex config in composer.json (just having flex installed globally enables the SYMFONY_REQUIRE)

kbond: nm, this is required.

kbond (approved):

PR #475, opened 2023-06-29
feat(lazy): Add memoized LazyValue

### LINKED ISSUE #473: Feature request: Single Lazy Value
First of all, thanks for the great bundle! It's been a massive improvement to our DX over the last few months ♥️

I have an interesting use case with regards to my entity model which has multiple relations between entities which need to be setup for the entity to be "valid".

For example, say we have a `Project` which has both a list of `Tasks` and a list of `Users`. A `Task` has a single `User` as the owner, but that owner _must_ exist in the list of `Project::users` for that `Task`:

```php
<?php

class User {}

class Task
{
    public function __construct(public readonly Project $project, public readonly User $owner) {}
}

class Project
{
    /** @var User[] */
    private array $users = [];
    /** @var Task[] */
    private array $tasks = [];

    public function addUser(User $user): void
    {
        $this->users[] = $user;
    }

    /** @return User[] */
    public function getUsers(): array
    {
        return $this->users;
    }

    public function addTask(Task $task): void
    {
        $this->tasks[] = $task;
    }

    /** @return Task[] */
    public function getTasks(): array
    {
        return $this->tasks;
    }
}
```

At the moment, my `TaskFactory` looks like this:

```php
class TaskFactory extends ModelFactory
{
    {
        $owner = UserFactory::new();

        return [
            'project' => ProjectFactory::new(['users' => [$owner]]),
            'owner'   => $owner,
        ];
    }
    
    {
        return Task::class;
    }
}
```

However, the UserFactory will actually create two different instances of a User, when in fact, I need them to be the same instance.
There are currently two ways to work around this (that I know of):
1. Use `UserFactory::createOne()` in `getDefaults()`, however this is [_not recommended_](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#many-to-one)
2. Everytime this factory is used in tests, you must first create the project and the owner and pass them both into the factorry

That being said, I think there is a very simple feature we could add to enable this use case by building off the existing [LazyValue](https://github.com/zenstruck/foundry/blob/1.x/src/LazyValue.php).

```php
class SingleLazyValue extends LazyValue
{
    private mixed $value = null;
    
    public function __invoke(): mixed
    {
        if ($this->value === null) {
            $this->value = parent::__invoke();
        }
        
        return $this->value;
    }
}
```

Which the `TaskFactory` would use like this:

```diff
    protected function getDefaults(): array
    {
-        $owner = UserFactory::new();
+        $owner = new SingleLazyValue(fn () => UserFactory::createOne());

        return [
            'project' => ProjectFactory::new(['users' => [$owner]]),
            'owner'   => $owner,
        ];
    }
```

I'm more than happy to submit a PR for this if you agree it's valuable. I'd also like some feedback on this approach (it does require either making `LazyValue` non-final, or duplicating it). Or potentially there's another approach I haven't considered?

ndench:
As discussed, this PR introduces the ability to memoize a LazyValue, which enables the following use-case:

```php
class TaskFactory extends ModelFactory
{
    {
        $owner = LazyValue::memoize(fn () => UserFactory::new());

        return [
            // Same User instance will be both added to the Project and set as the Task owner
            'project' => ProjectFactory::new(['users' => [$owner]]),
            'owner'   => $owner,
        ];
    }
    
    {
        return Task::class;
    }
}
```

I'm not entirely happy with the docs change, but I don't know how to explain it better without breaking LazyValue out into it's own section.

Closes: #473

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,11 @@ $(PSALM_BIN): vendor bin/tools/psalm/composer.json bin/tools/psalm/composer.lock
 	@${DOCKER_PHP_WITHOUT_XDEBUG} /usr/bin/composer bin psalm install
 	@touch -c $@ bin/tools/psalm/composer.json bin/tools/psalm/composer.lock
 
+.PHONY: docs
+docs: vendor ### Generate documentation to docs/output
+	@$(MAKE) --no-print-directory docker-start-if-not-running
+	@${DOCKER_PHP} ./bin/build-docs
+
 .PHONY: database-generate-migration
 database-generate-migration: database-drop-schema ### Generate new migration based on mapping in Zenstruck\Foundry\Tests\Fixtures\Entity
 	@${DOCKER_PHP} vendor/bin/doctrine-migrations migrations:migrate --no-interaction --allow-no-migration # first, let's load into db existing migrations
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -271,29 +271,7 @@ should have. `Faker`_ is available to easily get random data:
 .. note::
 
     ``getDefaults()`` is called everytime a factory is instantiated (even if you don't end up
-    creating it). If you have a value for one of your attributes that has side effects (ie
-    creating a file or fetching a random existing entity from another factory), you can wrap
-    the value in a ``LazyValue``. This ensures the value is only calculated when/if it's
-    needed.
-
-    .. code-block:: php
-
-        use Zenstruck\Foundry\Attributes\LazyValue;
-        use function Zenstruck\Foundry\lazy;
-
-        // ...
-
-        protected function getDefaults(): array
-        {
-            return [
-                // Call CategoryFactory::random() everytime this factory is instantiated
-                'category' => new LazyValue(fn() => CategoryFactory::random()),
-                // Memoize the result of the callback, so the same Category wil be used everytime
-                'category' => LazyValue::memoize(fn() => CategoryFactory::random()),
-
-                'file' => lazy(fn() => create_temp_file()), // or use the lazy() helper function
-            ];
-        }
+    creating it). `Lazy values`_ allows you to ensure the value is only calculated when/if it's needed.
 
 Using your Factory
 ~~~~~~~~~~~~~~~~~~
@@ -780,7 +758,8 @@ The following assumes the ``Comment`` entity has a many-to-one relationship with
 
 .. tip::
 
-    It is also recommended that your ``ModelFactory::getDefaults()`` return a ``Factory`` and not the created entity:
+    It is also recommended that your ``ModelFactory::getDefaults()`` return a ``Factory`` and not the created entity.
+    However, you can use `Lazy values`_ if you need to create the entity in the ``getDefaults()`` method.
 
     .. code-block:: php
 
@@ -788,10 +767,15 @@ The following assumes the ``Comment`` entity has a many-to-one relationship with
         {
             return [
                 // RECOMMENDED
+                // The Post will only be created when the factory is instantiated
                 'post' => PostFactory::new(),
                 'post' => PostFactory::new()->published(),
+                // The callback will be called when the factory is instantiated, creating the Post
+                'post' => LazyValue::new(fn () => PostFactory::createOne()),
+                'post' => lazy(fn () => PostFactory::new()->published()->create()),
 
-                // NOT RECOMMENDED - will potentially result in extra unintended Posts
+                // NOT RECOMMENDED
+                // Will potentially result in extra unintended Posts (if you override the value during instantiation)
                 'post' => PostFactory::createOne(),
                 'post' => PostFactory::new()->published()->create(),
             ];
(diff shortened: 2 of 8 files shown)


ndench:
Hmm I'm not entirely sure what to do about the CI failures.

src/LazyValue.php
```diff
@@ -19,27 +19,47 @@ final class LazyValue
...
      * @param callable():mixed $factory
      */
-    public function __construct(callable $factory)
+    public function __construct(callable $factory, bool $memoize = false)
     {
         $this->factory = $factory;
+        $this->memoize = $memoize;
+    }
+
+    public static function memoize(callable $factory): self
```
ndench, line 46: Should we add a named constructor `LazyValue::new` for non-memoized? Would make it consistent, but we can't make the contstructor private for BC.

tests/Unit/FunctionsTest.php
```diff
@@ -44,6 +47,29 @@ public function faker(): void
...
+     * @test
+     */
+    public function memoize(): void
+    {
+        $callback = new LazyValue(fn () => new \stdClass());
+        $value = memoize($callback);
+
+        $this->assertInstanceOf(LazyValue::class, $value);
+        $this->assertSame($value(), $value());
+    }
```
ndench: These duplicate the LazyValueTest. But I feel like they're necessary to ensure the functions are wired in correctly.

src/LazyValue.php
```diff
@@ -19,27 +19,47 @@ final class LazyValue
     /** @var callable():mixed */
     private $factory;
 
+    /** @var bool */
+    private $memoize;
+
+    /** @var mixed */
+    private $memoizedValue = null;
```
nikophil:
```suggestion
    private bool $memoize;
    private mixed $memoizedValue = null;
```

src/LazyValue.php
nikophil: I think this is a good idea, maybe we could add  a depercation layer for using `new Lazy()` directly?
see https://github.com/zenstruck/foundry/blob/1.x/src/FactoryCollection.php#L41-L43 as an example

tests/Unit/FunctionsTest.php
nikophil: that's fine IMO

nikophil (changes requested):
nice! :ok_hand: 

> Hmm I'm not entirely sure what to do about the CI failures.

I've fixed it here: https://github.com/zenstruck/foundry/pull/476
once it will be merged, you can rebase `1.x` branch

> I'm not entirely happy with the docs change, but I don't know how to explain it better without breaking LazyValue out into it's own section.

I'd say it's time for `LazyValue` to have its own section. Maybe we can keep the `.. note::` about the lazy, but shorten it and link to the new section?

ndench:
Thanks for the feedback @nikophil. I'll make those changes.

ndench:
@nikophil I've made those those changes. Would love some feedback on my LazyValues documentation section. 

Also, I can't quite figure out how to generate the docs to check the formatting.

docs/index.rst
```diff
@@ -854,6 +841,53 @@ The following assumes the ``Post`` entity has a many-to-many relationship with `
     // Example 5: create 3 Posts each with between 0 and 3 unique Tags
     PostFactory::createMany(3, ['tags' => TagFactory::new()->many(0, 3)]);
 
+Lazy values
+~~~~~~~~~~~
+
+The ``getDefaults()`` method is called everytime a factory is instantiated (even if you don't end up
+creating it). Sometimes, you might not want your value calculated every time. For example, ff you have a value for one
```
```suggestion
creating it). Sometimes, you might not want your value calculated every time. For example, if you have a value for one
```

kbond (commented):
Looks good to me, just a small typo.

> Also, I can't quite figure out how to generate the docs to check the formatting.

You should be able to run `bin/build-docs`, then open `docs/output/index.html`.

ndench:
Thanks @kbond. I've updated the README and the Makefile for generating the docs (hope that's ok!).

Here's some screenshots of the changes I made to the docs:

## The NOTE underneath **Model factories -> generate**
### Old

### New


## The TIP underneath **Doctrine relationships -> Many-to-one**
### Old

### New


## The new **Lazy values** section

kbond (approved):

kbond:
@nikophil, I think this is ready to go but can you check the makefile stuff?

nikophil:
Yeah sorry, I'm currently in holidays ⛱️
I might reconnect from next monday 😁

src/LazyValue.php
```diff
@@ -19,27 +19,56 @@ final class LazyValue
...
+        $this->memoize = $memoize;
+
+        if (!$calledInternally) {
+            trigger_deprecation('zenstruck/foundry', '1.34.0', "directly using LazyValue's constructor is deprecated. It will be private in v2. Use named constructors instead.");
+        }
+    }
+
+    public static function new(callable $factory): self
+    {
+        return new self($factory, false);
```
nikophil: shouldn't we call `self($factory, calledInternally: true)` ?

and `self($factory, memoize: true, calledInternally: true)` in `memoize()`

otherwise, we will trigger our own deprecations

tests/Unit/LazyValueTest.php
nikophil, line 1: I think some tests in this class will trigger deprecations. Either we must use named constructors, or tag these tests with `@group legacy`

Makefile
```diff
@@ -96,6 +96,11 @@ $(PSALM_BIN): vendor bin/tools/psalm/composer.json bin/tools/psalm/composer.lock
 	@${DOCKER_PHP_WITHOUT_XDEBUG} /usr/bin/composer bin psalm install
 	@touch -c $@ bin/tools/psalm/composer.json bin/tools/psalm/composer.lock
 
+.PHONY: docs
+docs: vendor ### Generate documentation to docs/output
```
nikophil, line 100: :+1:

nikophil (changes requested):
Hi!

the docs enhancement looks very good to me.

I think a small fix must be done in order to not auto-trigger our own deprecation errors

BTW, you can rebase `1.x` and the tests will be OK

src/LazyValue.php
ndench: Oops, good catch!

tests/Unit/LazyValueTest.php
ndench: Good point. I'll update them all to use named constructors and add a single legacy test to ensure the constructor still works as expected.

ndench:
Thanks @nikophil. I've rebased and resolved all the internal deprecations. I also added `FactoryTest::lazyMemoized_values_are_only_calculated_once` since there was a LazyValue test in there too.

nikophil (approved):

nikophil:
thanks @ndench

nikophil:
@ndench I've released this PR in v1.34.0

you'll can soon admire your docs update on symfony.com :wink:

ndench:
Thanks @nikophil! Really appreciate your time (and @kbond too)!

PR #476, opened 2023-06-30
fix: maker with cs fixer

nikophil:
since some times, maker bundle is shipped with cs fixer, so it fixes the cs of created files. This feature cannot be deactivated, so I updated the expected file to match the code style

this feature appeared in maker 1.49.0 so I decided to bump this version (dev req) in order to have the same behavior even with `--prefer-lowest` option

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
 
     services:
       mysql:
-        image: mysql:5.7
+        image: mysql:5.7.42
         env:
           MYSQL_ROOT_PASSWORD: 1234
         ports:
@@ -62,7 +62,7 @@ jobs:
           - 27017:27017
 
     env:
-      MYSQL_URL: mysql://root:1234@127.0.0.1:3306/zenstruck_foundry?serverVersion=5.7
+      MYSQL_URL: mysql://root:1234@127.0.0.1:3306/zenstruck_foundry?serverVersion=5.7.42
       PGSQL_URL: postgresql://postgres:1234@127.0.0.1:5432/zenstruck_foundry_${{ matrix.use-dama }}_${{ matrix.orm-db }}?charset=utf8&serverVersion=15
       MONGO_URL: mongodb://127.0.0.1:27017/dbName?compressors=disabled&amp;gssapiServiceName=mongodb
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 ARG PHP_VERSION
 
-FROM php:${PHP_VERSION}-cli
+FROM php:${PHP_VERSION}-cli-bullseye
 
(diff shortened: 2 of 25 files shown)


docker/Dockerfile
```diff
@@ -49,7 +49,7 @@ ARG UID=1001
 ARG XDEBUG_HOST="172.17.0.1"
 
 RUN addgroup --system --gid ${UID} ${USER} ; \
-    adduser --system --disabled-password --uid ${UID} --ingroup ${USER} ${USER} ; \
+    adduser --system --home /home/${USER} --disabled-password --uid ${UID} --ingroup ${USER} ${USER} ; \
```
nikophil, line 52: this is needed now the image uses the `bookworm`  version

.github/workflows/ci.yml
```diff
@@ -62,7 +62,7 @@ jobs:
           - 27017:27017
 
     env:
-      MYSQL_URL: mysql://root:1234@127.0.0.1:3306/zenstruck_foundry?serverVersion=5.7
+      MYSQL_URL: mysql://root:1234@127.0.0.1:3306/zenstruck_foundry?serverVersion=5.7.42
```
nikophil, line 65: this fixes a new deprecation in doctrine
https://github.com/doctrine/dbal/blob/3.6.x/src/Driver/AbstractMySQLDriver.php#L58-L63

docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG PHP_VERSION
 
-FROM php:${PHP_VERSION}-cli
+FROM php:${PHP_VERSION}-cli-bullseye
```
nikophil, line 3: at work we had a lot of subtle problems coming from the new `bookworm` release of debian

foundry's CI was failing as well for unknown reasons, so I preferred to stick to bullseye for now

tests/Functional/Bundle/Maker/MakeFactoryTest.php
```diff
@@ -11,6 +11,8 @@
 
 namespace Zenstruck\Foundry\Tests\Functional\Bundle\Maker;
 
+use Composer\InstalledVersions;
+use Composer\Semver\VersionParser;
```
ndench: Are these required? They don't appear to be used.

ndench (approved):
Looks great!

docker/Dockerfile
kbond: > this is needed now the image uses the bookworm version

Below you switched away from bookworm?

tests/Functional/Bundle/Maker/MakeFactoryTest.php
nikophil: indeed

docker/Dockerfile
nikophil: yes, but let's keep it, it will be there if we upgrade at some point

PR #486, opened 2023-08-04
minor: set Factory::isPersisting() protected

nikophil:
Hi!

a factory can't tell if it is persisting, which in some case could be helpful!

EDIT: @kbond do you have any idea why we have this erro in CI?
it's coming from https://github.com/doctrine/orm/pull/10785 and occurs only without dama, when global test state is flushed

something strange:
it does not occur when I call `make test -- --stop-on-failure --filter FactoryDoctrineCascadeTest`
but it occurs with the full test suite, whereas I suspect it occurs for the first test which uses doctrine
:thinking:

### DIFF
--- a/src/ZenstruckFoundryBundle.php
+++ b/src/ZenstruckFoundryBundle.php
@@ -28,7 +28,7 @@ final class ZenstruckFoundryBundle extends Bundle
 {
     public function boot(): void
     {
-        if (!Factory::isBooted()) {
+        if (!Factory::isBooted() && $this->container) {
             Factory::boot($this->container->get('.zenstruck_foundry.configuration'));
         }
     }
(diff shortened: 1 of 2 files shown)


src/ZenstruckFoundryBundle.php
```diff
@@ -28,7 +28,7 @@ final class ZenstruckFoundryBundle extends Bundle
 {
     public function boot(): void
     {
-        if (!Factory::isBooted()) {
+        if (!Factory::isBooted() && $this->container) {
```
nikophil, line 31: fixes a minor test in sca due to this change
https://github.com/symfony/symfony/commit/7a8457df3a2507fa4e254622847b429d5c67d733

kbond (approved):

PR #488, opened 2023-08-06
feat: disable persist globally

### LINKED ISSUE #471: Optionally disable persisting globally
Foundry is a great bundle! We had a custom builder pattern we were using before, but are in the process of moving everything to Foundry.

We have some use cases where we want nothing to persist, but there currently doesn't seem to be a way to globally disable persisting (without putting `->withoutPersisting()` in our AbstractFactory). 

Our use cases are:
1. Contract tests
	- Extends KernelTestCase to access services in the container (eg. the serializer)
	- Create some models and serialize them to ensure they meet the contract
	- No need to persist entities here
2. Previewing email notifications without sending them
	- Using faker to generate fake data to instantiate different notification classes
	- Render the email notification in the browser in dev
	- Allows us to quickly iterate email design without having to trigger the email notification 

Our "solution" to item 1 has been to call the internal `TestState::bootFoundryForUnitTests()` method in the `setUp()` method of our contract tests. This creates Foundry without an entity manager, so the entities aren't persisted even though we're technically in a KernelTestCase.

However, we don't have a solution for item 2. It would be great to be able to use `TestState::configure()` or `TestState::disablePersist()` at runtime before we instantiate any factories. 

Has anyone else had similar use cases? Do you think this would be a valuable addition to the library?

nikophil:
fixes https://github.com/zenstruck/foundry/issues/471

ping @ndench 

CI will be gren as soon as https://github.com/zenstruck/foundry/pull/489will be merged

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -15,6 +15,7 @@ use Doctrine\ORM\EntityRepository;
 use Doctrine\Persistence\ManagerRegistry;
 use Doctrine\Persistence\ObjectManager;
 use Faker;
+use Zenstruck\Foundry\Exception\CannotGetRepositoryException;
 use Zenstruck\Foundry\Exception\FoundryBootException;
 
 /**
@@ -40,6 +41,8 @@ final class Configuration
     private bool $flushEnabled = true;
 
     private bool $databaseResetEnabled = true;
+
+    private bool $persistEnabled = true;
 
     /**
      * @param string[] $ormConnectionsToReset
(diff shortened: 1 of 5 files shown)


kbond (approved):

ndench:
Nice work @nikophil! Unfortunately I'm on leave at the moment and haven't had a chance to test it out. But looking forward to giving it a try :tada:

nikophil:
released in [1.35.0](https://github.com/zenstruck/foundry/releases/tag/v1.35.0)

PR #489, opened 2023-08-06
fix: do not flush global state twice when dama is not enabled

### LINKED ISSUE #487: fix CI
https://github.com/zenstruck/foundry/actions/runs/5762507081

nikophil:
fixes #487

not sure why this happened now (maybe because of the new exception in doctrine/orm) but without dama, the first test using `ResetDatabase` was flushing twice the global state, hence the error

### DIFF
--- a/src/Test/DatabaseResetter.php
+++ b/src/Test/DatabaseResetter.php
@@ -37,7 +37,7 @@ final class DatabaseResetter
         return \class_exists(StaticDriver::class) && StaticDriver::isKeepStaticConnections();
     }
 
-    public static function resetDatabase(KernelInterface $kernel): void
+    public static function resetDatabase(KernelInterface $kernel, bool $damaIsEnabled): void
     {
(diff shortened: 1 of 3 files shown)


src/Test/DatabaseResetter.php
```diff
@@ -47,7 +47,7 @@ public static function resetDatabase(KernelInterface $kernel): void
 
         $databaseResetter->resetDatabase();
 
-        self::bootFoundry($kernel);
+        self::bootFoundry($kernel, flushGlobalState: $damaIsEnabled);
```
nikophil, line 50: we cannot directly use `self::isDAMADoctrineTestBundleEnabled()` because it checks if `StaticDriver::isKeepStaticConnections()` and at this point [it is disabled](https://github.com/zenstruck/foundry/blob/1.x/src/Test/ResetDatabase.php#L40) even if dama is used

kbond (approved):
Awesome, thanks for figuring this out!

PR #492, opened 2023-08-11
fix: replace doctrine:query:sql with dbal:run-sql

### LINKED ISSUE #491: doctrine:query:sql is deprecated
Looks like `doctrine:query:sql` used in the `ORMDatabseResetter` is deprecated since 2.2.

Since the foundry depends on > 2.5 I'd assume it's safe to replace

```Since doctrine/doctrine-bundle 2.2: The "Doctrine\Bundle\DoctrineBundle\Command\Proxy\RunSqlDoctrineCommand" (doctrine:query:sql) is deprecated, use dbal:run-sql command instead.```

KDederichs:
Fixes #491 by switching to the non deprecated command


kbond (approved):

kbond:
Thanks @KDederichs!

PR #494, opened 2023-08-11
feat(make:factory): use ObjectFactory for not persisted objects

nikophil:
here is the last blocker for merging https://github.com/zenstruck/foundry/pull/462  :smiley:


kbond (approved):
wooo!

PR #496, opened 2023-08-26
fix(split-factories): some fixes

nikophil:
PHPStan fixes:
- using `$this` inside factory was problematic with phpstan
- using `self::findOrCreate()` or any static method was also problematic
- phpstan did not understood factoryCollection->create() would return a list<>
- auto-detect for a `FactoryCollection` whether if the target class has persistence or not 

bug fixes:
- factory parameters are not invokable
- we must apply "before instantiate" callbacks before normalizing step
- `@ORM\Entity` and `@ODM\Document` annotations should be checked along with attributes to check if a class is persistable

### DIFF
--- a/phpunit-dama-doctrine.xml.dist
+++ b/phpunit-dama-doctrine.xml.dist
@@ -11,7 +11,11 @@
         <ini name="error_reporting" value="-1"/>
         <env name="KERNEL_CLASS" value="Zenstruck\Foundry\Tests\Fixtures\Kernel"/>
         <env name="USE_DAMA_DOCTRINE_TEST_BUNDLE" value="1"/>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0&amp;max[direct]=0&amp;quiet[]=indirect&amp;quiet[]=other"/>
+        <!--
+            max[self]=1 represents "The Zenstruck\Foundry\BaseFactory::new() method is considered final" deprecation.
+            It could be removed in 2.0
+        -->
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=1&amp;max[direct]=0&amp;quiet[]=indirect&amp;quiet[]=other"/>
         <env name="SHELL_VERBOSITY" value="0"/>
     </php>
 
--- a/src/BaseFactory.php
+++ b/src/BaseFactory.php
@@ -51,8 +51,9 @@ abstract class BaseFactory
 
     /**
      * @param Attributes|string $attributes
+     * @final
      */
-    final public static function new(array|callable|string $attributes = [], string ...$states): static
+    public static function new(array|callable|string $attributes = [], string ...$states): static
     {
         if (\is_string($attributes) || $states) {
             trigger_deprecation('zenstruck/foundry', '1.36', 'Passing states as strings is deprecated and this behavior will be removed in 2.0.', self::class, self::class);
@@ -242,10 +243,6 @@ abstract class BaseFactory
      */
     protected function normalizeAttribute(mixed $value, string $name): mixed
     {
-        if (\is_callable($value)) {
-            $value = $value();
-        }
-
         if ($value instanceof FactoryCollection) {
             $value = $value->all();
         }
@@ -265,9 +262,10 @@ abstract class BaseFactory
     }
 
     /**
+     * @internal
      * @param Attributes $attributes
      */
-    final protected function mergedAttributes(array|callable $attributes): array
+    protected function mergedAttributes(array|callable $attributes): array
     {
         return \array_merge(
             ...[
--- a/src/Object/ObjectFactory.php
+++ b/src/Object/ObjectFactory.php
@@ -112,14 +112,6 @@ abstract class ObjectFactory extends BaseFactory
         // execute "lazy" values
         $parameters = LazyValue::normalizeArray($parameters);
 
-        foreach ($this->beforeInstantiate as $hook) {
-            $parameters = $hook($parameters, static::class());
-
-            if (!\is_array($parameters)) {
-                throw new \LogicException('Before Instantiate event callback must return an array.');
-            }
-        }
-
         $object = ($this->instantiator ?? self::factoryManager()->defaultObjectInstantiator())($parameters, static::class());
 
         foreach ($this->afterInstantiate as $hook) {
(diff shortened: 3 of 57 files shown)


src/BaseFactory.php
```diff
@@ -51,8 +51,9 @@ public function __call(string $name, array $arguments): array
 
     /**
      * @param Attributes|string $attributes
+     * @final
      */
-    final public static function new(array|callable|string $attributes = [], string ...$states): static
+    public static function new(array|callable|string $attributes = [], string ...$states): static
     {
```
nikophil, line 57: this change is made to deprecate the usage of `PersistentObjectFactory` with "non-persistable" classes

tests/Fixtures/Factories/PostFactory.php
```diff
@@ -2,13 +2,33 @@
...
+ * @method static Post|Proxy                     find(object|array|mixed $criteria)
+ * @method static Post|Proxy                     findOrCreate(array $attributes)
+ * @method static Post|Proxy                     first(string $sortedField = 'id')
+ * @method static Post|Proxy                     last(string $sortedField = 'id')
+ * @method static Post|Proxy                     random(array $attributes = [])
+ * @method static Post|Proxy                     randomOrCreate(array $attributes = []))
+ * @method static Post[]|Proxy[]                 all()
+ * @method static Post[]|Proxy[]                 findBy(array $attributes)
+ * @method static Post[]|Proxy[]                 randomSet(int $number, array $attributes = []))
+ * @method static Post[]|Proxy[]                 randomRange(int $min, int $max, array $attributes = []))
```
nikophil, line 29: I added those, because some were creating problems with phpstan, now I'm also analyzing this factory with `make sca`

tests/allowed-deprecations.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "location": "Zenstruck\\Foundry\\Persistence\\PersistentObjectFactory",
+        "message": "The \"Zenstruck\\Foundry\\BaseFactory::new()\" method is considered final. It may change without further notice as of its next major version. You should not extend it from \"Zenstruck\\Foundry\\Persistence\\PersistentObjectFactory\".",
```
nikophil: this deprecation occurs in all tests, that's why I've added to the "allowed" list

phpunit-dama-doctrine.xml.dist
```diff
@@ -11,7 +11,11 @@
         <ini name="error_reporting" value="-1"/>
         <env name="KERNEL_CLASS" value="Zenstruck\Foundry\Tests\Fixtures\Kernel"/>
         <env name="USE_DAMA_DOCTRINE_TEST_BUNDLE" value="1"/>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0&amp;max[direct]=0&amp;quiet[]=indirect&amp;quiet[]=other"/>
+        <!--
+            max[self]=1 represents "The Zenstruck\Foundry\BaseFactory::new() method is considered final" deprecation.
+            It could be removed in 2.0
+        -->
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=1&amp;max[direct]=0&amp;quiet[]=indirect&amp;quiet[]=other"/>
```
nikophil, line 18: I tried to use the baseline or ignore files features of the bridge, but it still fails with sf 5.4

src/Object/ObjectFactory.php
```diff
@@ -128,4 +120,24 @@ final protected function instantiate(array $parameters): array
...
     }
+
+    /**
+     * @param Attributes $attributes
+     *
+     * @internal
+     */
+    final protected function mergedAttributes(array|callable $attributes): array
+    {
+        $attributes = parent::mergedAttributes($attributes);
```
nikophil, line 131: I'm not found of this kind of thing...

I'm afraid we'd enter in an "inheritance hell" at some point (`UserlandFactory` extends `PersistenceObjectFactory` extends `ObjectFactory` extends `BaseFactory`)
I'm really wondering if we cannot find another design where we'd use composition over inheritance, but I cannot find a nice pattern :shrug:

src/Object/ObjectFactory.php
```diff
@@ -112,14 +112,6 @@ final protected function instantiate(array $parameters): array
         // execute "lazy" values
         $parameters = LazyValue::normalizeArray($parameters);
 
-        foreach ($this->beforeInstantiate as $hook) {
-            $parameters = $hook($parameters, static::class());
-
-            if (!\is_array($parameters)) {
-                throw new \LogicException('Before Instantiate event callback must return an array.');
-            }
-        }
```
nikophil, line 121: we must apply before instantiate callbacks before normalization occurs

src/BaseFactory.php
```diff
@@ -265,9 +262,10 @@ protected function normalizeAttribute(mixed $value, string $name): mixed
     }
 
     /**
+     * @internal
      * @param Attributes $attributes
      */
-    final protected function mergedAttributes(array|callable $attributes): array
+    protected function mergedAttributes(array|callable $attributes): array
```
nikophil, line 268: this change is used to override `mergedAttributes` in `ObjectFactory`, in order to apply "before instantiate" callbacks

src/Persistence/PersistenceManager.php
```diff
@@ -117,6 +122,29 @@ public function isPersistEnabled(): bool
...
+        if (!$attributes) {
+            $attributes = self::annotationReader()?->getClassAnnotation($reflectionClass, ORMEntityAttribute::class);
+        }
+
+        if (!$attributes) {
+            $attributes = self::annotationReader()?->getClassAnnotation($reflectionClass, ODMDocumentAttribute::class);
+        }
+
+        return (bool) $attributes;
+    }
```
nikophil, line 146: I know a better way to do this would have been to check doctrine's mapping, and check whether type is set as attribute or annotation, but this way I can check it statically

src/Persistence/PersistentObjectFactory.php
```diff
@@ -301,10 +310,6 @@ final protected function normalizeAttribute(mixed $value, string $name): mixed
             return $value->isPersisted() ? $value->refresh()->object() : $value->object();
         }
 
-        if (\is_callable($value)) {
-            $value = $value();
-        }
```
nikophil, line 306: This was coming from a misunderstanding from my side... I've always thought we were allowed to do something like this:
```php
class SomeFactory extends ModelFactory
{
    {
		return [
			'field' => fn() => FieldFacotry::findOrCreate()
        ];
	}
}
```
but it actually introduce a bug if the value passed is invokable (I had the case in one of the project where I tested this version)

tests/Fixtures/Factories/PostFactory.php
```diff
@@ -22,6 +42,24 @@ public static function class(): string
...
+    }
+
+    /**
+     * @return FactoryCollection<Proxy<Post>>
+     * This method is here to test an edge case with the phpstan extension
+     */
+    public function methodUsingThis(): FactoryCollection
+    {
+        return $this->sequence([]);
+    }
```
nikophil, line 61: this was easier to put these method here but I can create a factory specific for static analysis

kbond (approved):
Wow, nice work here!

PR #498, opened 2023-08-28
minor: remove `faker()`/`defaultObjectInstantiator()` from `FactoryManager`

kbond:


kbond:
I'll fix this up tomorrow.

kbond:
@nikophil, any idea why the phpstan is failing on that line now?

nikophil:
this is so strange oO
"go home, PHPStan, you'r drunk"

nikophil:
hum I don't know :thinking: 
I'll have a closer look later if you don't find

kbond:
Ignoring for now - it's clearly invalid. I think, previously, the `iterable $factories` property wasn't type-hinted so phpstan didn't complain.

PR #504, opened 2023-09-12
docs: Move note about immutability into States section

pavelmaca:
Mentioning immutability after calling `$this->addState()` would save some troubles for newcomers.


kbond (approved):
Thanks @pavelmaca!

PR #506, opened 2023-09-14
feat: allow stories to call their own pool

### LINKED ISSUE #505: Question: Access pool from it's own story::build()
Hi,

In a Story's build method, I start by creating a set of objects and add them in a pool.

Later in that story I want to get a random object from that pool.
I find that this gets stuck in an infinite loop until PHP probably exists with memory-limits or stack-limits.

I think from quickly clicking through the code, that if I do `self::getRandom('poolname')` from within a Storys build method(), eventually Foundry tries to access the pool by doing `static::load()->pools['poolname']`. But I guess that `load()` thinks the Story is not loaded yet, and will call my Story's build() method again, which tries to access the pool, which acess my Story's build() method again, ....

Is it possible to get a random object from your own pool? If not, what would be a good method then?
use `faker()->randomElement($objects)` to access a random one?

nikophil:
fixes #505

### DIFF
--- a/tests/Fixtures/Stories/StoryWhichReadsItsOwnPool.php
+++ b/tests/Fixtures/Stories/StoryWhichReadsItsOwnPool.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Zenstruck\Foundry\Tests\Fixtures\Stories;
+
+use Zenstruck\Foundry\Story;
+use Zenstruck\Foundry\Tests\Fixtures\Factories\CategoryFactory;
+
+final class StoryWhichReadsItsOwnPool extends Story
+{
+    public function build(): void
+    {
+        $this->addState('php', CategoryFactory::new()->create(['name' => 'php']), 'some-pool');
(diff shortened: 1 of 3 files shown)


tests/Fixtures/Stories/StoryWhichReadsItsOwnPool.php
```diff
@@ -0,0 +1,21 @@
...
+
+final class StoryWhichReadsItsOwnPool extends Story
+{
+    public function build(): void
+    {
+        $this->addState('php', CategoryFactory::new()->create(['name' => 'php']), 'some-pool');
+        $this->addState('symfony', CategoryFactory::new()->create(['name' => 'symfony']), 'some-pool');
+
+        // story can access its own state
+        $this->addState('php2', self::get('php'), 'some-pool');
```
kbond:
```suggestion
        $this->addState('php2', self::getState('php'), 'some-pool');
```

Should fix the self-deprecation.

nikophil: fixed

PR #509, opened 2023-10-02
minor: allow Symfony 7.0

kbond:
This just allows foundry to be installed on Symfony 7.0. We still need to wait for some upstream dev library's to support 7.0 before we can add to the test matrix.


OskarStark:
I added a reminder in
* https://github.com/zenstruck/foundry/issues/510

PR #513, opened 2023-11-02
chore: remove Makefile

nikophil:


.env
```diff
@@ -0,0 +1,5 @@
+DATABASE_URL="mysql://root:1234@127.0.0.1:3307/foundry_test?serverVersion=5.7.42"
+#DATABASE_URL="postgresql://zenstruck:zenstruck@127.0.0.1:5433/zenstruck_foundry?serverVersion=15"
```
kbond:
```suggestion
```

kbond (changes requested):
Love it! Can you update the readme? After merging, we can remove our custom docker image, correct?

kbond:
Some thoughts:
- The non-dama tests we run on every permutation that includes a database is really slow... wondering how we can speed this up, only run on a single permutation? Maybe the code coverage job?
- We're no longer testing with migrations, this would need the auto-generate migrations before the test system I have in foundry-next
- Should we run the "no bundle" tests?

.env
nikophil: shouldn't we keep it? it shows how to switch to a postgres database

nikophil: Ok, I've added an example in the README

README.md
```diff
@@ -30,89 +30,85 @@ Want to watch a screencast 🎥 about it? Check out https://symfonycasts.com/fou
...
+# .env.local
+DATABASE_URL="postgresql://zenstruck:zenstruck@127.0.0.1:5433/zenstruck_foundry?serverVersion=15"
+
+# run test suite with postgreSQL
+$ vendor/bin/simple-phpunit
 ```
 
-### Using xdebug with PhpStorm
+The `.env.local` file can also be used to override the port of the database containers,
+if it does not meet your local requirements. You'll also need to override docker compose configuration:
```
nikophil, line 75: I exactly had this problem: another project is using 3307, so I decided to put this in the docs. This is simple and could be useful

config/cli-config.php
nikophil, line 1: this file is needed by any doctrine binary. It can be used for generating migrations, I'm in favor to keep it: it's not really usual to need new migrations in Foundry, but it can be useful

nikophil:
> The non-dama tests we run on every permutation that includes a database is really slow... wondering how we can speed this up, only run on a single permutation? Maybe the code coverage job?

ok, this is a simple solution to not mess with a more complex matrix, and even test one permutation with dama. Done ✅

> We're no longer testing with migrations, this would need the auto-generate migrations before the test system I have in foundry-next

migrations are tested in [`ORMDatabaseResetterTest`](https://github.com/zenstruck/foundry/blob/1.x/tests/Functional/ORMDatabaseResetterTest.php) and [`WithMigrationTest`](https://github.com/zenstruck/foundry/blob/1.x/tests/Functional/WithMigrationTest.php)

By the way, I was surprised that in `foundry-next`, you were running two whole test suites on all permutation in order to test migration, is it really needed? :thinking: 

The reset mode does not change the whole behavior of foundry, I don't think it deserves all this CI time!?

.env
kbond: > shouldn't we keep it? it shows how to switch to a postgres database

Oh, my bad, then we should make it the dsn that works out of the box with the docker image.

config/cli-config.php
kbond: What about https://github.com/kbond/foundry-next/blob/2.x/bin/console instead of `cli-config.php`? I've found it useful for viewing/dumping the bundle config w/o having to load this bundle into a project.

nikophil: ok I will do this

.env
nikophil: ok I'll do this

README.md
```diff
@@ -30,89 +30,89 @@ Want to watch a screencast 🎥 about it? Check out https://symfonycasts.com/fou
...
-USE_ORM=1
-USE_ODM=1
-USE_DAMA_DOCTRINE_TEST_BUNDLE=1
-SYMFONY_REQUIRE=5.4.* # allowed values: 5.4.* | 6.0.* | 6.1.* | 6.2.*
-PHP_VERSION=8.0 # allowed values: 8.0 | 8.1 | 8.2
-PREFER_LOWEST=1 # force composer to request lowest dependencies
-```
+```bash
+# start the container
+$ docker compose up --build --detach
```
kbond: I've never used the `--build` and `--detach` options. What is their purpose and is it required?

composer.json
```diff
@@ -65,6 +67,22 @@
...
         }
     },
+    "scripts": {
+        "test": [
+            "@test-no-dama",
+            "@test-dama",
+            "@test-no-foundry-bundle"
+        ],
+        "test-no-dama": "vendor/bin/simple-phpunit",
+        "test-dama": "vendor/bin/simple-phpunit -c phpunit.dama.xml.dist",
```
kbond:
```suggestion
        "test-no-dama": "vendor/bin/phpunit",
        "test-dama": "vendor/bin/phpunit -c phpunit.dama.xml.dist",
```

?

README.md
```diff
@@ -30,89 +30,89 @@ Want to watch a screencast 🎥 about it? Check out https://symfonycasts.com/fou
...
 
-You can also add these variables to the `.env` file to change the ports used by docker:
-```dotenv
-PGSQL_PORT=5434
-MONGO_PORT=27018
+# run test suite with all available permutations
+$ composer test
+
+# run only one permutation
+$ vendor/bin/simple-phpunit
```
kbond:
```suggestion
$ vendor/bin/phpunit
```

README.md
```diff
@@ -30,89 +30,89 @@ Want to watch a screencast 🎥 about it? Check out https://symfonycasts.com/fou
...
+
+# run test suite with dama/doctrine-test-bundle
+$ vendor/bin/simple-phpunit -c phpunit.dama.xml.dist
```
kbond:
```suggestion
$ vendor/bin/phpunit -c phpunit.dama.xml.dist
```

README.md
```diff
@@ -30,89 +30,89 @@ Want to watch a screencast 🎥 about it? Check out https://symfonycasts.com/fou
...
+
+# run test suite with postgreSQL instead of MySQL
+$ DATABASE_URL="postgresql://zenstruck:zenstruck@127.0.0.1:5433/zenstruck_foundry?serverVersion=15" vendor/bin/simple-phpunit
```
kbond:
```suggestion
$ DATABASE_URL="postgresql://zenstruck:zenstruck@127.0.0.1:5433/zenstruck_foundry?serverVersion=15" vendor/bin/phpunit
```

README.md
```diff
@@ -30,89 +30,89 @@ Want to watch a screencast 🎥 about it? Check out https://symfonycasts.com/fou
...
 
-You can execute any command into the php container using docker compose:
-```shell
-$ docker-compose exec php [your commmand] # or "docker compose" depending on your compose version
+```bash
+# .env.local
+DATABASE_URL="postgresql://zenstruck:zenstruck@127.0.0.1:5433/zenstruck_foundry?serverVersion=15"
+
+# run test suite with postgreSQL
+$ vendor/bin/simple-phpunit
```
kbond:
```suggestion
$ vendor/bin/phpunit
```

README.md
kbond: Could we use a more obscure port as the default?

.env
nikophil: actually I've added in the README how to switch to postgres, so I've removed the comment here

README.md
nikophil: I don't know... using default port + 1 is kind of a convention... because it is possible to override everything, I think we can leave all these ports like this

README.md
nikophil: `--detach` is the same as `-d` if you don't add it, docker runs in the terminal (and not detached mode) so you have to open a new terminal. I'd rather keep it, it is the way docker should be used in my opinion

about the `--build` I think it's not needed anymore, and docker builds automatically (it was not the case long time ago, and I kept this habits :sweat_smile:)

nikophil:
Hey!

I think this PR is almost ready (minus your future comments :sweat_smile:)

### About the migrations
I think having ONE test (ie: [`WithMigrationTest`](https://github.com/zenstruck/foundry/blob/1.x/tests/Functional/WithMigrationTest.php)) which uses `ResetDatabase` with migrations should be OK. One other test is also directly testing the reset database behavior: [`ORMDatabaseResetterTest`](https://github.com/nikophil/foundry/blob/chore/remove-makefile/tests/Functional/ORMDatabaseResetterTest.php)

I've used your technique which uses `bin/console` and which generates automatically the migrations, I think it is pretty cool.
In order not to generate a migration each time tests are started locally, I've conditioned it with an env var `TEST_MIGRATIONS` which I've enabled in the CI The migration works as well with mysql or postgres.   

### About Dama
I re-added it in the matrix, I think it is important to run the tests without dama at least once with postgre and once with mysql. Moreover [`ORMDatabaseResetterTest` cannot run with dama](https://github.com/nikophil/foundry/blob/chore/remove-makefile/tests/Functional/ORMDatabaseResetterTest.php#L27-L29) and we want to test this.
I've also used dama in the test coverage job in order to speed it up a little bit

I think code coverage fails because we're computing it with Foundry's bundle deactivated :shrug:

kbond:
I'm good to go here. 

> I think code coverage fails because we're computing it with Foundry's bundle deactivated 🤷

Should the coverage job run the test suite 3 times (capturing/uploading a separate coverage.xml for each):
1. dama
2. no-dama
3. no-bundle

kbond (approved):

PR #514, opened 2023-11-07
fix: exclude phpcsconfig from sync-with-template job

nikophil:


kbond:
As discussed, not needed.

PR #517, opened 2023-11-08
1.x bc layer

nikophil:
- refactor: deprecate passing callable to sequence
- refactor: deprecate usage of disable/enable_persisting without doctrine
- refactor: deprecate TestState for UnitTestConfig

### DIFF
--- a/src/Test/TestState.php
+++ b/src/Test/TestState.php
@@ -37,51 +37,51 @@ final class TestState
     private static array $globalStates = [];
 
     /**
-     * @deprecated Use TestState::configure()
+     * @deprecated Use UnitTestConfig::configure()
      */
     public static function setInstantiator(callable $instantiator): void
     {
-        trigger_deprecation('zenstruck\foundry', '1.23', 'Usage of TestState::setInstantiator() is deprecated. Please use TestState::configure().');
+        trigger_deprecation('zenstruck/foundry', '1.37.0', sprintf('Method "%s()" is deprecated and will be removed in Foundry 2.0. Use "%s::configure()" instead.', __METHOD__, UnitTestConfig::class));
 
         self::$instantiator = $instantiator;
     }
 
     /**
-     * @deprecated Use TestState::configure()
+     * @deprecated Use UnitTestConfig::configure()
      */
     public static function setFaker(Faker\Generator $faker): void
     {
-        trigger_deprecation('zenstruck\foundry', '1.23', 'Usage of TestState::setFaker() is deprecated. Please use TestState::configure().');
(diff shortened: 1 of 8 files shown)


src/Test/TestState.php
```diff
@@ -37,51 +37,51 @@ final class TestState
...
     }
 
     /**
-     * @deprecated Use bundle configuration
+     * @deprecated
      */
     public static function enableDefaultProxyAutoRefresh(): void
     {
-        trigger_deprecation('zenstruck\foundry', '1.23', 'Usage of TestState::enableDefaultProxyAutoRefresh() is deprecated. Please use bundle configuration under "auto_refresh_proxies" key.');
+        trigger_deprecation('zenstruck/foundry', '1.37.0', sprintf('Method "%s()" is deprecated and will be removed in Foundry 2.0 without replacement.', __METHOD__));
```
nikophil, line 64: @kbond could you please confirm this is true, before I add BC layer for the bundle configuration

src/Test/UnitTestConfig.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Faker;
+use Zenstruck\Foundry\Instantiator;
+
+final class UnitTestConfig
+{
+    public static function configure(?Instantiator $instantiator = null, ?Faker\Generator $faker = null): void
+    {
+        TestState::configureInternal($instantiator, $faker);
+    }
```
nikophil, line 15: it was really easier to create this dumb function `TestState::configureInternal()` which will be removed directly in Foundry 2.0 than to move everything from `TestState` here

src/Test/TestState.php
kbond: Yes, in 2.0 you won't be able to use bundle-less. When using in a unit test, you can't use persistence.

nikophil: I was actually wondering if there is no replacement for these "auto refresh" stuff? even in the bundle config? meaning conf `zenstruck_foundry.auto_refresh_proxies` is gone

kbond: My idea for 2.x, since we'll have `PersistentObjectFactory`, which does not use proxies and `PersistentProxyObjectFactory`, which does, that you'd opt into auto-refreshing when using the `PersistentProxyObjectFactory`. What would be the point of using a proxy if you're not wanting to auto-refresh?

kbond: I'm not sure how best to create a BC layer for this though...

nikophil: I'm actually wondering if we should not introduce now both `PersistentProxyObjectFactory` and `PersistentObjectFactory`

this way, we could have a smooth migration for `final` objects

PR #518, opened 2023-11-10
refactor: introduce PersistentObjectFactory and deprecate PersistentProxyObjectFactory with final

nikophil:


src/Factory.php
```diff
@@ -87,11 +88,23 @@ public function __call(string $name, array $arguments): array
...
+    final public function create(
+        array|callable $attributes = [],
+        /**
+         * @deprecated
+         * @internal
+         */
+        bool $noProxy = false
+    ): object
     {
+        if (\count(func_get_args()) === 2 && !str_starts_with(debug_backtrace(options: \DEBUG_BACKTRACE_IGNORE_ARGS, limit: 2)[1]['class'] ?? '', 'Zenstruck\Foundry')) {
```
kbond: Cool solution!

src/Factory.php
```diff
@@ -483,4 +497,10 @@ private function withCascadePersist(): static
 
         return $cloned;
     }
+
+    /** @interal */
```
kbond:
```suggestion
    /**
     * @internal
     */
```

src/Persistence/PersistentProxyObjectFactory.php
```diff
@@ -14,22 +14,28 @@
...
+abstract class PersistentProxyObjectFactory extends PersistentObjectFactory
 {
-    public function __construct()
+    final public static function new(array|callable|string $defaultAttributes = [], string ...$states): static
     {
-        parent::__construct(static::class(), calledInternally: true);
+        if ((new \ReflectionClass(static::class()))->isFinal()) {
+            trigger_deprecation(
+                'zenstruck\foundry', '1.37.0',
+                \sprintf('Using a proxy factory with a final class is deprecated and will throw an error in Foundry 2.0. Use "%s" instead, or unfinalize "%s" class.', PersistentProxyObjectFactory::class, static::class())
```
kbond: nice call here!

src/Factory.php
nikophil: In one of the previous BC PR, I added few parameters called `calledInternally`, I think I'll drop them in favor of this mechanism

psalm.xml
```diff
@@ -12,4 +12,9 @@
         <file name="tests/Fixtures/Maker/expected/can_create_factory_for_entity_with_repository_with_data_set_psalm.php" />
         <file name="tests/Fixtures/Maker/expected/can_create_factory_with_static_analysis_annotations_with_data_set_psalm.php" />
     </projectFiles>
+
+    <issueHandlers>
+        <ImplementedReturnTypeMismatch errorLevel="suppress" />
+        <MethodSignatureMismatch errorLevel="suppress" />
+    </issueHandlers>
```
nikophil, line 19: this is not ideal, but... psalm analysis passes :innocent:

src/Factory.php
```diff
@@ -87,11 +88,23 @@ public function __call(string $name, array $arguments): array
...
+     * @psalm-suppress ImplementedReturnTypeMismatch
      */
-    final public function create(array|callable $attributes = []): Proxy
+    final public function create(
+        array|callable $attributes = [],
+        /**
+         * @deprecated
+         * @internal
+         */
+        bool $noProxy = false
```
nikophil, line 101: we could prevent adding this parameter, and create a method:
```php
public function withProxy(): bool
{
    return $this instanceof PersistentObjectProxyFactory::class;
}
```
but this way, we cannot use the phpstan trick `@phpstan-return ($noProxy is true ? TObject: Proxy<TObject>)`
and it breaks the type system in multiple places :shrug: 

So I'd prefer to use this technique which is far from perfect but which seems acceptable for  a BC layer, WDYT?

phpstan-baseline.neon
```diff
@@ -1,5 +1,10 @@
 parameters:
 	ignoreErrors:
+		-
+			message: "#^Should not use node with type \"Expr_Eval\", please change the code\\.$#"
+			count: 1
+			path: src/AnonymousFactoryGenerator.php
+
```
nikophil, line 7: this error comes from `eval()` in `AnonymousFactoryGenerator`

kbond (approved):

PR #520, opened 2023-11-14
Document using `paratestphp/paratest`

### LINKED ISSUE #103: Add performance section about brianium/paratest

kbond:
Closes #103


kbond:
@benr77, as you just implemented in your project, do you mind giving this a quick review to ensure I captured everything?

benr77:
Looks good to me. The only thing I can think of to add, is that Global State should be minimised where possible, as this adds to the test startup time - not a worry when you are running your entire suite, but if you are just running a single test case or class then you have to wait for the global state to load first.

kbond:
Thanks, I added a warning about this to the global state section.

PR #523, opened 2023-11-18
refactor: deprecate config database_resetter

nikophil:
- refactor: deprecate config database_resetter.orm in favor of orm.reset
- refactor: deprecate config database_resetter.odm in favor of mongo.reset


kbond (approved):
❤️

PR #524, opened 2023-11-18
refactor: deprecate config auto_refresh_proxies

nikophil:
ok, this gave me serious headaches! 😅 :exploding_head:  but the PR is finally green :green_circle: 

we had some small incoherence in the model at different places, which made the test fail if the entities are refreshed from the db 

at first, I wanted to deprecate the option `auto_refresh_proxies` and default to true, but this could lead to :
- people who still don't define the option may have some trouble
- as well as people who specified it to `false` in the config: they would remove the option because of the deprecation, and the behavior will change implicitly.

What I propose is to make this explicit: let's force the user through deprecation to set this option to `true`, so they won't be surprised that some of their test fail. And let's deprecate the option in 2.0, where we won't use it anymore.
WDYT?


kbond (approved):
🎉

PR #529, opened 2023-11-29
refactor: remove not working BC layer on Factory::__construct()

nikophil:


kbond (approved):
To confirm, we're going to include an empty constructor in 2.0?

nikophil:
It's kinda weird, but I think we need to. It's actually really hard (impossible?) to detect that `__construct()` was called from a `parent::__construct()` in user land

kbond:
Yep not a big deal to leave it. I think in 2.x, we could trigger a deprecation if the user calls parent::__construct() (then remove in 3.x)

PR #541, opened 2023-12-21
refactor: better deprecations

nikophil:
This PRs fixes few things:
- remove `deprecations.php` file
- follow https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf where it seems to work (as he says, it's a "best effort" thing, there is no perfect solution)
- remove all `sprintf()` used along with `trigger_deprecation()` I think this will decrease the amount of deprecation is user land by a lot


src/Persistence/Proxy.php
```diff
@@ -14,6 +14,8 @@
 /**
  * @template TProxiedObject of object
  *
+ * @mixin TProxiedObject
```
nikophil, line 17: if all type hints are modified from `Zenstruck\Foundry\Proxy` to  `Zenstruck\Foundry\Persistence\Proxy` in user land, we must add this `@mixin` as well as all other deprecated methods. Otherwise SCA reports tons of errors

UPGRADE-2.0.md
```diff
@@ -0,0 +1,164 @@
...
+The main change is that we've introduced a separation between "object" factories,
+"persistence" factories and "persistence with proxy" factories.
+
+When Foundry 1.x was "persistence first", Foundry 2 is "object first".
+This would allow more decoupling from the persistence layer.
+
+## How to
+
+Every modification needed for a 1.x to 2.0 migration is covered by a deprecation.
+All you have to do is to upgrade to the last 1.x version, activate the deprecation helper, make the tests run, and
```
kbond:
```suggestion
All you have to do is to upgrade to the latest 1.x version, activate the deprecation helper, make the tests run, and
```

While I really hope we never touch 1.x again after 2.0 is released, that might not be realistic.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,164 @@
...
+```shell
+SYMFONY_DEPRECATIONS_HELPER="max[self]=0&amp;max[direct]=0&amp;quiet[]=indirect&amp;quiet[]=other"
+```
+
+Here is the full list of modifications needed:
+
+## Factory
+
+- `withAttributes()` and `addState()` are both deprecated in favor of `with()`
+- `sequence()` and `createSequence()` do not accept `callable` as a parameter anymore (how do we leverage `__index` instead??)
```
kbond: > (how do we leverage `__index` instead??)

This is something I changed in 2.x, correct? I don't remember looking closely at this deprecation layer, did you duplicate it?

UPGRADE-2.0.md
```diff
@@ -0,0 +1,164 @@
...
+public static function class(): string
+{
+    // ...
+}
+```
+
+As a rule of thumb to help you to choose between these two new factory parent classes:
+- only entities (ORM) or documents (ODM) should use `PersistentProxyObjectFactory`
+- you should only use `PersistentProxyObjectFactory` if you want to leverage "auto refresh" behavior
+- all "non-persistable" objects could use `PersistentObjectFactory`, since `Proxy` is only useful for entities/documents
```
kbond: This is a little weird - in 2.0, we'll have `ObjectFactory` for this and since this is an 1.x-2.0 upgrade guide it should show the new base class.

Couple ideas:
1. add `ObjectFactory` to 1.x (I am kind of thinking we should do this as in 2.0, you won't be able to use non-persistent objects with `PersistentObjectFactory` so would be a hard-BC break, right?)
2. rename this document `UPGRADE-1.?.md`, then add an `UPGRADE-2.0.md` in 2.x

kbond (commented):
Just reviewed the first bit of the upgrade doc for now. I'll get to the rest later.

UPGRADE-2.0.md
nikophil: It's still only in 2.x => https://github.com/kbond/foundry-next/blob/6741d5b4acbe39b2ade6064a746a91941f2d049e/src/FactoryCollection.php#L84

I think it is made to be able to have sequences with something like "title 1", "title 2", etc...

UPGRADE-2.0.md
nikophil: yeah, I'm indeed in favor to introduce `ObjectFactory` in 1.x :smile:

nikophil: https://github.com/zenstruck/foundry/pull/543

UPGRADE-2.0.md
nikophil: I've removed the reference to `__index`

UPGRADE-2.0.md
nikophil: I've re-worded this section

UPGRADE-2.0.md
nikophil, line 1: I have still added nothing about rector, I'm waiting to know how we will ship the rector rules

PR #543, opened 2023-12-27
feat: introduce ObjectFactory

nikophil:


src/Object/ObjectFactory.php
```diff
@@ -0,0 +1,166 @@
...
+/*
+ * This file is part of the zenstruck/foundry package.
+ *
+ * (c) Kevin Bond <kevinbond@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Zenstruck\Foundry\Object;
```
kbond: In foundry-next, I have this in `Zenstruck\Foundry`. I don't have a hard opinion on this I just didn't like the duplicated _Object_ word in the namespace and this will always be part of foundry core (once we make the split packages)

kbond (commented):
Should we document this at all?

src/Object/ObjectFactory.php
nikophil: fixed ✅

nikophil:
> Should we document this at all?

I think before merging `1.x-bc` we should upgrade the docs, in order to document the different factories available, and update all the documented api

Issue #540, opened 2023-12-15
[Feature] Add a event before/after reset database (and schema creation)


lsv:
As im working with postgresql, with the postgis extension - This means that you need to enable postgis as an extension on the database.

This works fine, but in my tests, I ofcourse use `ResetDatabase`, now this drops the database, and therefor it also "deletes" the extension.

So would it be possible to add an event after database creation, but before schema import?

The only way I can do this is by using migrate, but that tripled the time on my test suite.

nikophil:
Hi @lsv 

I think you can leverage [global state](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#global-state) here.

I have a similar problem: I need some sql views to be created in order to be used in the tests.

I'm using an invokable service to generate them:
```php
// /config/packages/zenstruck_foundry.php

$containerConfigurator->extension('zenstruck_foundry', [
    'global_state' => [
        ViewsGenerator::class,
    ],
]);
```
and it gets created before each test (actually, before the very first test of the suite, since I'm using dama)

lhapaipai:
Thank you @nikophil for your answer and the tip,
But I think the @lsv issue and the mine :smile: is that `createSchema` method will fail if we haven't enabled PostGIS extension.
The extension offers a set of functions and types that are used to generate our tables.
`global_state` is a bit too late.

```php
// vendor/zenstruck/foundry/src/Test/ORMDatabaseResetter.php
final class ORMDatabaseResetter extends AbstractSchemaResetter
{
    // ....

    public function resetDatabase(): void
    {
        $this->dropAndResetDatabase();

        // we need to make this sql query here 'CREATE EXTENSION postgis';
        // because our schema contain PostGIS functions
        
        $this->createSchema();
    }

    // ....
}
```

Dama couldn't help us either because `resetDatabase` seems to be called in the `beforeClassMethod`.

nikophil:
Hi @lhapaipai 

indeed, you're right!

Maybe you could enable `migrate` mode for the database resetter?
You can look at this issue: https://github.com/zenstruck/foundry/issues/477

This comes with a performance cost, which makes dama almost mandatory.

(it was exactly the same problem, and I gave exactly the same wrong answer :clown_face: )

lhapaipai:
Thank you @nikophil , I imagine this comes with performance cost, but at least it solves the problem !!

nikophil:
If you use dama, the performance cost is really limited, since the migrations are only ran once.

[...]

PR #552, opened 2024-01-31
feat: skip resetDatabase when migrations are up-to-date

haase-fabian:
I use the DamaDoctrineTestBundle in combination with `database_resetter.orm.reset_mode: migrate`.

Everytime the test suite is run, when there is a TestCase using the `ResetDatabase` trait, the database is reset.

This change skips the step when there are no new migrations and no unregistered migrations in the database.

This eliminates about 10s of initialization time to 0s for every test suit run in my case.

This assumes, that when DamaDoctrineTestBundle is installed, it is also used and correctly configured.

I use ddev for my projects, so I haven't locally tested the changes as I don't have php locally installed.

### DIFF
--- a/src/Test/DatabaseResetter.php
+++ b/src/Test/DatabaseResetter.php
@@ -26,20 +26,21 @@ use Zenstruck\Foundry\Factory;
 final class DatabaseResetter
 {
     private static bool $hasBeenReset = false;
+    private static bool $isDAMADoctrineTestBundleEnabled = false;
 
     public static function hasBeenReset(): bool
     {
         return self::$hasBeenReset;
     }
 
-    public static function isDAMADoctrineTestBundleAvailable(): bool
+    public static function isDAMADoctrineTestBundleEnabled(bool $cached = false): bool
     {
-        return \class_exists(StaticDriver::class);
-    }
+        if($cached) {
+            return self::$isDAMADoctrineTestBundleEnabled;
+        }
 
-    public static function isDAMADoctrineTestBundleEnabled(): bool
-    {
-        return \class_exists(StaticDriver::class) && StaticDriver::isKeepStaticConnections();
+        self::$isDAMADoctrineTestBundleEnabled = \class_exists(StaticDriver::class) && StaticDriver::isKeepStaticConnections();
+        return self::$isDAMADoctrineTestBundleEnabled;
     }
 
     public static function resetDatabase(KernelInterface $kernel, bool $damaIsEnabled): void
--- a/src/Test/DatabaseResetter.php
+++ b/src/Test/DatabaseResetter.php
@@ -32,6 +32,11 @@ final class DatabaseResetter
         return self::$hasBeenReset;
     }
 
+    public static function isDAMADoctrineTestBundleAvailable(): bool
+    {
+        return \class_exists(StaticDriver::class);
+    }
+
     public static function isDAMADoctrineTestBundleEnabled(): bool
     {
         return \class_exists(StaticDriver::class) && StaticDriver::isKeepStaticConnections();
--- a/src/Test/ORMDatabaseResetter.php
+++ b/src/Test/ORMDatabaseResetter.php
@@ -11,6 +11,7 @@
 
 namespace Zenstruck\Foundry\Test;
 
+use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
 use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\Persistence\ManagerRegistry;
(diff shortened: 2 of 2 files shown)


nikophil:
Hi @haase-fabian 

> Everytime the test suite is run, when there is a TestCase using the ResetDatabase trait, the database is reset.

This statement is not true when using dama

when we encounter a `TestCase` using `ResetDatabase` **for the first time**, we reset the database, but it is only made once per test suite, thanks to `ResetDatabase::shouldReset()` and the static property `DatabaseResetter::$hasBeenReset`.

Afterwards, if using dama, we never reset the schema, thanks to this condition:
https://github.com/zenstruck/foundry/blob/1.x/src/Test/ORMDatabaseResetter.php#L55-L58

Nonetheless, I do think your PR improves the perfs, but it also breaks the promise of the `ResetDatabase` which is to start from a fresh db. What if I have called `StaticDriver::commit();die;` somewhere in my tests for debugging purpose? The next time, the db won't be empty... I'm wondering how we can mitigate this?

> This assumes, that when DamaDoctrineTestBundle is installed, it is also used and correctly configured.

I think this is also problematic: dama is enabled from `phpunit.xml.dist` and could be disabled locally by creating a `phpunit.xml`. Actually, a coworker of mine disables it because he does not like using dama for some reasons. This PR would break the tests on his computer, because it just test the presence of `StaticDriver` class. I don't know if we can access some PHPUnit's internals, in order to know whether or not the extension is enabled :thinking:

haase-fabian:
Hi @nikophil,

>> Everytime the test suite is run, when there is a TestCase using the ResetDatabase trait, the database is reset.
>
> This statement is not true when using dama

I'm sorry for my ambiguous language. I was trying to say it's reset once if the suite encounters a TestCase with set DatabaseTrait.

I made the `isDAMADoctrineTestBundleEnabled` cachable.
This correctly identifies disabled DamaDoctrineTestBundles.

And I truncate all registered tables in case there are any leftover entries.
~~I do not know whether the Postgres version for disabling Foreign Key checks works correctly, however if it fails it will fall back to recreating the whole database. (same for other database platforms)~~

In my project, this takes about 0.8s now.

Thinking about introducing a flag to opt-in skipping the truncation part in a follow-up PR. 
```
    private function isResetUsingTruncate(): bool {
        if(isset($_SERVER['FOUNDRY_RESET_DETECTION'])) {
            return $_SERVER['FOUNDRY_RESET_DETECTION'] !== 'unsafe';
        }

        return true;
    }
```
Disabling the truncate results in performances around 0.08s.

nikophil:
Hey @haase-fabian 

this is just an idea, but do you think we could somehow leverage the proposal [here](https://github.com/zenstruck/foundry/issues/540)?

and manage what you want to achieve is userland based on an event system?
I feel like this would be the silver bullet solution, giving each user the flexibility s.he wants

haase-fabian:
Yes, I would definitely appreciate extensibility of any kind of bundle.
I don't understand why many bundles violate the [Open–closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle).

However, I also think, that bundles should provide reasonable default behavior to be as close to zero-configuration as possible.
Therefore, making the database reset in 'migration' mode as fast as possible without having to set up each new project.
At least for the most common use cases.

In regard to an event based system:
I don't know exactly the capabilities, but from a search in the documentation, I think it's lacking decorating around the method.
Therefore, I would opt for decorators rather than events.

A rough idea:
- 1. Make ORMDatabaseResetter etc. implement DatabaseResetterInterface
- 2. Use protected methods and make them non-final
- 3. Inject the DatabaseResetter implementation via the container
- 4. Utilize [Decorator stacks](https://symfony.com/doc/current/service_container/service_decoration.html#stacking-decorators)

The interface gets passed a context variable:
```php
interface DatabaseResetterInterface
{
    public function resetDatabase(array $context): void;
    public function resetSchema(array $context): void;
}
```

Then the stack itself could look similar to this:
```yaml
stack:
    - DamaDatabaseResetter::class //adds ["isDAMADoctrineTestBundleEnabled" => true|false] to $context
    - KernelProvider::class // adds ["kernel" => $kernel, ...$configuration], maybe also ["application" => $app]
    - ChainResetter::class // delegates to both ORMDatabaseResetter and ODMDatabaseResetter if available
        -
            - MigrationDatabaseResetter::class
            - ORMDatabaseResetter::class // holds actual reset logic
        - ODMDatabaseResetter::class // holds actual reset logic
```

And in user land, one can easily override specific behavior:
```php
#[When(env: 'test')]
#[AsDecorator(ORMDatabaseResetter::class)]
class ViewResetter implements DatabaseResetterInterface
{
    public function __construct(#[AutowireDecorated] private object $inner, private ManagerRegistry $registry, ...)
    {

    }
    public function resetDatabase(array $context): void {
        if($this->shouldReset()){
            //install postgres extensions
            $this->inner->resetDatabase($context);
            //add missing views
        }
    }
    ...
}
```

nikophil:
since https://github.com/zenstruck/foundry/pull/690, you can hook in the reset database mechanism, and do what you want

PR #554, opened 2024-02-02
docs(bc layer): document known BC breaks

nikophil:


UPGRADE-2.0.md
```diff
@@ -96,6 +96,34 @@ return static function (RectorConfig $rectorConfig): void {
...
+```diff
+$proxyPost = PostProxyFactory::createOne();
+- $proxyPost->setTitle();
+- $proxyPost->setBody(); // 💥
++$proxyPost->_withoutAutoRefresh(
++   function(Post $object) {
++       $proxyPost->setTitle();
++       $proxyPost->setBody();
++   }
++);
```
nikophil, line 123: actually the example that I'm using was also creating an error in 1.x, but it's easier to understand than the complex behavior which really triggers the error

UPGRADE-2.0.md
```diff
@@ -96,6 +96,34 @@ return static function (RectorConfig $rectorConfig): void {
 };
 ```
 
+## Known BC breaks
+
+The following error cannot be covered by Rector rules nor by the deprecation layer.
+```php
```
kbond: php?

UPGRADE-2.0.md
kbond: Agreed.

kbond (approved):

PR #556, opened 2024-02-05
feat: support `doctrine/orm` 3

kbond:
TODO:

- [ ] sqlite changes. We don't currently test but there are some breaking changes when using `ResetDatabase` with sqlite

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "doctrine/doctrine-migrations-bundle": "^2.2|^3.0",
         "doctrine/mongodb-odm": "^2.4",
         "doctrine/mongodb-odm-bundle": "^4.4.0|^5.0",
-        "doctrine/orm": "^2.11",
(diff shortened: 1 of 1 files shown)


nikophil:
won't it be a PITA to maintain tests for both versions?

kbond:
Possibly, but if we bump the min orm to latest 2.x, perhaps it won't be a problem?

kbond:
I think some of the work done in https://github.com/symfony/maker-bundle/pull/1439 could help us here.

jrushlow:
`MakeTestRunner::configureDatabase()` has the logic we needed to get sqlite to work with the new `orm` - https://github.com/symfony/maker-bundle/blob/d9574302ee67c23ae5c803556f2b779c3b0b94a2/src/Test/MakerTestRunner.php#L156

nikophil:
closed in favor of https://github.com/zenstruck/foundry/pull/569

PR #559, opened 2024-02-13
refactor(bc layer): deprecate passing states as string to Factory::new()

nikophil:


kbond (approved):
Yes, thank you. I've always regretted that I did that.

PR #562, opened 2024-02-17
use phpunit attributes

jrushlow:
Starting in PHPUnit 11 - the use of metadata in doc-comments is _hard_ deprecated. Results in tests passing as expected but PHPUnit deprecation warnings everywhere...

### DIFF
--- a/src/Test/Factories.php
+++ b/src/Test/Factories.php
@@ -11,6 +11,8 @@
 
 namespace Zenstruck\Foundry\Test;
 
+use PHPUnit\Framework\Attributes\After;
+use PHPUnit\Framework\Attributes\Before;
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
 use Zenstruck\Foundry\ChainManagerRegistry;
 use Zenstruck\Foundry\Exception\FoundryBootException;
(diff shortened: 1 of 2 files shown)


src/Test/Factories.php
```diff
@@ -57,6 +60,7 @@ public static function _setUpFactories(): void
      * @internal
      * @after
      */
+    #[After]
```
jrushlow, line 63: having both the attribute && the doc-comment `@after` does not trigger a deprecation in PHPUnit 11.

jrushlow:
Not sure if ya'll use any sort of `@legacy` comments like we do in MakerBundle. If so, let me know and I'll add them to the doc-comments. E.g. `@after - @legacy - can be removed when PHPUnit <10 support is dropped.`

nikophil:
Hi @jrushlow 

sorry to have been quite on this lately.

Because we only use phpunit `^9.5`, I don't think we'll need any `@group legacy` for now?

could rebase your PR, please?

jrushlow:
@nikophil rebase done

don't think the ci failure is related.. looks like a `php-cs-fixer` type of problem

nikophil:
absolutely, I'll fix this  today

nikophil:
thanks @jrushlow

PR #565, opened 2024-02-26
feat: foundry 2.0 🎉

nikophil:
I still need to polish some stuff (mainly some things in CI)

we also need to target a branch `2.x` in `zenstruck/foundry`


composer.json
```diff
@@ -9,85 +9,70 @@
         {
             "name": "Kevin Bond",
             "email": "kevinbond@gmail.com"
+        },
+        {
+            "name": "Nicolas PHILIPPE",
+            "email": "nikophil@gmail.com"
```
nikophil, line 15: :innocent:

kbond: +100!

PR #569, opened 2024-03-20
feat: support `doctrine/orm` 3

nikophil:
had to create a new PR for this 

I'm wondering if somehow we should add more CI permutations using `doctrine/orm:^2.0`?
I think there is only one permutation remaining which uses 2.x: the one with `--lowest`

### DIFF
--- a/src/Bundle/Maker/Factory/ODMDefaultPropertiesGuesser.php
+++ b/src/Bundle/Maker/Factory/ODMDefaultPropertiesGuesser.php
@@ -16,6 +16,8 @@ use Symfony\Component\Console\Style\SymfonyStyle;
 
 /**
  * @internal
+ *
+ * @phpstan-import-type AssociationFieldMapping from ODMClassMetadata
  */
 class ODMDefaultPropertiesGuesser extends AbstractDoctrineDefaultPropertiesGuesser
 {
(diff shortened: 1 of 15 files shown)


src/Bundle/Maker/Factory/ODMDefaultPropertiesGuesser.php
```diff
@@ -40,14 +40,14 @@ public function __invoke(SymfonyStyle $io, MakeFactoryData $makeFactoryData, Mak
                 continue;
             }
 
-            $fieldName = $item['fieldName'];
+            $fieldName = $item['fieldName']; // @phpstan-ignore-line
```
nikophil, line 43: this is kinda strange: `$metadata->associationMappings` documented as 
```php
@psalm-var array<string, AssociationFieldMapping>
```
which is a psalm pseudo type refering to an array:
```php
@psalm-type AssociationFieldMapping = array{ ... }
```

but phpstan outputs the following error:
```
Access to offset 'fieldName' on an unknown class Doctrine\ODM\MongoDB\Mapping\AssociationFieldMapping.
```

kbond (approved):

alexandre-daubois:
Thanks! 😊

PR #570, opened 2024-03-20
fix: only run maker test on php > 8.0

nikophil:
`symfony/maker-bundle` ships its own phpcs-fixer version. The one we need, which accepts our config is `v3.49.0` which is shiiped from [maker bundle 1.55.0](https://github.com/symfony/maker-bundle/releases/tag/v1.55.0)

but this version is incompatible with our lowest CI permutation.

This PR skips maker tests on CI with `--prefer-lowest` option and php 8.0

The problem will be fixed in Foundry 2.0, where we bump php 8.1


nikophil:
GHA is buggy right now: all the workflow have successfully ended, but it says that some are still in progress :shrug:

PR #571, opened 2024-03-22
rector: rewrite phpdoc

nikophil:
work in progress... but 90% OK!


utils/rector/RewriteFactoryPhpDoc
```diff
@@ -0,0 +1,3 @@
+PHPUnit 10.5.13 by Sebastian Bergmann and contributors.
+
+Required argument for option "--filter" is missing
```
jdreesen, line 3: It looks as if this was committed by mistake.

nikophil: wow nice spot, thanks :sweat_smile:

PR #580, opened 2024-03-30
[1.x] fix using stories with simple objects

### LINKED ISSUE #573: Using story pools without persisting
I'm using foundry without doctrine and manually save my objects with this function in my factories : 

```
    {
        return $this->withoutPersisting()
            ->afterInstantiate(
                fn (Establishment $establishment) => $this->repository->save($establishment),
            );
    }
``` 

I'm facing an issue while using stories and pools.  

If I call `$this->addToPool()` inside a story, I get the following error : 

```
RuntimeException : Should not get doctrine's object manager when persist is disabled.
```

What I understood is that I wouldn't be able to use pools because of the `normalizeObject` private method of `Story` : 

```
        // ensure proxies are persisted
        if (!$object->isPersisted()) {
            $object->save();
        }
```

Is there a workaround for this use case ? 

Thanks to this : https://github.com/zenstruck/foundry/pull/488 we can globally disable persist. Would it make sense to check for this configuration in the Story ?

jrushlow:
- fix using stories with objects that can't be persisted

Current behavior:
image

This is intended to be legacy code that shouldn't be merged / released in 2.x.

fixes: #573

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -59,6 +59,8 @@ class Factory
     /** @var callable[] */
     private array $afterPersist = [];
 
+    private bool $shouldPersist = true;
+
     /**
      * @param class-string<TObject> $class
      */
@@ -156,10 +158,10 @@ class Factory
         }
 
         if (!$this->isPersisting(calledInternally: true)) {
-            return $noProxy ? $object : new ProxyObject($object);
+            return $noProxy ? $object : new ProxyObject($object, $this->shouldPersist);
         }
 
-        $proxy = new ProxyObject($object);
+        $proxy = new ProxyObject($object, $this->shouldPersist);
 
         if ($this->cascadePersist && !$postPersistCallbacks) {
             return $proxy;
@@ -228,6 +230,7 @@ class Factory
 
         $cloned = clone $this;
         $cloned->persist = false;
+        $cloned->shouldPersist = false;
 
         return $cloned;
     }
--- a/src/Proxy.php
+++ b/src/Proxy.php
@@ -50,6 +50,8 @@ final class Proxy implements \Stringable, ProxyBase
     public function __construct(
         /** @param TProxiedObject $object */
         private object $object,
+
+        private bool $shouldPersist = true,
     ) {
         if ((new \ReflectionClass($object::class))->isFinal()) {
             trigger_deprecation(
@@ -174,6 +176,10 @@ final class Proxy implements \Stringable, ProxyBase
 
     public function _save(): static
     {
+        if (!$this->shouldPersist) {
+            return $this;
+        }
+
         $this->objectManager()->persist($this->object);
 
         if (Factory::configuration()->isFlushingEnabled()) {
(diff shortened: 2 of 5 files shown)


src/Factory.php
```diff
@@ -59,6 +59,8 @@ class Factory
     /** @var callable[] */
     private array $afterPersist = [];
 
+    private bool $shouldPersist = true;
```
nikophil, line 62: wasn't it possible to directly use `$this->persist`?

src/Proxy.php
```diff
@@ -50,6 +50,8 @@ final class Proxy implements \Stringable, ProxyBase
     public function __construct(
         /** @param TProxiedObject $object */
         private object $object,
+
+        private bool $shouldPersist = true,
```
nikophil, line 54: I think we should directly disable autorefrieshing when `shouldPersist` is false

nikophil (commented):
thanks for this @jrushlow 

On one hand, I definitely think this feature should land in Foundry, but on another hand, I'm a little bit reluctant to add new features to foundry 1.x. 

any thoughts @kbond?

Anyway, I'll double check in 2.x if stories work without persistence

src/Factory.php
jrushlow: I want to say no - but I can't remember why we couldn't use `$this->persist`. When I get some bandwidth this week, i'll dive back into this PR and see if we can't tidy things up..

Note to self:
run step debugger through https://github.com/zenstruck/foundry/blob/b66ed4ca2272491294928cb8ef6910c5c42c7954/src/Story.php#L175 to find out why `$this->persist` was a no-go

jrushlow:
> On one hand, I definitely think this feature should land in Foundry, but on another hand, I'm a little bit reluctant to add new features to foundry 1.x.

I'd make the argument that this is a bug fix rather than a new feature. Having said that, I completely understand the thought process behind merging this into `1.x` vs merging this / doing something else in `2.x`. I'm good with it either way.

src/Factory.php
nikophil: don't bother with debugging or what, I think it will be okay to merge as-is. First, because as you said, it's a bug fix. And secondly, it kinda OK to not have a super perfect code in 1.x which in few months will be legacy

tests/Functional/StoryTest.php
```diff
@@ -237,4 +239,14 @@ public function story_can_access_its_own_pool(): void
 
         self::assertContains($item->_real()->getName(), ['php', 'symfony']);
     }
+
+    /**
+     * @test
+     */
```
nikophil, line 245: because we're using a final class (`SomeObject`) with the legacy factory class `ModelFactory`:
```suggestion
    /**
     * @test
     * @group legacy
     */
```

by the way, shouldn't we add a test using `PersistentObjectFactory` and `ObjectFactory`?

src/Factory.php
nikophil: I've tested your PR and used `Factory::$persist instead of the new property `Factory::$shouldPersist` and all tests are green :thinking:

src/Proxy.php
nikophil: maybe we should also add a condition in methods `_delete()`, `_refresh()`. And maybe throw an exception in `_disableAutoRefresh()`?

kbond:
Hey, sorry for the delay reviewing. I'm inclined to add this to 2.x only. We are really close to a release and it's getting to be a real burden to port stuff from 1.x to 2.x.

nikophil:
based on @kbond's comment I'm closing this, this works in 2.x
see https://github.com/zenstruck/foundry/pull/589

PR #587, opened 2024-04-05
fix(2.x): restore PHPUnit error handler

nikophil:
this PR reproduces the behavior here: https://github.com/zenstruck/foundry/pull/577

At some point I'm wondering if we should not add in our tests permutations phpunit >= 10, WDYT?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
         "bamarni/composer-bin-plugin": "^1.8",
         "dama/doctrine-test-bundle": "^7.0|^8.0",
         "doctrine/common": "^3.2",
+        "doctrine/collections": "^1.7|^2.0",
         "doctrine/doctrine-bundle": "^2.10",
         "doctrine/doctrine-migrations-bundle": "^2.2|^3.0",
         "doctrine/mongodb-odm-bundle": "^4.6|^5.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -28,6 +28,7 @@
         "bamarni/composer-bin-plugin": "^1.8",
         "dama/doctrine-test-bundle": "^7.0|^8.0",
         "doctrine/common": "^3.2",
+        "doctrine/collections": "^2.0",
```
nikophil, line 31: I added this because of this error: https://github.com/zenstruck/foundry/actions/runs/8572205459/job/23494200535

I don't know why it occurs now.... but even with php 8.1 it happen...
it's just cheating with tests permutations :innocent:

composer.json
```diff
@@ -28,6 +28,7 @@
         "bamarni/composer-bin-plugin": "^1.8",
         "dama/doctrine-test-bundle": "^7.0|^8.0",
         "doctrine/common": "^3.2",
+        "doctrine/collections": "^1.7|^2.0",
```
nikophil, line 31: `doctrine/collections` < 1.7 is not compatible with phpunit 8.3, because of some return types

src/phpunit_helper.php
nikophil, line 1: I didn't know where to put this new function, so I created this file :shrug:

kbond: You didn't want to add a `KernelHelper` as done in #577?

nikophil: yes but I'm not sure what's best 
I think both are OK

kbond (approved):

PR #590, opened 2024-04-09
feat: "in memory" behavior

### LINKED ISSUE #533: support "in-memory" repositories
### Some words about the feature:
with a DDD approach, all repositories would be interfaces, which opens the door to have multiple implementations: usually a Doctrine one and a "in-memory" one. 

In one of my project, we use two different kernels in test: the common one, and a `InMemoryKernel` one, which injects in-memory repositories instead of doctrine ones.

Most of the factories have the following method, which will store the new object in a property of the in-memory repository:

```php
// SomeObjectFactory.php

    public static function inMemory(InMemorySomeObjectRepository $inMemorySomeObjectRepository): self
    {
        return self::new()->withoutPersisting()
            ->afterInstantiate(
                static fn (SomeObject $object) => $inMemorySomeObjectRepository->save($object)
            )
        ;
    }
```

Beside of this, I'm testing all my "doctrine" and "in-memory" repositories with the exact same class, using an abstract test class and two implementations which only have a `setUp()` method. This way I can be sure both implementations behave the same way and I know I can safely replace my doctrine repositories by the in-memory ones. Resulting in super fast kernel tests! 

Another very cool benefit is that we do not have to mock repositories anymore! Once the in-memory factories are initialized, we just need to call `$factory->create()` and the object is directly available in the tested code.

My current implementation works well, but is very perfectible, mainly because in some cases it is a little bit hard to work with relationships. I really think this is a very nice feature, which should be in Foundry!

### About the implementation I have in mind:

First, I'd like this behavior to be globally enabled or disabled with some kind of marker. Not sure which kind of marker: maybe  in a first implementation, a simple function like `enable_in_memory()` will suffice. But IMO the best way would be an attribute on the test method or test class.

We must introduce an interface for the in memory repositories: 
```php
/**
 * @template T of object
 */
interface InMemoryRepository
{
    /**
     * @param T $object
     */
    public function _save(object $object): void; 
    // the underscore prefix is needed in order to not conflict 
    // with a potential `save()` method in `SomeObjectRepositoryInterface`
}
```

Then, when the option is enabled, we need a way to  create in-memory factories. We'd need to hook in the factory creation process, in order to expose only in-memory factories, when the feature is enabled.

This will need a little bit of refactoring because currently, when a factory is not a service, we create it with a static call: `Factory::new()`

My suggestion would be to change how `FactoryRegistry` behaves and to create the factory inside of it, if we don't find it in the container. 
```diff
// Zenstruck\Foundry\FactoryRegistry
-    public function get(string $class): ?Factory
+    public function get(string $class): Factory
     {
         foreach ($this->factories as $factory) {
             if ($class === $factory::class) {
                return $factory;
             }
         }
 
-        return null;
+        return new $class(); // todo: handle `ArgumentCountError`
     }
```
(it would become a.... `FactoryFactory` :scream:  but this name is really awful, I think we can keep the current name)

Then, we could extract an interface from the `FactoryRegistry` and decorate it with `InMemoryFactoryRegistry`:

```php
// Zenstruck\Foundry\InMemory\InMemoryFactoryRegistry
public function get(string $class): Factory
{
    if (/** in memory is not enabled */) {
        return $this->decorated->get($class);			
    }

    return $this->decorated
        ->get($class)
        ->withoutPersisting()
        ->afterInstantiate(
            static fn (object $object) => $this->findInMemoryRepository($class)->_save($object)
        )
    ;
}
```

We also need a way to guess the in-memory repository from the factory's class name. One of the solutions would be to introduce an new attribute `#[AsInMemoryRepository(class: Object::class)]`.

And voilà! :tada: this is all we need as a first step.
Next step would be to to provide a in-memory version of `RepositoryDecorator` because, I'd really like to be able to use things like `::findOrCreate()` within the in-memory tests. And then `RepositoryAssertions` should also be needed! But let's keep simple for a first iteration 😅

As a bonus, I think we can isolate all the code into a `Zenstruck\Foundry\InMemory` namespace, which will eventually have its own repo in the future.

Do you have any thoughts about this?

nikophil:
This PR introduces "in-memory" behavior

fixes https://github.com/zenstruck/foundry/issues/533

It makes integration testing with "in memory" repositories very easy!

Here is the proposal:
Given some kind of DDD or hexagonal architecture, repositories in domain are usually interfaces, which main implementation is a Doctrine one. "in memory" version of these repositories will exist and would be injected in the container in test environment.

This PR exposes a new interface [`Zenstruck\Foundry\InMemory\InMemoryRepository`](https://github.com/nikophil/foundry/blob/feat/in-memory-behavior/src/InMemory/InMemoryRepository.php), e.g.:

```php
/**
 * @implements InMemoryRepository<Address>
 */
final class InMemoryAddressRepository implements InMemoryRepository, DomainAddressRepository
{
    /** @use InMemoryRepositoryTrait<Address> */
    use InMemoryRepositoryTrait;

    public static function _class(): string
    {
        return Address::class;
    }

    // + all methods implementing "DomainAddressRepository"
}
```

It also exposes a new attribute `#[AsInMemoryTest]`, which will disable persistence of the factories, and register an "after instantiate" hook, which will store the objects in their respective "in memory" repositories:

```php
#[AsInMemoryTest]
class SomeTest extends KernelTestCase
{
    private InMemoryAddressRepository $addressRepository;
 
    {
        $this->addressRepository = self::getContainer()->get(InMemoryAddressRepository::class);
    }

    #[Test]
    public function object_should_be_accessible_from_in_memory_repository(): void
    {
        $address = AddressFactory::createOne();

        self::assertSame([$address], $this->addressRepository->_all());
    }
}
```

A `GenericInMemoryRepository` is also added for convenience, when the given in memory repository is missing.


src/InMemory/functions.php
```diff
@@ -0,0 +1,22 @@
...
+namespace Zenstruck\Foundry\InMemory;
+
+use Zenstruck\Foundry\Configuration;
+
+/**
+ * Enable "in memory" repositories globally.
+ */
+function enable_in_memory(): void
+{
+    Configuration::instance()->enableInMemory();
```
nikophil, line 21: At first, I did something like:
```php
Configuration::instance()->persistence()->enableInMemory();
```

and the `$inMemory` boolean was stored in `PersistenceManager`. But then, all `InMemoryTest` did not worked when no persistence is activated, which is a non-sense :shrug: 

Then I decided to store this in the `Configuration` class

src/Configuration.php
```diff
@@ -82,12 +95,29 @@ public static function isBooted(): bool
...
+
+    public function enableInMemory(): void
+    {
+        $this->inMemory = true;
+    }
+
+    public function isInMemoryEnabled(): bool
+    {
+        return $this->inMemory;
+    }
```
nikophil, line 122: At first, I did something like:
```php
Configuration::instance()->persistence()->enableInMemory();
```

and the `$inMemory` boolean was stored in `PersistenceManager`. But then, all `InMemoryTest` did not worked when no persistence is activated, which is a non-sense :shrug: 

Then I decided to store this in the `Configuration` class

src/Configuration.php
```diff
@@ -82,12 +95,36 @@ public static function isBooted(): bool
...
 
     public static function shutdown(): void
     {
         StoryRegistry::reset();
         self::$instance = null;
     }
+
+    public function enableInMemory(): void
+    {
+        $this->inMemory = true;
```
smnandre: Maybe throw if `null === $this->inMemoryRepositoryRegistry`  ?

src/InMemory/AsInMemoryTest.php
```diff
@@ -0,0 +1,36 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Zenstruck\Foundry\InMemory;
+
+/**
+ * @author Nicolas PHILIPPE <nikophil@gmail.com>
+ */
+#[\Attribute(\Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
+final class AsInMemoryTest
```
smnandre, line 20: 😍

smnandre:
This is brilliant

kbond (approved):
This looks good, let's roll after some docs?

nikophil:
> let's roll after some docs?

yes! I'll add some doc soon. I also think I will release Foundry 2.4 before, A second PR around "in memory" will land before: I want to provide a `InMemoryRepositoryDecorator`, so that we can make assertions as easily as we use to do with doctrine repos

nikophil:
@kbond I've finally added some docs, would you mind to have a look, please?

docs/index.rst
```diff
@@ -2149,6 +2149,77 @@ You will need to configure manually Foundry. Unfortunately, this may mean duplic
 
 .. _stories:
 
+In-memory behavior
```
```suggestion
In-memory Behavior
```

docs/index.rst
```diff
@@ -2149,6 +2149,77 @@ You will need to configure manually Foundry. Unfortunately, this may mean duplic
...
+~~~~~~~~~~~~~~~~~~
+
+Foundry allows to use "in-memory" repositories in your factories. This is mainly useful for `DDD <https://en.wikipedia.org/wiki/Domain-driven_design>`_
+applications or with `hexagonal architecture <https://en.wikipedia.org/wiki/Hexagonal_architecture_(software)>`, where
+repositories in domain are usually interfaces for which main implementations are a Doctrine one. You can tell Foundry to
```
```suggestion
repositories in a domain are usually interfaces for which main implementations are a Doctrine one. You can tell Foundry to
```

docs/index.rst
```diff
@@ -2149,6 +2149,77 @@ You will need to configure manually Foundry. Unfortunately, this may mean duplic
...
+    }
+
+Then, the "in-memory" repository should be used in Symfony's container, as the main repository implementation:
+
+.. code-block:: yaml
+
+    # config/services.yaml
+    when@test:
+        services:
+            App\Domain\Address\DomainAddressRepositoryInterface: '@App\Infrastructure\Address\InMemoryAddressRepository'
```
kbond: Doesn't this have an impact on non-`#[AsInMemoryTest]` tests?

nikophil: you are right, I was currently reviewing this, and came to the same conclusion. In my project where I use this, I'm using a `InMemoryKernel`

nikophil: I replaced it with a simple sentence, with no code, I'll see in the future how I improve this section.

nikophil:
after more than one year, I'm merging this PR :tada: 

I still need few things before releasing, but that's very close :)

docs/index.rst
```diff
@@ -2104,6 +2104,71 @@ You will need to configure manually Foundry. Unfortunately, this may mean duplic
 
 .. _stories:
 
+In-memory Behavior
+~~~~~~~~~~~~~~~~~~
+
+Foundry allows to use "in-memory" repositories in your factories. This is mainly useful for `DDD <https://en.wikipedia.org/wiki/Domain-driven_design>`_
+applications or with `hexagonal architecture <https://en.wikipedia.org/wiki/Hexagonal_architecture_(software)>`, where
```
jdreesen, line 2111: It seems like this link is broken in the rendered docs: 

[grafik](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#in-memory-behavior)

nikophil: Hi @jdreesen 

thanks for reporting! I've [already fixed it](https://github.com/zenstruck/foundry/pull/904) but the docs are only generated once per day

Issue #516, opened 2023-11-07
2.x checklist


kbond:
_(to be moved to 2.x PR)_

- [x] remove deprecated code
- [ ] migrate remaining features from foundry-next
- [ ] deprecate `Zenstruck\Foundry\anonymous()` in favor of `Zenstruck\Foundry\factory()`
- [ ] split packages? (foundry-persistence, foundry-orm, foundry-mongo)
- [ ] remove requirement/need for `zenstruck/callback`
- [ ] [Attributes can be a callable that accepts an int](https://github.com/kbond/foundry-next/blob/6741d5b4acbe39b2ade6064a746a91941f2d049e/src/Factory.php#L23) (marking the current # being created when creating a collection)

nikophil:
- [ ] `inverse_one_to_many_relationship_without_cascade()` https://github.com/kbond/foundry-next/pull/12
- [x] fixed by https://github.com/kbond/foundry-next/pull/17
	- [x] `one_to_many_with_two_relationships_same_entity()` 
	- [x] `one_to_many_with_two_relationships_same_entity_and_adders()`
	- [x] `can_use_adder_as_attributes()`
	- [x] `inverse_many_to_many_with_two_relationships_same_entity()`
	- [x] `one_to_many_polymorphic_with_nested_collection_relationship()`
- [x] `ModelFactoryTest::can_set_state_via_new` => https://github.com/zenstruck/foundry/pull/559
- [x] fixed by https://github.com/kbond/foundry-next/pull/22
	- [x] `ODMModelFactoryTest::can_use_adder_as_attributes`
	- [x] `ODMModelFactoryTest::can_find_or_create_from_proxy_of_object`
	- [x] `ODMModelFactoryTest::can_hydrate_embed_many_fields`
	- [x] `ODMModelFactoryTest::can_use_factory_for_embedded_object`


## Static analysis
- [x] I think we have a complex SCA problem:
```php
// SomeObjectFactory extends PersistentProxyObjectFactory
return $this->instantiateWith(
    static fn () => new SomeObject()
)
```
this code results in this error (I've replaced the full object FQCN by `T` for better readability)
> Parameter #1 $instantiator of method Zenstruck\Foundry\ObjectFactory<T&Zenstruck\Foundry\Persistence\Proxy<T>>::instantiateWith()
[...]

PR #594, opened 2024-04-20
doc: update

kbond:
EDIT (from @nikophil)

I think the doc on symfony.com is updated as soon as the doc is modified (disregarding releases). So, here is the plan for 2.x:
- finish and validate this PR without merging it
- once validated, copy this PR into another which will target 2.x and add missing pieces
- merge both doc PR + release both 1.x and 2.x 🚀

todo for 2.x documentation:
- [x] document `ArrayFactory`
- [x] ~document `__index` for factory collections~ this was already documented
- [x] `make:factories`: document `--with-phpdoc` option and add change the note about `@method`  
- [x] remove documentation of the config `auto_refresh_proxies`

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -198,6 +198,11 @@ This command will generate a ``PostFactory`` class that looks like this:
 
 .. tip::
 
+    You can also inherit from `Zenstruck\Foundry\Persistence\PersistentObjectFactory`. Which will create regular objects
+    without proxy (see :ref:`Proxy object section <object-proxy>` for more information).
+
+.. tip::
+
     You can globally configure which namespace the factories will be generated in:
 
     .. configuration-block::
@@ -211,6 +216,38 @@ This command will generate a ``PostFactory`` class that looks like this:
                         default_namespace: 'App\\MyFactories'
 
     You can override this configuration by using the ``--namespace`` option.
+
+.. note::
+
+    The generated ``@method`` docblocks above enable autocompletion with PhpStorm but
+    causes errors with PHPStan. To support PHPStan for your factory's, you need to *also*
+    add the following dockblocks:
+
+    .. code-block:: php
+
+        /**
+         * ...
+         *
+         * @phpstan-method Proxy<Post> create(array|callable $attributes = [])
+         * @phpstan-method static Proxy<Post> createOne(array $attributes = [])
+         * @phpstan-method static Proxy<Post> find(object|array|mixed $criteria)
+         * @phpstan-method static Proxy<Post> findOrCreate(array $attributes)
+         * @phpstan-method static Proxy<Post> first(string $sortedField = 'id')
+         * @phpstan-method static Proxy<Post> last(string $sortedField = 'id')
+         * @phpstan-method static Proxy<Post> random(array $attributes = [])
+         * @phpstan-method static Proxy<Post> randomOrCreate(array $attributes = [])
+         * @phpstan-method static list<Proxy<Post>> all()
+         * @phpstan-method static list<Proxy<Post>> createMany(int $number, array|callable $attributes = [])
+         * @phpstan-method static list<Proxy<Post>> createSequence(array|callable $sequence)
+         * @phpstan-method static list<Proxy<Post>> findBy(array $attributes)
+         * @phpstan-method static list<Proxy<Post>> randomRange(int $min, int $max, array $attributes = [])
+         * @phpstan-method static list<Proxy<Post>> randomSet(int $number, array $attributes = [])
+         * @phpstan-method static RepositoryProxy<Post> repository()
+         */
+        final class PostFactory extends ModelFactory
+        {
+            // ...
+        }
 
 In the ``defaults()``, you can return an array of all default values that any new object
 should have. `Faker`_ is available to easily get random data:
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -207,45 +209,10 @@ This command will generate a ``PostFactory`` class that looks like this:
                 zenstruck_foundry:
                     make_factory:
                         default_namespace: 'App\\MyFactories'
-                    make_story:
-                        default_namespace: 'App\\MyStories'
 
     You can override this configuration by using the ``--namespace`` option.
 
-
-.. note::
```
kbond, line 216: I think this is no longer applicable - @nikophil, can you confirm?

docs/index.rst
```diff
@@ -2046,150 +1967,73 @@ This way, there is just one place to set your config.
...
+.. code-block:: yaml
 
-            // Configure faker to be used by your factories.
-            'faker' => [
+    zenstruck_foundry:
 
-                // Change the default faker locale.
-                'locale' => null,
+        # Whether to auto-refresh proxies by default (https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#auto-refresh)
+        auto_refresh_proxies: null
```
nikophil, line 2033: I don't think we should document this, since it has no effect

docs/index.rst
```diff
@@ -129,65 +129,67 @@ This command will generate a ``PostFactory`` class that looks like this:
...
+     * @method static PostRepository|ProxyRepositoryDecorator repository()
+     * @method static Post[]|Proxy[]                          all()
+     * @method static Post[]|Proxy[]                          createMany(int $number, array|callable $attributes = [])
+     * @method static Post[]|Proxy[]                          createSequence(iterable|callable $sequence)
+     * @method static Post[]|Proxy[]                          findBy(array $attributes)
+     * @method static Post[]|Proxy[]                          randomRange(int $min, int $max, array $attributes = [])
+     * @method static Post[]|Proxy[]                          randomSet(int $number, array $attributes = [])
      */
-    final class PostFactory extends ModelFactory
+    final class PostFactory extends PersistentProxyObjectFactory
```
nikophil, line 155: don't we miss an introduction which explains the difference between `PersistentProxyObjectFactory`, `PersistentObjectFactory` and `ObjectFactory`? with pros and cons for each?

maybe also a word on `ArrayFactory`?

docs/index.rst
nikophil: Actually this is still true. If you have a `@method` which basically is a lie, you'll need the `@phpstan-method` annotation.

But I've just tested with foundry 2 + phpstorm 2023.3, and autocompletion comes out of the box, even with `PersistentProxyObjectFactory`

I think this is because phpstorm really improved how it handles generics + we now return `@return T&Proxy<T>` which is easier to understand than the old `Proxy<T>` which is a `@mixin` of a template, which phpstorm notoriously does not handle very well.

So... maybe we should remove all the `@method` even in doc? On the other hand, not anybody uses phpstorm. So we can document both `@method` and `@phpstan-method` in a `.. note::`: "if autocomplete does not work out of the box, you can still add `@method`, but be sure to add `@phpstan-method` as well..."?

I think we should also remove these phpdoc from the `make:factory` command, and make it optional with an option `--add-phpdocs`. WDYT?

nikophil (changes requested):

nikophil:
something else: WDYT about documenting [this little, but still useful, behavior](https://github.com/zenstruck/foundry/blob/2.x/tests/Integration/Persistence/GenericFactoryTestCase.php#L172) (the stuff with `__index`)?

docs/index.rst
```diff
@@ -207,50 +209,15 @@ This command will generate a ``PostFactory`` class that looks like this:
                 zenstruck_foundry:
                     make_factory:
                         default_namespace: 'App\\MyFactories'
-                    make_story:
-                        default_namespace: 'App\\MyStories'
```
nikophil, line 211: shouldn't we keep this? based on https://github.com/zenstruck/foundry/issues/516#issuecomment-2033692112 I think we should (and upmerge the related PR)

docs/index.rst
nikophil: I've made the change here: https://github.com/zenstruck/foundry/pull/611

because the PR targets 1.x (which is correct, since we've changed the api in 1.x), and because I'm not willing to make this change in 1.x (an it won't work, because my answer is only pertinent for 2.x and the new proxy mechanism) I've added back the note about `@method` here, we'll modify it in a further PR for 2.x

docs/index.rst
nikophil: actually this comments refers to 2.x, and this PR target 1.x

docs/index.rst
nikophil: `ArrayFactory` are introduced in 2.x, it should be documented in another PR

I've added a note about `ObjectFactory` and `PersistentObjectFactory` in this PR

kbond (commented):
Thanks for making these changes @nikophil!

Your plan for syncing this with 2.x makes sense. We'll need to update https://github.com/zenstruck/foundry/blob/1.x/.symfony.bundle.yaml (probably on 1.x and 2.x)

nikophil (approved):

PR #602, opened 2024-05-07
Default to false for 'embedded' and 'targetDocument' mappings

melkamar:
When an ODM reference is configured, the 'embedded' key may not be present at all in the association mapping. The missing key should be treated as implicit false. Without this the make:factory command will crash.


nikophil:
hi @melkamar 

thanks for this! would you mind fixing the phpstan error? :pray:

melkamar:
Looks good now 🙂

kbond (approved):

PR #610, opened 2024-06-02
fix: prevent refresh error with autorefresh

nikophil:
when an proxified object is deleted, we should be able to call its methods and access its properties...

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,5 +1,3 @@
-version: '3.9'
-
 services:
     mysql:
         image: mysql:5.7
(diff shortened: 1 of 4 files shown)


docker-compose.yaml
```diff
@@ -1,5 +1,3 @@
-version: '3.9'
```
nikophil, line 1: prevents a warning in docker

src/Persistence/IsProxy.php
```diff
@@ -114,14 +115,19 @@ public function _repository(): ProxyRepositoryDecorator
 
     private function _autoRefresh(): void
     {
-        if (!$this->_autoRefresh) {
+        // "??=" fixes in ProxyHelper where object gets created without any props initialized
+        if (!($this->_autoRefresh ??= true)) {
```
nikophil, line 119: since `symfony/var-exporter` 7.1, there is a strange bug where `IsProxy::$_autorefresh` is not initialized, resulting in a stranger bug:
```
TypeError: Value of type null returned from ZenstruckFoundryTestsFixtureEntityGenericEntityProxy::__get() must be compatible with unset property ZenstruckFoundryTestsFixtureEntityGenericEntityProxy::$_autoRefresh of type bool
```

what's strange is that it seems there is not difference between `symfony/var-exporter` 7.0.8 and 7.1.0 :shrug:

PR #611, opened 2024-06-03
feat(2.x, make:factory): add phpdoc conditionnally with --with-phpdoc option

nikophil:


src/Persistence/IsProxy.php
```diff
@@ -114,7 +114,7 @@ public function _repository(): ProxyRepositoryDecorator
 
     private function _autoRefresh(): void
     {
-        if (!$this->_autoRefresh) {
+        if (!$this->_autoRefresh ??= true) {
```
nikophil: see https://github.com/zenstruck/foundry/pull/610/files#r1623927636

kbond (approved):
Once the conflict is resolved, merge away!

PR #613, opened 2024-06-07
docs: update docs for foundry v2

nikophil:


.symfony.bundle.yaml
```diff
@@ -1,3 +1,3 @@
-branches: ["1.x", "1.23.x"]
-maintained_branches: ["1.x", "1.23.x"]
+branches: ["2.x", "1.x"]
+maintained_branches: ["2.x", "1.x"]
```
nikophil, line 2: Do you think we need to keep `1.23.x`? I don't event remember why we decided to keep it :sweat_smile: (maybe a BC break in 1.24.... :roll_eyes: )

nikophil, line 2: I put the `2.x` in first position, I don't know how it would impact the order in the symfony docs. I'm assuming the first one is the default one

kbond: No, let's drop now!

PR #615, opened 2024-06-08
fix(orm): reset database instead of dropping the schema when using migrations

vincentchalamon:
### Description of the bug

On Foundry 2.x, the `resetSchema` method calls `dropSchema` whatever mode is configured. On Foundry 1.x, the same call detects the `migrate` mode and resets the database instead of dropping the schema.

I think this was made this way in case of multiple schemas in the same database (which is the case in my client project), and to prevent any undesired behaviors (e.g: `doctrine:schema:drop` doesn't seem to remove all schemas from the database in case of multiple schemas).

### Possible solution

Backporting some code from 1.x to detect the mode on `dropSchema` and reset the database instead fixes this bug.


nikophil:
thanks @vincentchalamon

PR #616, opened 2024-06-10
Fix `Instantiator::allowExtra` example

norkunas:
Fixes an example which leads to `Expected parameter of type 'string', 'string[]' provided`


nikophil:
thanks @norkunas

PR #629, opened 2024-06-11
test: add test with multiple ORM schemas

### LINKED ISSUE #618: [Tests] use 2 schemas in the same database
after merging https://github.com/zenstruck/foundry/pull/615, in order to prevent any regressions in "reset database" behavior, I think we should reproduce in test a case where 2 schemas exist in the same database

ping @vincentchalamon

vincentchalamon:
Fixes #618

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -15,6 +15,7 @@ use Symfony\Component\Console\Output\NullOutput;
 use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\Filesystem\Filesystem;
 use Zenstruck\Foundry\ORM\AbstractORMPersistenceStrategy;
+use Symfony\Component\Finder\Finder;
 use Zenstruck\Foundry\Tests\Fixture\TestKernel;
 
 require \dirname(__DIR__).'/vendor/autoload.php';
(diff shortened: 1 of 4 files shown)


tests/bootstrap.php
```diff
@@ -29,6 +30,11 @@
     $fs->remove(__DIR__.'/Fixture/Migrations');
     $fs->mkdir(__DIR__.'/Fixture/Migrations');
 
+    // restore custom migrations
```
kbond: Can we use `Symfony\Component\Filesystem\Filesystem` to just copy the entire directory?

kbond (commented):
This looks great, thanks for working on this Vincent!

Regarding the failing CI (I'm not familiar with multiple schemas). Do you know what's going on there?

tests/bootstrap.php
vincentchalamon: Filesystem only allows to copy files, not directories

vincentchalamon:
Hi @kbond,

> Regarding the failing CI (I'm not familiar with multiple schemas). Do you know what's going on there?

The `doctrine:schema:update` command doesn't seem to support multiple schemas. To handle such specific use case, I had to separate `migrate` mode from `schema` mode fixtures. Push in progress.

This PR is still in WIP till I find a clean solution to test this specific use case.

tests/bootstrap.php
kbond: `Filesystem::mirror()` should do the trick (but this isn't a big deal)

kbond (approved):

nikophil:
Hi @vincentchalamon 

I'm wondering if a simpler solution would not be to declare two entity managers, affect one entity to this entity manager, and configure the test kernel to not reset this specific EM. WDYT?

vincentchalamon:
> Hi @vincentchalamon
> 
> I'm wondering if a simpler solution would not be to declare two entity managers, affect one entity to this entity manager, and configure the test kernel to not reset this specific EM. WDYT?

I'm afraid using multiple entity managers won't solve the original issue. The idea is to add non-regression tests to ensure the database is fully and properly removed before running the migrations (cf. https://github.com/zenstruck/foundry/pull/615).

The idea here is to have a custom migration which should fail the tests if not run properly (for instance, if `AbstractORMPersistenceStrategy::resetSchema` method is updated).

nikophil:
ok thanks for explanations, I understand better now!

nikophil (approved):

PR #631, opened 2024-06-11
fix(2.x): add back second parameter for after persist callbacks

### LINKED ISSUE #628: [2.x] ArgumentCountError thrown when passing $attributes array to closure in event
| Q | A
| --- | ---
| Symfony Version | 6.4.8
| Foundry Version | 2.0.1
| PHP Version | 8.3.8
| PHPUnit Version | 9.6.19

I've been upgrading my project to use Foundry v2.0. Everything has gone relatively smoothly except updating an event in one of my factory classes. I suspect the problem is on my end and not a bug, but I'm not sure what the problem is.

I'm using Foundry's default bundle configuration.

In a test I create an object as follows:

```php
$object = SomeEntityFactory::createOne([
    // standard entity properties

    // extra property
    'account' => [
        'name' => 'Account name',
        'uid' => 'UID',
    ],
]);
```

In 1.x, I did this:
```php
final class SomeEntityFactory extends ModelFactory
{
    // ...

    {
        return $this
            ->instantiateWith((new Instantiator())->allowExtraAttributes(['account']))
            ->afterPersist(function (SomeEntity $object, array $attributes) {
                // do some stuff with $attributes['account']
            });
    }

    // ...
}
```

And for 2.x, I currently have this:
```php
final class SomeEntityFactory extends PersistentProxyObjectFactory
{
    // ...

    {
        return $this
            ->instantiateWith(Instantiator::withConstructor()->allowExtra('account'))
            ->afterPersist(function (SomeEntity $object, array $attributes) {
                // do some stuff with $attributes['account']
            });
    }

    // ...
}
```

The latter causes an error in 2.x:
> ArgumentCountError : Too few arguments to function App\Tests\Factory\SomeEntity\SomeEntityFactory::App\Tests\Factory\SomeEntity\{closure}(), 1 passed in /path-to-project/vendor/zenstruck/foundry/src/Persistence/PersistentObjectFactory.php on line 230 and exactly 2 expected

PHPStan gives the following error:

> phpstan: Parameter # 1 $callback of method Zenstruck\Foundry\Persistence\PersistentObjectFactory<App\Entity\SomeEntity&Zenstruck\Foundry\Persistence\Proxy<App\Entity\SomeEntityt>>::afterPersist() expects callable(App\Entity\SomeEntity&Zenstruck\Foundry\Persistence\Proxy<App\Entity\SomeEntity>): void, Closure(App\Entity\SomeEntity, array): void given

> Parameter # 2 $attributes of passed callable is required but accepting callable does not have that parameter. It will be called without it.

I've gone over the documentation and upgrade guide to see what I might be doing wrong. I've attempted to debug the issue and that's only muddied things further.

I feel like I'm missing something obvious. What would be the correct way to achieve the above in 2.x?

Thank you!

nikophil:
fixes https://github.com/zenstruck/foundry/issues/628

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -562,14 +562,6 @@ they were added.
             // $object is the instantiated object
             // $attributes contains the attributes used to instantiate the object and any extras
         })
-        ->afterPersist(function(Proxy $proxy, array $attributes) {
-            /* @var Post $proxy */
-            // this event is only called if the object was persisted
-            // $proxy is a Proxy wrapping the persisted object
-            // $attributes contains the attributes used to instantiate the object and any extras
-        })
-
-        // if the first argument is type-hinted as the object, it will be passed to the closure (and not the proxy)
         ->afterPersist(function(Post $object, array $attributes) {
             // this event is only called if the object was persisted
             // $object is the persisted Post object
--- a/src/Persistence/PersistentObjectFactory.php
+++ b/src/Persistence/PersistentObjectFactory.php
@@ -35,7 +35,7 @@ abstract class PersistentObjectFactory extends ObjectFactory
 {
     private bool $persist;
 
-    /** @var list<callable(T):void> */
+    /** @var list<callable(T, Parameters):void> */
     private array $afterPersist = [];
 
     /** @var list<callable(T):void> */
@@ -226,11 +226,13 @@ abstract class PersistentObjectFactory extends ObjectFactory
 
         $this->tempAfterPersist = [];
 
-        foreach ($this->afterPersist as $callback) {
-            $callback($object);
-        }
-
         if ($this->afterPersist) {
+            $attributes = $this->normalizeAttributes($attributes);
+
+            foreach ($this->afterPersist as $callback) {
+                $callback($object, $attributes);
+            }
+
             $configuration->persistence()->save($object);
         }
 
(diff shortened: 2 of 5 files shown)


src/Persistence/PersistentObjectFactory.php
```diff
@@ -226,11 +226,13 @@ final public function create(callable|array $attributes = []): object
 
         $this->tempAfterPersist = [];
 
-        foreach ($this->afterPersist as $callback) {
-            $callback($object);
-        }
-
         if ($this->afterPersist) {
+            $attributes = $this->normalizeAttributes($attributes);
```
nikophil, line 230: not a huge fan of this:
- this is already done in `parent::create()`
- we won't pass attributes added in `beforeInstantiate` callbacks

any thought @kbond ?

docs/index.rst
```diff
@@ -562,14 +562,6 @@ they were added.
             // $object is the instantiated object
             // $attributes contains the attributes used to instantiate the object and any extras
         })
-        ->afterPersist(function(Proxy $proxy, array $attributes) {
```
nikophil, line 565: we're not doing this anymore... This is a BC break between 1.x and 2.x :shrug:

src/Persistence/PersistentObjectFactory.php
kbond: Hmm, I agree, this doesn't feel great. I guess I totally missed this part when working on 2.0...

PR #634, opened 2024-06-12
fix: Attributes added in `beforeInstantiate` are not passed to `afterPersist`

kbond:
Followup to quick fix in #631 that demonstrates an edge case it didn't fix.

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -28,11 +28,18 @@ abstract class Factory
     /** @var Attributes[] */
     private array $attributes;
 
+    /**
+     * Memoization of normalized parameters
+     *
+     * @internal
+     * @var Parameters|null
+     */
+    protected array|null $normalizedParameters = null;
+
     // keep an empty constructor for BC
     public function __construct()
     {
     }
-
 
     /**
      * @param Attributes $attributes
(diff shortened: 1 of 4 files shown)


src/Factory.php
```diff
@@ -28,12 +28,19 @@ abstract class Factory
     /** @var Attributes[] */
     private array $attributes;
 
+    /**
+     * Memoization of normalized parameters
+     *
+     * @internal
+     * @var Parameters|null
+     */
+    protected array|null $normalizedParameters = null;
```
kbond, line 37: @nikophil, don't we need to set this back to `null` on clone?

nikophil: need to test this

nikophil: ok, so... an easy test like this one (that I added to our test suite) was failing because of the memoization:
```php
    public function can_create_different_objects_based_on_same_factory(): void
    {
        $factory = Object1Factory::new(['prop1' => 'first object']);
        $object1 = $factory->create();
        self::assertSame('first object-constructor', $object1->getProp1());

        $object2 = $factory->create(['prop1' => 'second object']);
        self::assertSame('second object-constructor', $object2->getProp1());

        $object3 = $factory->with(['prop1' => 'third object'])->create();
        self::assertSame('third object-constructor', $object3->getProp1());
    }
```

I'm really not proud of the solution I found (replace `$this->normalizedParameters ??=` by `$this->normalizedParameters =` in `Factory::normalizeParameters()`, but it's the only simple one I found :shrug:

nikophil:
closed in favor of https://github.com/zenstruck/foundry/pull/745 because I couldn't push force after the rebase

PR #635, opened 2024-06-13
fix: unproxy args in proxy objects

nikophil:
Here is a first attempt to mitigate the problem when proxy objects are passed to entity manager.

I also tried to explain the problem in the docs.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -231,21 +231,21 @@ This command will generate a ``PostFactory`` class that looks like this:
          * @method static Post[]|Proxy[] randomRange(int $min, int $max, array $attributes = []))
          * @method static Post[]|Proxy[] randomSet(int $number, array $attributes = []))
          *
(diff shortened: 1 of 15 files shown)


kbond (approved):

tests/Integration/ORM/EntityProxyFactoryRelationshipTest.php
```diff
@@ -50,6 +51,20 @@ public function doctrine_proxies_are_converted_to_foundry_proxies(): void
         $this->assertInstanceOf(StandardCategory::class, $category);
     }
 
+    /**
+     * @test
+     */
+    public function it_can_set_value_of_a_many_to_one(): void
```
kbond: As suggested in slack, could you add a test for one-to-many (adding a proxied object to a collection)?

kbond (changes requested):

PR #636, opened 2024-06-13
docs: kernel must be shutdown sometimes

matthieumota:
I begin to use Foundry on a new project.

Follow the docs to write a first tests and we can't create client after boot kernel (Kernel is boot by Trait [`Factories`](https://github.com/zenstruck/foundry/blob/2.x/src/Test/Factories.php)).

We need to ensure kernel is shutdown or create client before call factory.


nikophil:
That's right!

thanks @matthieumota

PR #645, opened 2024-06-20
fix: create ArrayCollection if needed

### LINKED ISSUE #474: `FactoryCollection` should create `ArrayCollection` if needed
when typing entities collections with `\Doctrine\Common\Collections\Collection` an error occurs:
```php
class Post
{
    /** @var Collection<Comment> */
    private Collection $comments;
}

PostFactory::createOne(
	['comments' => CommentFactory::new()->many(1)]
);
```

The above code results in an error:
```bash
TypeError: Cannot assign array to property App\Domain\Model\Post::$comments of type Doctrine\Common\Collections\Collection
```

although I'm not really sure how we should detect we need collections over arrays :thinking: (try/catch? reflection? use `symfony/property-info`?)

One workaround I'm using also causes problems. 
```php
PostFactory::createOne(
	['comments' => new ArrayCollection(CommentFactory::createMany(1))]
);
```

Above code results in this error:
```
Doctrine\Persistence\Mapping\MappingException: The class 'Zenstruck\Foundry\Proxy' was not found in the chain configured namespaces App\Domain
```

nikophil:
_please review https://github.com/zenstruck/foundry/pull/653 before, as this PR is built is the top of the other_

fixes https://github.com/zenstruck/foundry/issues/474

this is a really naive approach, nonetheless, this could be helpful. 
It will only work if the property is actually type hinted with `Doctrine\Common\Collections\Collection`.
It won't create an `ArrayCollection` if the property is only type hinted with phpdoc, or if it has a complex type hint (for instance, I saw in few occasions something like `private array|Collection $relations`). 

Another cleaner approach would have been to use doctrine metadata in `Hydrator` but this seemed convoluted.

But I think the current implementation is good enough for now, since the type hint with `Collection` is how it is documented by Doctrine.

Maybe in the future we might support other `\Traversable` objects?


kbond (approved):
> this is a really naive approach, nonetheless, this could be helpful.

I think this is OK for now. From what I can see, we have the ability to change/enhance later w/o BC breaks.

PR #648, opened 2024-06-21
feature: allow to use `Factory::create()` and factory service in data providers

### LINKED ISSUE #532: [Foundry 2] Allow objects creation in data provider
Maybe we could use lazyness behavior of the proxy system and delay persistence when the object is accessed

nikophil:
fixes https://github.com/zenstruck/foundry/issues/532

Since https://github.com/sebastianbergmann/phpunit/pull/5913, PHPUnit's extensions `bootstrap()` methods are called before data provider and since https://github.com/sebastianbergmann/phpunit/pull/5917, data provider related events are dispatched synchronously (before this, they were deferred, which wasn't useful for us).

This now opens the door for cool things:
- we can boot Foundry "for kernel tests" before the data providers are called, allowing to use service factories in data providers
- we can now detect whether or not we are inside a data provider, and decide to return a proxy instead of actually calling `create()` in the data provider.

so, after this PR will be merged, all of these will be working:
```php
final class SomeUnitTest extends TestCase
{
    #[DataProvider('provide')]
    public function testSomethingWithoutDB(SomeObject $object): void
    {
        // ...
    }

    public static function provide(): iterable
    {
        yield [SomeObjectFactory::createOne()];
        yield [SomePersistentObjectFactory::createOne()]; // won't try to store anything in db
        yield [SomePersistentProxyObjectFactory::createOne()]; // won't try to store anything in db
    }
}

final class SomeIntegrationTest extends KernelTestCase
{
    #[DataProvider('provide')]
    public function testSomethingWithDB(SomeObject $object): void
    {
        // ...
    }

    public static function provide(): iterable
    {
        yield [SomeProxyObjectPersistentFactory::createOne()];
        yield [SomeProxyObjectPersistentFactoryAsAService::createOne()]; 
    }
}
```

Since we **need** to return a proxy object, I decided that trying to call `NotAProxyButStillPersistentFactory::create()` inside a data provider will throw an error, but I'm open to change this.

Sadly, the feature will land in PHPUnit 11.4, which will be released in October. So, we have plenty of time to test it :)

---

Beside of this, the newly merged PR in PHPUnit, will allow to use the ["in-memory" behavior](https://github.com/zenstruck/foundry/pull/590) in data providers, which was my last blocker for this awesome feature.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   tests:
-    name: P:${{ matrix.php }}, S:${{ matrix.symfony }}, D:${{ matrix.database }}, PU:${{ matrix.phpunit }}${{ matrix.deps == 'lowest' && ' (lowest)' || '' }}${{ matrix.use-dama == 1 && contains(matrix.database, 'sql') && ' (dama)' || '' }}${{ !contains(matrix.database, 'sql') && '' || matrix.use-migrate == 1 && ' (migrate)' || ' (schema)' }}
+    name: P:${{ matrix.php }}, S:${{ matrix.symfony }}, D:${{ matrix.database }}, PU:${{ matrix.phpunit }}${{ matrix.deps == 'lowest' && ' (lowest)' || '' }}${{ matrix.use-dama == 1 && contains(matrix.database, 'sql') && ' (dama)' || '' }}${{ !contains(matrix.database, 'sql') && '' || matrix.use-migrate == 1 && ' (migrate)' || ' (schema)' }}${{ matrix.use-phpunit-extension == 1 && ' (phpunit extension)' || '' }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -19,6 +19,7 @@ jobs:
         database: [ mysql, mongo ]
         use-dama: [ 1 ]
         use-migrate: [ 0 ]
+        use-phpunit-extension: [ 0 ]
         phpunit: [ 9 ]
         exclude:
           - php: 8.1
@@ -32,6 +33,7 @@ jobs:
             database: none
             use-dama: 1
             use-migrate: 0
+            use-phpunit-extension: 0
             phpunit: 9
           - php: 8.3
             deps: highest
@@ -39,6 +41,7 @@ jobs:
             database: mysql|mongo
             use-dama: 1
             use-migrate: 0
+            use-phpunit-extension: 0
             phpunit: 9
           - php: 8.3
             deps: highest
@@ -46,6 +49,7 @@ jobs:
             database: pgsql|mongo
             use-dama: 1
             use-migrate: 0
+            use-phpunit-extension: 0
             phpunit: 9
           - php: 8.3
             deps: highest
@@ -53,6 +57,7 @@ jobs:
             database: pgsql
             use-dama: 0
             use-migrate: 0
+            use-phpunit-extension: 0
             phpunit: 9
           - php: 8.3
             deps: highest
@@ -60,6 +65,7 @@ jobs:
             database: sqlite
             use-dama: 0
             use-migrate: 0
+            use-phpunit-extension: 0
             phpunit: 9
           - php: 8.3
             deps: lowest
@@ -67,6 +73,7 @@ jobs:
             database: sqlite
             use-dama: 0
             use-migrate: 0
+            use-phpunit-extension: 0
             phpunit: 9
           - php: 8.3
             deps: lowest
@@ -74,6 +81,7 @@ jobs:
             database: mysql
             use-dama: 1
             use-migrate: 0
+            use-phpunit-extension: 0
             phpunit: 9
           - php: 8.3
             deps: highest
@@ -81,6 +89,7 @@ jobs:
             database: mysql
             use-dama: 1
             use-migrate: 1
+            use-phpunit-extension: 0
             phpunit: 9
           - php: 8.3
             deps: highest
@@ -88,6 +97,7 @@ jobs:
             database: mysql|mongo
             use-dama: 1
             use-migrate: 0
+            use-phpunit-extension: 0
             phpunit: 10
           - php: 8.3
             deps: highest
@@ -95,11 +105,21 @@ jobs:
             database: mysql|mongo
             use-dama: 1
             use-migrate: 0
+            use-phpunit-extension: 0
+            phpunit: 11
+          - php: 8.3
+            deps: highest
+            symfony: '*'
+            database: mysql|mongo
+            use-dama: 1
+            use-migrate: 0
+            use-phpunit-extension: 1
             phpunit: 11
     env:
       DATABASE_URL: ${{ contains(matrix.database, 'mysql') && 'mysql://root:root@localhost:3306/foundry?serverVersion=5.7.42' || contains(matrix.database, 'pgsql') && 'postgresql://root:root@localhost:5432/foundry?serverVersion=15' || contains(matrix.database, 'sqlite') && 'sqlite:///%kernel.project_dir%/var/data.db' || '' }}
       MONGO_URL: ${{ contains(matrix.database, 'mongo') && 'mongodb://127.0.0.1:27017/dbName?compressors=disabled&amp;gssapiServiceName=mongodb' || '' }}
       USE_DAMA_DOCTRINE_TEST_BUNDLE: ${{ matrix.use-dama == 1 && contains(matrix.database, 'sql') && 1 || 0 }}
+      USE_FOUNDRY_PHPUNIT_EXTENSION: ${{ matrix.use-phpunit-extension }}
       PHPUNIT_VERSION: ${{ matrix.phpunit }}
     services:
       postgres:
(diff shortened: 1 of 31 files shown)


src/Test/Factories.php
```diff
@@ -26,7 +29,16 @@ trait Factories
      * @before
      */
     #[Before]
-    public static function _bootFoundry(): void
+    public function _beforeHook(): void
+    {
+        $this->_bootFoundry();
+        $this->_loadDataProvidedProxies();
+    }
```
nikophil, line 35: I had an issue with the order of execution, so I grouped the two methods under the same hook

this is actually what made me create [this PR](https://github.com/sebastianbergmann/phpunit/pull/5890)https://github.com/sebastianbergmann/phpunit/pull/5890

src/ZenstruckFoundryBundle.php
```diff
@@ -244,6 +244,11 @@ public function process(ContainerBuilder $container): void
         }
     }
 
+    public function shutdown(): void
+    {
+        Configuration::shutdown();
+    }
```
nikophil, line 250: prevents some problems where we boot Symfony's kernel before the tests: foundry was keeping its "booted" stated, wheras it should have been shutdown

src/Persistence/PersistentProxyObjectFactory.php
```diff
@@ -36,6 +37,18 @@
...
+    /**
+     * @return T&Proxy<T>
+     */
+    final public function create(callable|array $attributes = []): object
+    {
+        try {
+            return parent::create($attributes);
+        } catch (FoundryNotBooted) {
+            return ProxyGenerator::wrapFactory($this, $attributes);
+        }
```
nikophil, line 49: here is the magic :magic_wand: :tada:

tests/bootstrap.php
```diff
@@ -49,6 +50,4 @@
     }
 
     $kernel->shutdown();
-
-    $fs->remove(__DIR__.'/../var');
```
nikophil, line 53: I really don't know why, but this was creating a problem in the migration tests:
https://github.com/zenstruck/foundry/actions/runs/9610817264/job/26508088592#step:6:35

tests/bootstrap.php
```diff
@@ -50,5 +51,5 @@
 
     $kernel->shutdown();
 
-    $fs->remove(__DIR__.'/../var');
+    Configuration::shutdown();
```
nikophil, line 54: This was needed otherwise in migrate mode, Foundry stays "booted" and did not throw `NotBootedException` in `PersistentProxyObjectFactory`

phpunit-10.xml.dist
```diff
@@ -19,5 +19,6 @@
   </source>
   <extensions>
     <bootstrap class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension" />
+    <bootstrap class="Zenstruck\Foundry\PHPUnit\Extension" />
```
nikophil: ~I'm planing to create a new xml file (sadly): `phpunit-11.4-with-extension.xml`~

~BTW, we need to check if this could not be added dynamically (so we won't have multiple phpunit.xml, beside of 9/10 versions)~

EDIT: see #683

src/PHPUnit/BootFoundryOnDataProviderMethodCalled.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function notify(Event\Test\DataProviderMethodCalled $event): void
+    {
+        if (is_a($event->testMethod()->className(), KernelTestCase::class, allow_string: true)) {
+            static $kernelIsBooted = false;
+
+            if (!$kernelIsBooted) {
+                $this->kernel->boot();
+                $kernelIsBooted = true;
+            }
```
nikophil, line 33: thanks to this, we can use service factories in data providers.

One downside of this, is that we're kinda always boot the kernel, even for tests like `vendor/bin/phpunit --filter SomeUnitTest`, because all data providers are always called. This may result in a longer bootstrap phase.

I'm wondering if we should not enable this in Foundry's config, WDYT?

src/Configuration.php
```diff
@@ -66,23 +67,29 @@ public function assertPersistanceEnabled(): void
...
+
     public static function instance(): self
     {
         if (!self::$instance) {
-            throw new FoundryNotBooted('Foundry is not yet booted. Ensure ZenstruckFoundryBundle is enabled. If in a test, ensure your TestCase has the Factories trait.');
+            throw new FoundryNotBooted();
         }
 
-        return \is_callable(self::$instance) ? (self::$instance)() : self::$instance;
+        return self::$instance;
```
nikophil, line 81: it felt cumbersome to call the callback on each call, wasn't it?

.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
         database: [ mysql, mongo ]
         use-dama: [ 1 ]
         use-migrate: [ 0 ]
+        use-phpunit-extension: [ 0 ]
```
nikophil, line 22: in a further PR I'll simplify this whole matrix:
 - Symfony 7.0 is not maintained anymore
 - I think we can use a default value for `use-dama`, `use-migrate`, `phpunit` and `use-phpunit-extension`, this will prevent the need to specify them in every permutation

tests/Fixture/TestKernel.php
```diff
@@ -101,6 +101,7 @@ protected function configureContainer(ContainerBuilder $c, LoaderInterface $load
                             'alias' => 'Model',
                         ],
                     ],
+                    'controller_resolver' => ['auto_mapping' => true],
```
nikophil, line 104: this was throwing a deprecation

src/Test/PHPUnit/BootFoundryOnDataProviderMethodCalled.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function notify(Event\Test\DataProviderMethodCalled $event): void
+    {
+        if (\is_a($event->testMethod()->className(), KernelTestCase::class, allow_string: true)) {
+            static $kernelIsBooted = false;
+
+            if (!$kernelIsBooted) {
+                $this->kernel->boot();
+                $kernelIsBooted = true;
+            }
```
nikophil, line 40: thanks to this, we can use service factories in data providers.

One downside of this, is that we're kinda always boot the kernel, even for tests like `vendor/bin/phpunit --filter SomeUnitTest`, because all data providers are always called. This may result in a longer bootstrap phase.

I'm wondering if we should not enable this in Foundry's config, WDYT?

tests/Integration/DataProvider/DataProviderForServiceFactoryInKernelTestCaseTest.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->assertSame('router-constructor', $providedData->getProp1());
+    }
+
+    public static function createObjectFromServiceFactoryInDataProvider(): iterable
+    {
+        if ('1' !== ($_ENV['USE_FOUNDRY_PHPUNIT_EXTENSION'] ?? null)) {
+            yield \FOUNDRY_SKIP_DATA_PROVIDER => [null];
+
+            return;
+        }
```
nikophil, line 54: this is the unique way to skip these kind of test

see https://github.com/sebastianbergmann/phpunit/issues/5926 which would be helpful for this

src/Test/Factories.php
```diff
@@ -26,7 +28,48 @@ trait Factories
...
+        }
+
+        static::$class = null; // @phpstan-ignore staticProperty.notFound
+        static::$kernel = null; // @phpstan-ignore staticProperty.notFound
+        static::$booted = false; // @phpstan-ignore staticProperty.notFound
+
+        // integration test
+        Configuration::bootForDataProvider(
+            static fn() => static::getContainer()->get('.zenstruck_foundry.configuration') // @phpstan-ignore staticMethod.notFound
+        );
```
nikophil: there is one downside of this: if the kernel has never been booted, the first any test will boot the kernel, thus, there will a waiting of few seconds:

```term
$ rm -rf var/cache/*

# phpunit calls all dataprovider, thus it will warmup the kernel
$ phpunit --filter SomeUnitTest

# a second call would be fast
$ phpunit --filter SomeUnitTest
```

I think this is acceptable. But maybe we can make the feature "allow service factory to be usable i dataproviders" optional. (or we can wait someone complains :innocent: )

src/Persistence/ProxyGenerator.php
```diff
@@ -43,6 +46,19 @@ public static function wrap(object $object): Proxy
...
+     * @template T of object
+     *
+     * @param PersistentProxyObjectFactory<T> $factory
+     * @phpstan-param Attributes $attributes
+     *
+     * @return T&Proxy<T>
+     */
+    public static function wrapFactory(PersistentProxyObjectFactory $factory, callable|array $attributes): Proxy
+    {
+        return self::generateClassFor($factory)::createLazyProxy(static fn() => unproxy($factory->create($attributes))); // @phpstan-ignore-line
```
nikophil, line 59: I'm using "our" proxy, but I think I'll directly use `ProxyHelper::generateLazyProxy()`. This way I think I'll find a way to open this feature to all kind of factories (and not only proxy ones)

kbond (approved):
This is really nice! ❤️

PR #659, opened 2024-06-28
fix: can create inversed one to one

### LINKED ISSUE #655: [2.0] Problem with inversed `OneToOne` relationships
Hello there,

I'm upgrading a project from 1.38 to 2.0. With the Rector rule, the upgrade is really smooth, thanks @nikophil !

I just have a minor issue, in one Factory, in the `defaults()` method, I'm doing something like this :

```
{
    return [
        'customer' => CustomerFactory::new()->withUser(),
    ];
}
```

And in Customer factory :
```
public function withUser(): self
{
    return $this->with(static fn (): array => ['user' => UserFactory::new()]);
}
```

With Foundry 1.38, my Customer was containing an User, since 2.0 User is always null.
Is it something deprecated or a bug ?

nikophil:
fixes #655

### DIFF
--- a/src/ORM/OrmV2PersistenceStrategy.php
+++ b/src/ORM/OrmV2PersistenceStrategy.php
@@ -31,27 +31,47 @@ final class OrmV2PersistenceStrategy extends AbstractORMPersistenceStrategy
 
         $association = $this->getAssociationMapping($parent, $field);
 
-        if (null === $association) {
-            $inversedAssociation = $this->getAssociationMapping($child, $field);
-
-            if (null === $inversedAssociation || !$metadata instanceof ClassMetadataInfo) {
-                return null;
-            }
-
-            if (!\is_a($parent, $inversedAssociation['targetEntity'], allow_string: true)) { // is_a() handles inheritance as well
-                throw new \LogicException("Cannot find correct association named \"{$field}\" between classes [parent: \"{$parent}\", child: \"{$child}\"]");
-            }
-
-            if (ClassMetadataInfo::ONE_TO_MANY !== $inversedAssociation['type'] || !isset($inversedAssociation['mappedBy'])) {
-                return null;
-            }
-
-            $association = $metadata->getAssociationMapping($inversedAssociation['mappedBy']);
+        if ($association) {
+            return new RelationshipMetadata(
+                isCascadePersist: $association['isCascadePersist'],
+                inverseField: $metadata->isSingleValuedAssociation($association['fieldName']) ? $association['fieldName'] : null,
+                isCollection: $metadata->isCollectionValuedAssociation($association['fieldName']),
+            );
         }
 
+        $inversedAssociation = $this->getAssociationMapping($child, $field);
+
+        if (null === $inversedAssociation || !$metadata instanceof ClassMetadataInfo) {
+            return null;
+        }
+
+        if (!\is_a(
+            $parent,
+            $inversedAssociation['targetEntity'],
+            allow_string: true
+        )) { // is_a() handles inheritance as well
+            throw new \LogicException(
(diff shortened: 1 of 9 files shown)


src/ORM/OrmV3PersistenceStrategy.php
```diff
@@ -47,8 +48,9 @@ public function relationshipMetadata(string $parent, string $child, string $fiel
         }
 
         return new RelationshipMetadata(
-            isCascadePersist: $association->isCascadePersist(),
+            isCascadePersist: ($inversedAssociation ?? $association)->isCascadePersist(),
             inverseField: $metadata->isSingleValuedAssociation($association->fieldName) ? $association->fieldName : null,
+            isCollection: ($inversedAssociation ?? $association) instanceof ToManyAssociationMapping
```
nikophil: I'm not sure about those... but we want to know if the "inversed" association is a collection here, if it exsts. It's the only way to know that we are handling an "inverse one to one"

src/Persistence/PersistentObjectFactory.php
```diff
@@ -275,8 +275,25 @@ protected function normalizeParameter(string $field, mixed $value): mixed
...
+
+            // handle inversed OneToOne
+            if ($relationshipMetadata && !$relationshipMetadata->isCollection && $inverseField = $relationshipMetadata->inverseField) {
+                $this->tempAfterPersist[] = static function(object $object) use ($value, $inverseField, $pm) {
+                    $value->create([$inverseField => $object]);
+                    $pm->refresh($object);
+                };
+
+                // creation delegated to afterPersist hook - return empty array here
+                return null;
```
nikophil, line 280: I'm wondering if it is OK to return null here.
Can a "OneToOne" relationship be "not nullable" on the both sides of the relationship?

EDIT: I think it can:
```php
#[Entity]
class Contact
{
    #[ORM\OneToOne(targetEntity: Address::class, inversedBy: 'contact')]
    public Address $address;

    public function __construct()
    {
        $this->address = new Address($this);
    }
}

#[Entity]
class Address
{
    public function __construct(
	    #[ORM\OneToOne(targetEntity: Contact::class, mappedBy: 'address')]
	    public Contact $contact;
    ) {}
}
```
and then, it would not be valid to return `null` here :shrug:

src/ORM/OrmV2PersistenceStrategy.php
```diff
@@ -52,6 +52,7 @@ public function relationshipMetadata(string $parent, string $child, string $fiel
         return new RelationshipMetadata(
             isCascadePersist: $association['isCascadePersist'],
             inverseField: $metadata->isSingleValuedAssociation($association['fieldName']) ? $association['fieldName'] : null,
+            isCollection: $metadata->isCollectionValuedAssociation($association['fieldName']),
```
nikophil: need to fix this, to mimic how it is done with orm v3

src/Persistence/PersistentObjectFactory.php
nikophil: I think this very specific case, where both side of `OneToOne` are mandatory, could not be fixed : it is a circular dependency. Maybe we could find a way to emit a meaningful error for the user?

Also, maybe we could find a way to return something else than `null` here: in Foundry2, we used to return a `PostPersistCallback` which was not added to the normalized parameters. Maybe we could mimic this (but it would complexify the whole normalization phase)

BTW, I just figured out that you did notice this case was error prone: https://github.com/zenstruck/foundry/blob/2.x/tests/Integration/ORM/EntityFactoryRelationshipTestCase.php#L183

kbond: Yes, I don't remember the exact context of what happened but I think you're right, it's the circular dependency issue.

kbond (approved):

PR #666, opened 2024-07-03
tests: automatically create cascade persist permutations

### LINKED ISSUE #597: [2.x] Automatically test all "cascade" combinations in tests
I think we could get rid of `CascadeEntity...` classes in tests by using Doctrine's event [`loadClassMetadata`](https://www.doctrine-project.org/projects/doctrine-orm/en/3.1/reference/events.html#reference-events-load-class-metadata)

Somehow, we could automatically hook with this even and generate all "cascade" combinations and run tests against it.

I'm not sure if this is possible, but I think it could lower the complexity of the tests. Let's experiment :microscope: :scientist:

nikophil:
fixes https://github.com/zenstruck/foundry/issues/597

This PR introduces a way to automatically create all permutations for "cascade persist" relationships.

By using a combination of the trait `WithEntityRelationShip`, and the attribute `UsingRelationShips`, a data provider is automatically created with all possible combinations of the "cascade persist" and "standard" associations.

ex:
```php
final class EntityRelationshipTest extends KernelTestCase
{
    use WithEntityRelationShip, Factories, ResetDatabase;

    // a matrix with all relationships will be created. Here, for example, there will be 4 permutations:
    // [`Contact::$category` => "cascade", `Category::$contacts` => "cascade"]
    // [`Contact::$category` => "standard", `Category::$contacts` => "cascade"]
    // [`Contact::$category` => "cascade", `Category::$contacts` => "standard"]
    // [`Contact::$category` => "standard", `Category::$contacts` => "standard"]
    #[UsingRelationShips(Contact::class, ['category'])]
    #[UsingRelationShips(Category::class, ['contacts'])]
    /**
     * @test
     * @dataProvider provideCascadeRelationshipsCombination
     */
    public function many_to_one(): void
    {
        // add some test which used `Contact::$category` and `Category::$contacts`
    }
}
```

Here are some thoughts about this work:
- We must add the data provider on each method using the attribute `UsingRelationShips`. I don't think it is possible to automatically add it, but I've added a check which enforces to add the provider if the attribute is used
- It is currently not possible to have another data provider. But for now, the tests about "orm relationships" do not use data providers, so it's good enough
- I do think we should test ALL these permutations (I mean, the full matrix, with full cases), we may find some weird edge cases
- sadly, I had to introduce a (really tiny) PHPUnit extension for this kind of test: it is used to be able to access current method name within the data provider. That's the only way I've found to do it


tests/Fixture/DoctrineCascadeRelationship/WithEntityRelationShip.php
```diff
@@ -0,0 +1,108 @@
...
+        $result = [];
+        $total = pow(2, count($permutations));
+
+        for ($i = 0; $i < $total; $i++) {
+            $temp = [];
+            for ($j = 0; $j < count($permutations); $j++) {
+                $temp[$j] = new DoctrineCascadeRelationshipMetadata(
+                    class: $permutations[$j]['class'],
+                    field: $permutations[$j]['field'],
+                    cascade: (bool) (($i >> $j) & 1)
```
nikophil: thank you chatGPT :sweat_smile:

tests/Integration/ORM/EntityRelationshipTest.php
```diff
@@ -0,0 +1,45 @@
...
+{
+    use WithEntityRelationShip, Factories, ResetDatabase;
+
+    /**
+     * @test
+     * @dataProvider provideCascadeRelationshipsCombination
+     */
+    #[UsingRelationShips(StandardContact::class, ['category'])]
+    #[UsingRelationShips(StandardCategory::class, ['contacts'])]
+    public function many_to_one(): void
```
nikophil: I'm not sure for this specific test if the inverse relationship is even used :thinking:

tests/Fixture/DoctrineCascadeRelationship/WithEntityRelationShip.php
```diff
@@ -0,0 +1,108 @@
...
+            $changeCascadePersistListener->withMetadata($this->getProvidedData());
+        } else {
+            throw new \LogicException(sprintf('When using attribute "%s", you must use "provideCascadeRelationshipsCombination" as a data provider.', UsingRelationShips::class));
+        }
+    }
+
+    /**
+     * @return iterable<list<DoctrineCascadeRelationshipMetadata>>
+     */
+    private static function provideCascadeRelationshipsCombination(string $methodName): iterable
```
nikophil: I need to add some tests for this stuff as well

tests/Fixture/DoctrineCascadeRelationship/UsingRelationShips.php
```diff
@@ -0,0 +1,20 @@
...
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Tests\Fixture\DoctrineCascadeRelationship;
+
+/**
+ * @author Nicolas PHILIPPE <nikophil@gmail.com>
+ */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+final class UsingRelationShips
```
kbond:
```suggestion
final class UsingRelationships
```

tests/Fixture/DoctrineCascadeRelationship/WithEntityRelationShip.php
```diff
@@ -0,0 +1,108 @@
...
+use Doctrine\ORM\Mapping\MappingException;
+use PHPUnit\Framework\Attributes\Before;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Zenstruck\Foundry\Persistence\PersistenceManager;
+use Zenstruck\Foundry\Tests\Integration\RequiresORM;
+
+/**
+ * @author Nicolas PHILIPPE <nikophil@gmail.com>
+ */
+trait WithEntityRelationShip
```
kbond:
```suggestion
trait WithEntityRelationship
```

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         # deps: [ highest ]
         # without-dama: [ 0 ]
         # use-phpunit-extension: [ 0 ]
-        # phpunit: [ 9 ]
+        # phpunit: [ 11 ]
```
nikophil, line 24: I think that from now, PHPUnit 11 should be the default phpunit version on which we're running tests

tests/Fixture/DoctrineCascadeRelationship/ChangesEntityRelationshipCascadePersist.php
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * @return iterable<list<DoctrineCascadeRelationshipMetadata>>
+     */
+    public static function provideCascadeRelationshipsCombinations(): iterable
+    {
+        if (!\getenv('DATABASE_URL')) {
+            // this test requires the ORM, but trait RequiresORM is analysed after data provider are called
+            // then we need to return at least one empty array to avoid an error
+            // in PHPUnit 12, we will be able to use #[RequiresEnvironmentVariable('DATABASE_URL')] to prevent this
```
nikophil, line 69: https://github.com/sebastianbergmann/phpunit/pull/6074  :blush:

.github/workflows/ci.yml
```diff
@@ -8,42 +8,42 @@ on:
...
-    name: P:${{ matrix.php }}, S:${{ matrix.symfony }}, D:${{ matrix.database }}, PU:${{ matrix.phpunit || 9 }}${{ matrix.deps == 'lowest' && ' (lowest)' || '' }}${{ matrix.without-dama == 0 && contains(matrix.database, 'sql') && ' (dama)' || '' }}${{ matrix.use-phpunit-extension == 1 && ' (phpunit extension)' || '' }}
+    name: P:${{ matrix.php }}, S:${{ matrix.symfony }}, D:${{ matrix.database }}, PU:${{ matrix.phpunit || 11 }}${{ matrix.deps == 'lowest' && ' (lowest)' || '' }}${{ matrix.without-dama == 0 && contains(matrix.database, 'sql') && ' (dama)' || '' }}${{ matrix.use-phpunit-extension == 1 && ' (phpunit extension)' || '' }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         php: [ 8.1, 8.2, 8.3, 8.4 ]
         symfony: [ 6.4.*, 7.1.*, 7.2.* ]
         database: [ mysql, mongo ]
+        phpunit: [ 9, 11 ]
```
nikophil, line 19: since all relationships combinations are ran only with PHPUnit 11, I'd rather having this version to be first class citizen of the CI

kbond (approved):

PR #671, opened 2024-07-05
Fix faker php urls

BackEndTea:
Faker now lives on fakerphp.org instead of fakerphp.github.io.

The github domain also has a weird issue where it redirects to the http page of faker rather than https

See https://github.com/FakerPHP/fakerphp.github.io/issues/100

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -494,7 +494,7 @@ Sequences help to create different objects in one call:
 Faker
 ~~~~~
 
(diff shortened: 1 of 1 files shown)


nikophil (approved):

nikophil:
Thanks @BackEndTea

BackEndTea:
Thank you @nikophil !

PR #672, opened 2024-07-09
fix: handle proxies when refreshing entity in `Proxy::getState()`

### LINKED ISSUE #649: Global state relationships and flush_after
Hello,

I recently migrated to version 2.

I used to use `Factory::flushAfter` in my functional tests.

Using DAMA + global stories, when I create a object in a story (not global this one) using a relationship created in a global story I got a `A new entity was found through the relationship` error only if story is load with a `flush_after`.
I use `flush_after` because in my case, two entities has to created in the same transaction.

My config file : 
```php
<?php

declare(strict_types = 1);

use ...;

return static function (ZenstruckFoundryConfig $config, ContainerConfigurator $container): void {
    $config->faker()->locale('fr_FR');
    $config->globalState([
        GlobalState\ActivityStory::class,
		...,
        GlobalState\BusinessProviderStory::class,
        GlobalState\BusinessProviderOfferStory::class,
    ]);
    $config->orm(['auto_persist' => true, 'reset' => ['connections' => ['default'], 'entity_managers' => ['default'], 'mode' => 'migrate']]);
};
```

My call in test : `flush_after(static fn () => CreditNoteStory::load());`

If I update my business logic (ignoring error on missing linked entity in the same transaction), and remove the `flush_after` call, I haven't any error

Thank you for your help

nikophil:
fixes #649

### DIFF
--- a/src/Story.php
+++ b/src/Story.php
@@ -14,6 +14,8 @@ namespace Zenstruck\Foundry;
 use Zenstruck\Foundry\Exception\PersistenceNotAvailable;
 use Zenstruck\Foundry\Persistence\Exception\NoPersistenceStrategy;
 use Zenstruck\Foundry\Persistence\Exception\RefreshObjectFailed;
+use Zenstruck\Foundry\Persistence\Proxy;
+use Zenstruck\Foundry\Persistence\ProxyGenerator;
 
 /**
  * @author Kevin Bond <kevinbond@gmail.com>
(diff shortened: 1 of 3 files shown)


src/Story.php
```diff
@@ -132,7 +134,12 @@ final protected function getState(string $name): mixed
         }
 
         try {
-            return Configuration::instance()->persistence()->refresh($this->state[$name], force: true); // @phpstan-ignore argument.templateType
+            $isProxy = $this->state[$name] instanceof Proxy;
+
+            $unwrappedObject = ProxyGenerator::unwrap($this->state[$name]);
+            Configuration::instance()->persistence()->refresh($unwrappedObject, force: true); // @phpstan-ignore argument.templateType
+
+            return $isProxy ? ProxyGenerator::wrap($unwrappedObject) : $unwrappedObject; // @phpstan-ignore argument.templateType
```
nikophil, line 142: I did not find anything simpler/cleaner that passes the test :shrug:

kbond (approved):

Issue #656, opened 2024-06-27
[2.0] No autocompletion for Proxy objects


simondaigre:
With 2.0, I don't have autocompletion for Proxy methods (_real, etc ...) on proxied objects.

Using this Phpdoc : 
```
/**
 * @extends PersistentProxyObjectFactory<User>
 */
```

Aerendir:
It works for me in PhpStorm: did you added the `use` statement?

```php
use Zenstruck\Foundry\Persistence\PersistentProxyObjectFactory;
use Zenstruck\Foundry\Persistence\Proxy;

/**
 * @extends PersistentProxyObjectFactory<Receipt>
 *
...

simondaigre:
Yes I have the use statement.

PR #674, opened 2024-07-18
minor: add missing generic return phpdoc

ekvedaras:
This was driving me nuts.

Please let me know if I missed some reason why this should not be here :)

### DIFF
--- a/src/ObjectFactory.php
+++ b/src/ObjectFactory.php
@@ -41,6 +41,8 @@ abstract class ObjectFactory extends Factory
 
     /**
      * @final
+     *
+     * @return T
      */
     public function create(callable|array $attributes = []): object
     {
--- a/src/Persistence/PersistentObjectFactory.php
+++ b/src/Persistence/PersistentObjectFactory.php
@@ -203,6 +203,9 @@ abstract class PersistentObjectFactory extends ObjectFactory
         static::repository()->truncate();
     }
 
+    /**
+     * @return T
+     */
     final public function create(callable|array $attributes = []): object
     {
         $object = parent::create($attributes);


nikophil:
Hi @ekvedaras 

thanks for this.

just to know, is it improving your IDE's autocompletion?

because static analysis tools are inferring the return type from the parent method `Factory::create()` where the return type is actually present.

ekvedaras:
Yes, this is for IDE. `Factory::createOne()` is understood fine, but `Factory::new()->create()` looses the type without this.

ekvedaras:
Although now I'm testing this on actual project instead of this repo only and it is still getting confused and thinks it returns a proxy 🤔 Not wrong, but doesn't understand the entity

ekvedaras:
Hmm, can't make it to work on an actual project. Don't really see what's wrong now. I know it is possible as I've built similar factory thing myself, however it was simpler. So either phpstorm is getting lost somewhere in the code or symfony plugin is interfering (don't know if it does anything with foundry).

I will play around a bit more later. If I can't make it work, will close.

nikophil:
yeah, autocomplete for proxies is broken :/

https://github.com/zenstruck/foundry/issues/656

The whole "type system" is fine, PHPStan is tested and OK, but somehow PHPStorm does not undderstand something. My first guess was to add a bunch of `@method` to `PersistentProxyObjectFactory`, but I failed to make it work... I have to dig a little bit more in this way...

ekvedaras:
Probably has something to do with depth. If I for example extend `Factory<User1>` `create()` will work. Anything else like `ObjectFactory<User1>` won't anymore unless it is a static call like `createOne()`. Might simply be a bug/limitation of phpstorm.

nikophil:
see https://github.com/zenstruck/foundry/pull/675 where the solution is wider

PR #675, opened 2024-07-21
fix: add phpdoc to improve proxy factories autocompletion

nikophil:
fixes https://github.com/zenstruck/foundry/issues/656

ping @ekvedaras (I'm closing your PR in favor of this one) / @simondaigre

PR #676, opened 2024-07-22
Improve return types of Factory

andersonamuller:
At least PHPStorm was not able to autocomplete without these changes.

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -36,6 +36,8 @@ abstract class Factory
 
     /**
      * @param Attributes $attributes
+     *
+     * @return static<T>
      */
     final public static function new(array|callable $attributes = []): static
     {
@@ -130,6 +132,8 @@ abstract class Factory
 
     /**
      * @param Attributes $attributes
+     *
+     * @return static<T>
      */
     final public function with(array|callable $attributes = []): static
     {
@@ -173,7 +177,7 @@ abstract class Factory
     /**
      * Override to adjust default attributes & config.
      *
-     * @return static
+     * @return static<T>
      */
     protected function initialize(): static
     {


nikophil:
Hi @andersonamuller 

thanks for this PR.

could you show me which kind of autocompletion problem this PR fixes?

those return types create errors in static analysis, but I'm wondering if it is not a false positive...

nikophil:
https://phpstan.org/r/36d05c9d-1309-4a9a-87b2-91235c23a796  :shrug:

andersonamuller:
Sorry @nikophil. I haven't digged deep enough. The issue is with the PHPStorm static analyzer.

When I try something like:
```php
$object = MyFactory::new()->with([...])->create();
```

The autocomplete for the `$object` only shows methods of the `\Doctrine\Persistence\ObjectRepository\Proxy` class.

When I use 
```php
$object = MyFactory::createOne([...]);
```
Then is all fine. Or with the patch I submitted then PHPStorm is able to resolve the type.

nikophil:
Would you mind to test this PR?
https://github.com/zenstruck/foundry/pull/675

Everything is fine for me with this fix

andersonamuller:
I tested that PR and unfortunately it doesn't help with my case. I had to add the method annotation to my factory class.

```php
/**
 * @extends PersistentObjectFactory<MyObject>
 * 
 * @method MyObject create()
 */
final class MyFactory extends PersistentObjectFactory {}
```

nikophil:
ok thanks, I'll try using `with()` and check what's going on. I'd really like end factory to not need these `@method` annotations

nikophil:
hey @andersonamuller 

may I ask what's your PHPStorm version? in all my tests have the same result:
```php
// 
MyFactory::new()->create();
```

PR #681, opened 2024-08-03
[Doc] Fix CategoryStory codeblock

smnandre:
Codeblock was not nested in the tip and was badly rendered


nikophil:
thanks @smnandre

PR #683, opened 2024-08-07
chore: add a wrapper for PHPUnit binary

nikophil:
I think if we introduce a PHPUnit extension (see #648), it will be harder to test all permutations locally, and to mix dama, the new extension, phpunit version...

Here is a proposal of a wrapper for PHPUnit, which will add the right PHPUnit extension, and automatically upgrade to the right version, based on environment variables:
- `PHPUNIT_VERSION`
- `USE_DAMA_DOCTRINE_TEST_BUNDLE`
- `USE_FOUNDRY_PHPUNIT_EXTENSION` (soon to come)

then, in order to run the test, we just have to run `./phpunit`, it will forward to PHPUnit all options or arguments

Would you accept this? I'll add some docs in the readme about this as well


.github/workflows/ci.yml
```diff
@@ -242,27 +218,6 @@ jobs:
         env:
           DOCS_DIR: 'docs/'
 
-  build-docs:
```
kbond, line 245: This is no longer needed?

kbond (commented):
I like this - especially how it simplifies the CI.

.github/workflows/ci.yml
nikophil: Yes, since we have doctor-rst

kbond (approved):

smnandre:
Hi @nikophil 

Do you plan to release / document this for everyone or is it "just" an internal tool ?

nikophil:
Hi @smnandre 

I'm not sure to understand your question 😅 

It is indeed a very internal tool for Foundry development, which does not aim to be shipped (this makes me think that I've forgotten to add it to `.gitattributes`).
What would you use it for?

PR #686, opened 2024-08-09
feat: allow to configure migrations configuration files

### LINKED ISSUE #685: Allow to reset many databases in migrate mode
When `orm.reset.mode` is `migrate` , the `ORMDatabaseResetter` will simply run `doctrine:migrations:migrate`. This makes impossible to use this mode when you have migrations for many dtabases.

Could this library provide a `configurations` option so that each of its values is passed to `doctrine:migrations:migrate`’s `--configuration` option?

MatTheCat:
This PR introduces a new `orm.reset.migrations.configurations` configuration node allowing a schema reset to run `doctrine:migrations:migrate` for each given configuration. This is useful because using `doctrine/migrations`, you can only target one single connection or entity manager per configuration file.

Fixes #685

### DIFF
--- a/src/ORM/AbstractORMPersistenceStrategy.php
+++ b/src/ORM/AbstractORMPersistenceStrategy.php
@@ -152,7 +152,18 @@ abstract class AbstractORMPersistenceStrategy extends PersistenceStrategy
 
     private function createSchema(Application $application): void
     {
-        if (self::RESET_MODE_MIGRATE === $this->config['reset']['mode']) {
+        if (self::RESET_MODE_SCHEMA === $this->config['reset']['mode']) {
+            foreach ($this->managers() as $manager) {
+                self::runCommand($application, 'doctrine:schema:update', [
+                    '--em' => $manager,
+                    '--force' => true,
+                ]);
+            }
+
+            return;
+        }
+
+        if (!$migrationsConfigurations = $this->config['reset']['migrations']['configurations']) {
             self::runCommand($application, 'doctrine:migrations:migrate', [
                 '--no-interaction' => true,
             ]);
@@ -160,10 +171,10 @@ abstract class AbstractORMPersistenceStrategy extends PersistenceStrategy
             return;
         }
 
-        foreach ($this->managers() as $manager) {
-            self::runCommand($application, 'doctrine:schema:update', [
-                '--em' => $manager,
-                '--force' => true,
+        foreach ($migrationsConfigurations as $migrationsConfiguration) {
+            self::runCommand($application, 'doctrine:migrations:migrate', [
+                '--configuration' => $migrationsConfiguration,
+                '--no-interaction' => true,
             ]);
         }
     }
--- a/src/ZenstruckFoundryBundle.php
+++ b/src/ZenstruckFoundryBundle.php
@@ -114,6 +114,28 @@ final class ZenstruckFoundryBundle extends AbstractBundle implements CompilerPas
                                     ->defaultValue(AbstractORMPersistenceStrategy::RESET_MODE_SCHEMA)
                                     ->values([AbstractORMPersistenceStrategy::RESET_MODE_SCHEMA, AbstractORMPersistenceStrategy::RESET_MODE_MIGRATE])
                                 ->end()
+                                ->arrayNode('migrations')
+                                    ->addDefaultsIfNotSet()
+                                    ->children()
+                                        ->arrayNode('configurations')
+                                            ->info('Migration configurations')
+                                            ->defaultValue([])
+                                            ->scalarPrototype()->end()
+                                            ->validate()
+                                                ->ifTrue(function(array $configurationFiles): bool {
+                                                    foreach ($configurationFiles as $configurationFile) {
+                                                        if (!\is_file($configurationFile)) {
+                                                            return true;
+                                                        }
+                                                    }
+
+                                                    return false;
+                                                })
+                                                ->thenInvalid('At least one migrations configuration file does not exist.')
+                                            ->end()
+                                        ->end()
+                                    ->end()
+                                ->end()
                             ->end()
                         ->end()
                     ->end()


src/ORM/AbstractORMPersistenceStrategy.php
```diff
@@ -152,18 +152,29 @@ private function dropAndResetDatabase(Application $application): void
...
             ]);
 
             return;
         }
 
-        foreach ($this->managers() as $manager) {
-            self::runCommand($application, 'doctrine:schema:update', [
-                '--em' => $manager,
-                '--force' => true,
+        foreach ($migrationsConfigurations as $migrationsConfiguration) {
```
nikophil, line 174: What if no migration configuration is provided? I'm surprized that the tests still pass 🤔

src/ZenstruckFoundryBundle.php
```diff
@@ -114,6 +114,16 @@ public function configure(DefinitionConfigurator $definition): void
                                     ->defaultValue(AbstractORMPersistenceStrategy::RESET_MODE_SCHEMA)
                                     ->values([AbstractORMPersistenceStrategy::RESET_MODE_SCHEMA, AbstractORMPersistenceStrategy::RESET_MODE_MIGRATE])
                                 ->end()
+                                ->arrayNode(AbstractORMPersistenceStrategy::RESET_MODE_MIGRATE)
+                                    ->addDefaultsIfNotSet()
+                                    ->children()
+                                        ->arrayNode('configurations')
+                                            ->info('Migration configurations')
+                                            ->defaultValue([])
+                                            ->scalarPrototype()->end()
```
nikophil, line 123: Could you add a check on the existence of the files?

nikophil (changes requested):
Please add a test for this 🙏

src/ORM/AbstractORMPersistenceStrategy.php
MatTheCat: In that case `doctrine:migrations:migrate` is run as before without its `--configuration` option. I changed the conditions order to simplify the code so now it’s done in line 167.

src/ZenstruckFoundryBundle.php
MatTheCat: Done. Not sure if relative paths can be configured like this though :thinking:

nikophil: yes, I'm wondering if you should not concat with `kernel.project_dir` or something like this. But I'm wondering if this param is accessible here

src/ZenstruckFoundryBundle.php
```diff
@@ -114,6 +114,28 @@ public function configure(DefinitionConfigurator $definition): void
                                     ->defaultValue(AbstractORMPersistenceStrategy::RESET_MODE_SCHEMA)
                                     ->values([AbstractORMPersistenceStrategy::RESET_MODE_SCHEMA, AbstractORMPersistenceStrategy::RESET_MODE_MIGRATE])
                                 ->end()
+                                ->arrayNode(AbstractORMPersistenceStrategy::RESET_MODE_MIGRATE)
+                                    ->addDefaultsIfNotSet()
+                                    ->children()
+                                        ->arrayNode('configurations')
+                                            ->info('Migration configurations')
```
nikophil, line 121: maybe we could add a little bit more hints?
```suggestion
                                            ->info('Absolute paths to migration configuration files')
```

src/ZenstruckFoundryBundle.php
MatTheCat: Seems like parameters value have to be injected: https://symfony.com/doc/current/configuration/using_parameters_in_dic.html

Not sure how to do it in a bundle class though.

nikophil: I think you can just use `$container->getParameter('kernel.project_dir')`

nikophil: maybe we can change the strategy and require absolute paths?
and we'd accept `%kernel.project_dir%/path/to/conf.php` as well as `/aboslute/path/to/conf.php`

and then validate it at runtime (that's what is done in doctrine migration for conf `doctrine_migrations.migrations_paths` for instance)

MatTheCat: I was looking at e.g. Symfony translator’s `default_path` and there is no restriction on whether the path is absolute or relative, but as there is no reliable way to resolve a relative path the recipe simply defaults to an absolute one:

```yaml
framework:
    translator:
        default_path: '%kernel.project_dir%/translations'
```

That would be the simplest and perhaps more intuitive way :thinking:

nikophil: yeah I think we agree on this, let's not enforce anything, and just check `file_exist($migrationConfiguration)` at runtime.

We'll add an example using `%kernel.project_dir%` in the docs

MatTheCat: The check can still be done at compile time since parameters will be resolved when the `configurations` node validates.

Should I move it into `AbstractORMPersistenceStrategy::createSchema`?

nikophil: I'd say the sooner the better! if the params are resolved in the validation phase, let's keep it here

src/ZenstruckFoundryBundle.php
MatTheCat: I changed to `Migration configuration files` since we won’t check if paths are absolute or not; is that okay?

MatTheCat:
@nikophil not sure how to add a test; my attempt fails and I don’t understand why :thinking:

nikophil:
hum this error `Can't create database 'cms'; database exists` already drove me crazy few days ago, I need to deep dive into it... we're doing some weird stuff with migrations and with this "cms" schema.

Another thing is that I think we should not have a full test suite with "migration with configuration", or even for "migration" at all. But this goes way beyond the purpose of this PR.

What about this: please, remove all the modification about the CI (sorry for your wasted time around it), and let's merge your PR.  I've manually tested your branch on a project of mine and can confirm it is working. I will then work on a cleaner solution for this whole thing in the tests

MatTheCat:
No worries, I procrastinated a lot on this one..!

Reverted every test-related change :broom:

nikophil (approved):
LGTM

PR #688, opened 2024-08-23
doc: Fix range call instead of many

ternel:
`many()` seems to have only one parameter, see https://github.com/zenstruck/foundry/blob/2.x/src/Factory.php#L107 and https://github.com/zenstruck/foundry/blob/2.x/src/FactoryCollection.php#L39

So, if we want "between 0 and 10 comments", `range()` seems to be the good one: https://github.com/zenstruck/foundry/blob/2.x/src/FactoryCollection.php#L53


nikophil:
thanks @ternel

Issue #658, opened 2024-06-28
[2.0] ResetDatabase with DamaDoctrineTestBundle and symfony/doctrine-messenger library


maloy1984:
After migrate to version 2.0. 
My test was broken, because database not refresh before each test.
I have warning

```
Exception in third-party event subscriber: There is already an active transaction

#0 /var/www/vendor/doctrine/dbal/src/Driver/PDO/Connection.php(119): Doctrine\DBAL\Driver\PDO\PDOException::new(Object(PDOException))

#1 /var/www/vendor/dama/doctrine-test-bundle/src/DAMA/DoctrineTestBundle/Doctrine/DBAL/StaticDriver.php(60): Doctrine\DBAL\Driver\PDO\Connection->beginTransaction()

#2 /var/www/vendor/dama/doctrine-test-bundle/src/DAMA/DoctrineTestBundle/PHPUnit/PHPUnitExtension.php(57): DAMA\DoctrineTestBundle\Doctrine\DBAL\StaticDriver::beginTransaction()
```

My stack 

- Symfony 6.4
- Php 8.3 
- PhpUnit 11

After debug, I found problem (I think so)
In function 
`PersistenceManager::resetDatabase `

line 97 - 102
```

if ($isDAMADoctrineTestBundleEnabled) {
    // re-enable static connections
     StaticDriver::setKeepStaticConnections(true);
}

$shutdownKernel();
```
In function `$shutdownKernel` call `KernelTestCase::ensureKernelShutdown`
And in `KernelTestCase::ensureKernelShutdown` exist code 
```
if ($container instanceof ResetInterface) {
    $container->reset();
}
```
And if install library symfony/doctrine-messenger for doctrine transport
In this `reset` execute query 
`$this->executeStatement(sprintf('UNLISTEN "%s"', $this->configuration['table_name']));`

But static connection in  DAMADoctrineTestBundle already is true.

If in `PersistenceManager::resetDatabase ` first call `$shutdownKernel();` and then set static connection, all works fine.

Siniliote:
I have same problem.
I remove `symfony/doctrine-messenger` and go use RabbitMq ;)

nikophil:
hey, this has been fixed by https://github.com/zenstruck/foundry/pull/690 which will be released soon!

PR #690, opened 2024-08-25
refactor: make "database reset" mechanism extendable

### LINKED ISSUE #540: [Feature] Add a event before/after reset database (and schema creation)

nikophil:
This is an experiment to refactor "database reset" mechanism.

The goal is to decouple the code and to allow end users to extend it, thanks to decoration - I'll add an example in the docs on how to hook in the database reset mechanism

This work was inspired by [this comment](https://github.com/zenstruck/foundry/pull/552#issuecomment-1940801283), I think I also do prefer decoration over an event system in Foundry

fixes https://github.com/zenstruck/foundry/issues/540 https://github.com/zenstruck/foundry/issues/658

~(let's merge https://github.com/zenstruck/foundry/pull/686 before this one, I'll take care of the rebase)~ _done_


src/Persistence/ResetDatabase/ResetDatabaseManager.php
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * @param iterable<DatabaseResetterInterface> $databaseResetters
+     * @param iterable<SchemaResetterInterface> $schemaResetters
+     */
+    public function __construct(
+        private iterable $databaseResetters,
+        private iterable $schemaResetters
+    ) {
+    }
```
nikophil: I'm not 100% sure that we should split database / schema reset mechanisms, but since Mongo does not need any "database reset", this seemed the way to go

nikophil: I'm also not sure that we should keep "reset database/schema" naming, as it may have no sense for some DBMS 
WDYT of something like `beforeClassResetter` / `beforeTestResetter`?

src/ORM/ResetDatabase/OrmDatabaseResetterTrait.php
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * @internal
+ * @author Nicolas PHILIPPE <nikophil@gmail.com>
+ */
+trait OrmDatabaseResetterTrait
+{
+    protected function dropAndResetDatabase(Application $application): void
+    {
+        foreach ($this->connections() as $connectionName) {
```
nikophil: a real robust implementation would be to have a service per connection, but this sounds a bit overkill

src/ORM/ResetDatabase/DamaDatabaseResetter.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->decorated->resetDatabase($kernel);
+
+        if (PersistenceManager::isOrmOnly()) {
+            // add global stories so they are available after transaction rollback
+            Configuration::instance()->stories->loadGlobalStories();
+        }
+
+        // shutdown kernel before re-enabling static connections
+        // this would prevent any error if any ResetInterface execute sql queries (example: symfony/doctrine-messenger)
+        $kernel->shutdown();
```
nikophil, line 46: fixes https://github.com/zenstruck/foundry/issues/658

jongotlin:
I have tested this PR and it works great for me. Thank you!

src/Persistence/ResetDatabase/DatabaseResetterInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Persistence\ResetDatabase;
+
+use Symfony\Component\HttpKernel\KernelInterface;
+
+interface DatabaseResetterInterface
```
kbond:
```suggestion
interface DatabaseResetter
```

src/Persistence/ResetDatabase/SchemaResetterInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Persistence\ResetDatabase;
+
+use Symfony\Component\HttpKernel\KernelInterface;
+
+interface SchemaResetterInterface
```
kbond:
```suggestion
interface SchemaResetter
```

src/ORM/ResetDatabase/OrmDatabaseResetterInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\ORM\ResetDatabase;
+
+use Zenstruck\Foundry\Persistence\ResetDatabase\DatabaseResetterInterface;
+use Zenstruck\Foundry\Persistence\ResetDatabase\SchemaResetterInterface;
+
+interface OrmDatabaseResetterInterface extends DatabaseResetterInterface, SchemaResetterInterface
```
kbond:
```suggestion
interface OrmDatabaseResetter extends DatabaseResetterInterface, SchemaResetterInterface
```

src/ORM/ResetDatabase/OrmDatabaseResetter.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Zenstruck\Foundry\ORM\ResetDatabase;
+
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/**
+ * @internal
+ * @author Nicolas PHILIPPE <nikophil@gmail.com>
+ */
+final class OrmDatabaseResetter extends AbstractOrmResetter implements OrmDatabaseResetterInterface
```
kbond:
```suggestion
final class OrmSchemaDatabaseResetter extends AbstractOrmResetter implements OrmDatabaseResetterInterface
```

src/ORM/ResetDatabase/AbstractOrmResetter.php
```diff
@@ -0,0 +1,66 @@
...
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
+use Doctrine\DBAL\Platforms\SQLitePlatform;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Zenstruck\Foundry\Persistence\SymfonyCommandRunner;
+
+/**
+ * @author Nicolas PHILIPPE <nikophil@gmail.com>
+ */
+abstract class AbstractOrmResetter
```
kbond:
```suggestion
abstract class BaseOrmResetter
```

kbond (approved):
I like this, just a few CS notes.

src/Persistence/ResetDatabase/ResetDatabaseManager.php
kbond: > WDYT of something like beforeClassResetter / beforeTestResetter?

It should be beforeFirstTestResetter / beforeTestResetter, right?

I prefer database/schema reset but I don't have a strong opinion. It makes more sense in my head but my experience is 100% SQL.

src/ORM/ResetDatabase/OrmDatabaseResetterTrait.php
kbond: We'll wait for this to be an issue.

kbond:
If not in the docs, at least in the PR description, I'd like to see an example of hooking into this in userland.

src/ORM/ResetDatabase/OrmDatabaseResetterTrait.php
nikophil: yes of course

src/Persistence/ResetDatabase/ResetDatabaseManager.php
nikophil: I kinda like beforeFirstTestResetter / beforeTestResetter :)

I think it is much more easier to reason about if someone wants to extend the reset db mecahnism

PR #691, opened 2024-08-29
Have `flush_after()` return the callback's return

HypeMC:
Same as #442.

I'm not sure why this was removed in 2.x, but it makes migrating to v2 harder.


nikophil (approved):

nikophil:
thanks  @HypeMC

nikophil:
released in `v2.0.8`

PR #692, opened 2024-09-02
refactor: extract reset:migration tests in another testsuite

### LINKED ISSUE #694: [Tests] provide tests for "migration configurations"
https://github.com/zenstruck/foundry/pull/686

nikophil:
This is an experiment, in order to take "reset: migrate" permutation in another testsuite

(by the way, this fixes #694)


.github/workflows/ci.yml
```diff
@@ -145,8 +134,63 @@ jobs:
...
+    name: Test migration - D:${{ matrix.database }}, ${{ matrix.use-dama == 1 && ' (dama)' || '' }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        database: [ mysql, pgsql, sqlite ]
+        use-dama: [ 0, 1 ]
+        exclude:
+          - database: sqlite
+            use-dama: 0
```
nikophil: there is a problem with migrate + sqlite + "no dama" :
because we're calling `dropAndResetDatabase()` which does not actually drop the database for sqlite:
https://github.com/zenstruck/foundry/blob/2.x/src/ORM/AbstractORMPersistenceStrategy.php#L126-L129

I'm wondering if we should not remove the actual file `sqlite.db` here, but I'm not sure we can access it easily :thinking: 

WDYT @kbond ?

tests/Fixture/TestMigrationKernel.php
```diff
@@ -0,0 +1,125 @@
...
+                    ],
+                ],
+            ]);
+        }
+
+        $c->loadFromExtension('doctrine_migrations', [
+            'migrations_paths' => [
+                'Zenstruck\\Foundry\\Tests\\Fixture\\Migrations' => '%kernel.project_dir%/tests/Fixture/Migrations',
+            ],
+            'transactional' => false,
```
nikophil, line 115: I also discovered that transactional mode + MySQL is buggy - at least in our tests, I'm not using MySQL currently, I can live with this, till a user comes with a problem around it

.github/workflows/ci.yml
```diff
@@ -145,8 +134,70 @@ jobs:
...
+        use-dama: [ 0, 1 ]
+        with-migration-configuration-file:
+          - ''
+          - 'tests/Fixture/MigrationTests/configs/migration-configuration.php'
+          - 'tests/Fixture/MigrationTests/configs/migration-configuration-transactional.php'
+        exclude:
+          - database: sqlite
+            use-dama: 0
+          - database: mysql
+            with-migration-configuration-file: 'tests/Fixture/MigrationTests/configs/migration-configuration-transactional.php'
```
nikophil, line 107: I also discovered that for an obscure reason, transactional mode + MySQL is buggy - at least in our tests, I'm not using MySQL currently in my day-to-day job, so I can live with this. If a user comes with a problem around it, I'll dig a little bit more!

.github/workflows/ci.yml
nikophil: ok, I've actually removed the .db file, and it fixes the problem

src/ORM/ResetDatabase/BaseOrmResetter.php
```diff
@@ -30,6 +35,19 @@ public function __construct(
     ) {
     }
 
+    final public function resetBeforeEachTest(KernelInterface $kernel): void
+    {
+        if (self::$inFirstTest) {
+            self::$inFirstTest = false;
+
+            return;
+        }
```
nikophil, line 44: prevents the to run twice the schema create in the first test, when not using dama.

kbond (approved):

PR #693, opened 2024-09-02
Fix Psalm TooManyTemplateParams

ddeboer:
Since Foundry 2.0.8, due to https://github.com/zenstruck/foundry/pull/675/files#diff-f1158fbec0bb446d87dbd5346fba683c58ccec1a98b1787511ec7786585e8c31R40,
Psalm throws:

```
ERROR: TooManyTemplateParams - MyFactory.php - TimeslotFactory<Timeslot> has too many template params, expecting 0 (see https://psalm.dev/184)
            'timeslot' => TimeslotFactory::new(),
```

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -36,10 +36,8 @@ abstract class Factory
 
     /**
      * @param Attributes $attributes
-     *
-     * @return static<T>
      */
-    final public static function new(array|callable $attributes = []): static // @phpstan-ignore return.phpDocType
+    final public static function new(array|callable $attributes = []): static
     {
         if (Configuration::isBooted()) {
             $factory = Configuration::instance()->factories->get(static::class);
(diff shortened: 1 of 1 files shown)


nikophil:
Hi @ddeboer 

can't we add any `@psalm-suppress` annotation which would forget about this problem, the same way it was done for PHPStan?

This was initially made to please autocompletion in IDE... 

Or maybe, we can keep `@return static<T>` and add `@phpstan-return static` (I think psalm understands both `@phpstan-return` and `@psalm-return`, and they have higher priority than a simple `@return`)

but afterwards, we must check that autocompletion [still works](https://github.com/zenstruck/foundry/pull/675#issuecomment-2247272826)

ddeboer:
Thanks for your feedback @nikophil. Interestingly, I’m getting the same (IntelliJ) autocompletion results with this PR’s changes, so without the `@return static<T>` statements. Could you verify?

nikophil:
I've just checked in PHPStorm the exact same examples I used in https://github.com/zenstruck/foundry/pull/675#issuecomment-2247272826 and the autocomplete is working for all of them :tada: 

BTW, that's pretty strange, I don't think I've upgraded PHPStorm since then :thinking:

nikophil:
your PR is released :wink:

PR #699, opened 2024-10-02
fix: allow non object state in stories

### LINKED ISSUE #697: Regression on story states for non object state
Since 2.0.5, story states seems to be broken for any values added with `addState` that are not objects.

`addState` allows `mixed` type, however in the `getState` method, we now call a `refresh` on `PersistenceManager` that expect an object. Hence the error in my case : 

```
TypeError: Zenstruck\Foundry\Persistence\PersistenceManager::refresh(): Argument #1 ($object) must be of type object, array given, called in /srv/api/vendor/zenstruck/foundry/src/Story.php on line 140
```

https://github.com/zenstruck/foundry/blob/6105a364f14dd2a29e3a76eb5127ce2f24b1958d/src/Story.php#L135

Brewal:
fixes #697

### DIFF
--- a/src/Story.php
+++ b/src/Story.php
@@ -143,7 +143,7 @@ abstract class Story
             $unwrappedObject = ProxyGenerator::unwrap($this->state[$name]);
             Configuration::instance()->persistence()->refresh($unwrappedObject, force: true);
 
-            return $isProxy ? ProxyGenerator::wrap($unwrappedObject) : $unwrappedObject;
+            return $isProxy ? ProxyGenerator::wrap($unwrappedObject) : $unwrappedObject; // @phpstan-ignore argument.templateType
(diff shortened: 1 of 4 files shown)


nikophil:
Hi @Brewal 

thanks for this fix!

could you please add a test which ensures there won't be a regression anymore?

Brewal:
@nikophil sure, I'll do that soon.

Brewal:
@nikophil I added some assertions in StoryTest. Hope that's good enough. The tests do break without the fix.

nikophil (approved):

nikophil:
super nice, thank you for this fix!

nikophil:
your fix is released in [2.1.0](https://github.com/zenstruck/foundry/releases/tag/v2.1.0)

PR #703, opened 2024-10-18
docs: add missing docs

### LINKED ISSUE #702: Support for Object Creation from Named Constructors in Zenstruck Foundry
Hi,

I checked the library and read through the documentation, but there is currently no way to create a new object using a named constructor. In my opinion, this should be addressed, as named constructors are very popular in DDD (Domain-Driven Design) practices today.

Thank you!

nikophil:
- **docs: document new option orm.migrations.configurations** documents https://github.com/zenstruck/foundry/issues/685
- **docs: add some doc for the instantiator** (fixes #702)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -2067,6 +2067,19 @@ Objects can be fetched from pools in your tests, fixtures or other stories:
     ProvinceStory::getRandomRange('be', 1, 4); // between 1 and 4 random Province|Proxy's from "be" pool
     ProvinceStory::getPool('be'); // all Province|Proxy's from "be" pool
 
+Static Analysis
+---------------
+
+Psalm
+~~~~~
+
+A Psalm extension is shipped with the library.
+Please, enable it with:
+
+.. code-block:: terminal
+
+    $ vendor/bin/psalm-plugin enable zenstruck/foundry
+
 Bundle Configuration
 --------------------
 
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -648,9 +648,17 @@ You can customize the instantiator in several ways:
...
+        ->instantiateWith(Instantiator::use(function(string $title): object {
+            return new Post($title); // ... your own logic
+        }))
+
         // the instantiator is just a callable, you can provide your own
-        ->instantiateWith(function(array $attributes, string $class): object {
-            return new Post(); // ... your own logic
+        ->instantiateWith(function(string $title): object {
+            return new Post($title); // ... your own logic
         })
```
nikophil: I don't know if we should promote this... AFAIK this method won't benefit from the hydration step, so it can lead to buggy behaviors.

I'm even wondering if we should not deprecate this feature, in favor of `Instantiator::use()`... WDYT?

kbond: This doesn't work I don't think. Unless wrapped in Instantiator::use(), the callable's args will be array/class-string.

But to your comment, kind of thing we need an escape hatch for users to do their own thing completely.

nikophil: yes, you're right, I've rollbacked to the original form

kbond (approved):

Issue #696, opened 2024-09-30
Using `Zenstruck\Foundry\Test\Factories` triggers deprecations


deluxetom:
Similar to https://github.com/zenstruck/foundry/issues/680 and https://github.com/zenstruck/foundry/issues/650

I tried everything to ignore those:
- Removed `phpunit-bridge`
- Updated the deprecation level (via `SYMFONY_DEPRECATIONS_HELPER`) but thats just in PHPUnit, it shouldn't impact Symfony

<img width="1364" alt="Screenshot 2024-09-30 at 9 43 32 AM" src="https://github.com/user-attachments/assets/e383bc4d-07c8-4dfd-8560-b6d539881784">

I'm not sure what else to try to remove those internal notices.
I use Factories in my data fixtures, `dev` and `test` environments

Let me know if I missed something

nikophil:
Hi,

sorry for the late answer... I think I've finally found an elegant solution for this

see https://github.com/zenstruck/foundry/pull/704

nikophil:
fixed by https://github.com/zenstruck/foundry/pull/704

PR #704, opened 2024-10-18
refactor: fix proxy system and introduce psalm extension

### LINKED ISSUE #701: Psalm errors with Attributes phpstan-type declaration
Hi !

I should have opened this issue a month ago (sorry about that) since I commented [here](https://github.com/zenstruck/foundry/issues/623#issuecomment-2342855336)  thinking it was related.

As mentionned in the linked comment, since `2.0.8` I started getting these kind of Psalm errors: 
```
InvalidArgument - src/DataFixtures/AppFixtures.php:128:70 - Argument 1 of App\Foundry\Factory\Node\EndConversationNodeFactory::createOne expects Zenstruck\Foundry\Persistence\Attributes, but array{position: App\Entity\Position, script: App\Entity\Script} provided (see https://psalm.dev/004)
```
From what I understand: Psalm doesn't seem to appreciate the fact that I pass an array instead of an instance of a non existing class Attributes to my factories like this:
```
EndConversationNodeFactory::createOne([
    'script' => $script,
]);
```

After some investigations it seems that these "errors" have been introduced by this [commit](https://github.com/zenstruck/foundry/commit/65c1cc2f6dc034435e325e7913bee41313f1b6a0)

and particularly by these added lines in the PersistentProxyObjectFactory 
image

if I remove `@method ` annotations, everything gets back to normal on my end, I'm trying to understand what's happening and my best guess is that  @method annotations should not use phpstan aliases like Attributes. 

Hope I was clear enough. If you need me to provide more information, just ask, I'd be happy to do so.

nikophil:
fixes https://github.com/zenstruck/foundry/issues/701 https://github.com/zenstruck/foundry/issues/696

I've decided to completely remove all methods in `PersistenceProxyObjectFactory`, so we do not override `@final` methods anymore. We still need to figure out a solution for `PersistenceObjectFactory::create()`, I'll do this in a further PR.

Psalm does not understand the following:
```php
/**
 * @extends PersistentProxyObjectFactory<User>
 */
class UserFactory extends PersistentProxyObjectFactory{}

/**
 * @template T of object
 * @extends PersistentObjectFactory<T&Proxy<T>>
 */
abstract class PersistentProxyObjectFactory extends PersistentObjectFactory{}

/**
 * @template T of object
 */
abstract class PersistentObjectFactory{}
```
it always consider the template to be `User` and not `User&Proxy<User>`. I think because we're breaking covariance (`T&Proxy<T>` is wider than `T`)

I've finally decided to introduce a psalm extension, we will be able to fix nearly every typing problem in here.


kbond (approved):
Looks good! I'm unfamiliar with psalm plugins, will this be auto-registered in projects using foundry and psalm?

nikophil:
>  I'm unfamiliar with psalm plugins, will this be auto-registered in projects using foundry and psalm?

yes, thanks to [the change in composer.json](https://github.com/zenstruck/foundry/blob/cff5ae531b5fac15b6e7ec79052a25c46d129b02/composer.json#L74-L76)

deluxetom (approved):
Great solution, thank you!

src/Factory.php
```diff
@@ -25,17 +25,18 @@
  */
 abstract class Factory
 {
-    /** @var Attributes[] */
+    /** @phpstan-var Attributes[] */
```
nikophil, line 28: everytime we use phpstan pseudo types, we should use `@phpstan-` annotations because Psalm seems to not understand it

and moreover, see this comment from stof [here](https://github.com/symfony/symfony/pull/50687#discussion_r1281529038), I think it is relevant

I'll make another PR where I fix all of the related docblocks

src/ObjectFactory.php
```diff
@@ -58,6 +56,7 @@ public function create(callable|array $attributes = []): object
 
         $parameters = $this->normalizeParameters($parameters);
         $instantiator = $this->instantiator ?? Configuration::instance()->instantiator;
+        /** @var T $object */
```
nikophil, line 59: it seems like PHPStorm does not only interprets methods signature (like PHPStan would do) but it events guesses stuff from the methods' body. I had really hard times to find but it basically fixes 90% of auto-complete problems :tada:

src/Persistence/PersistentObjectFactory.php
```diff
@@ -140,42 +126,40 @@ public static function random(array $criteria = []): object
...
-     * @final
-     *
      * @return T
      *
      * @throws \RuntimeException If no objects exist
      */
     public static function first(string $sortBy = 'id'): object
     {
-        return static::repository()->firstOrFail($sortBy);
+        /** @var T $object */
```
nikophil, line 135: same, this is here to help auto-complete

kbond (approved):

PR #714, opened 2024-10-31
fix: Change `RepositoryDecorator::inner()` visibility to public

### LINKED ISSUE #713: `RepositoryDecorator::inner()` is a private method
The documentation mentions that we can call `$repository->inner()` to get the real wrapped repository from a proxy (see https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#repository-proxy).

Unfortunately, since Foundry v2, it seems that the new `RepositoryDecorator` class that provides the `inner()` method, declares it as a private method.

I'm not sure if this change was intentional, but it would be very helpful to get the `inner()` method back to public :)

Thanks!

marienfressinaud:
Fixes https://github.com/zenstruck/foundry/issues/713

`RepositoryDecorator::inner()` is documented as public in the documentation, but its visibility changed during the release of Foundry 2.

### DIFF
--- a/src/Persistence/RepositoryDecorator.php
+++ b/src/Persistence/RepositoryDecorator.php
@@ -229,7 +229,7 @@ class RepositoryDecorator implements ObjectRepository, \IteratorAggregate, \Coun
     /**
      * @return ObjectRepository<T>
      */
-    private function inner(): ObjectRepository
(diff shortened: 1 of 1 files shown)


nikophil:
thanks @marienfressinaud

PR #718, opened 2024-11-05
Fix missing comma in upgrade doc UPGRADE-2.0.md

### LINKED ISSUE #716: [UPGRADE-2.0.md] Missing comma in example code
**Description**: In the UPGRADE-2.0.md file, there is a minor syntax error in the provided PHP example code for configuring RectorConfig. Specifically, there is a missing comma in the withPaths array, which could cause a syntax error when copying the code directly.

[UPGRADE-2.0.md file](https://github.com/zenstruck/foundry/blob/2.x/UPGRADE-2.0.md?plain=1#L56)

**Code Example**: Here's the section of the example code with the missing comma:

```
return RectorConfig::configure()
    ->withPaths([
        // add all paths where your factories are defined and where Foundry is used
        'src/Factory'
        'tests'
    ])
    ->withSets([FoundrySetList::UP_TO_FOUNDRY_2])
;
```

**Proposed Fix**: A comma should be added after 'src/Factory' to ensure the array is correctly formatted. Here’s the corrected code:

```
return RectorConfig::configure()
    ->withPaths([
        'src/Factory',
        'tests'
    ])
    ->withSets([FoundrySetList::UP_TO_FOUNDRY_2])
;
```

**Suggested Solution**: Update the UPGRADE-2.0.md file with this corrected example code to prevent any syntax errors for users following the guide.

justpilot:
Closes #716

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -53,7 +53,7 @@ use Zenstruck\Foundry\Utils\Rector\FoundrySetList;
 return RectorConfig::configure()
     ->withPaths([
         // add all paths where your factories are defined and where Foundry is used
-        'src/Factory'
+        'src/Factory',
         'tests'
     ])
     ->withSets([FoundrySetList::UP_TO_FOUNDRY_2])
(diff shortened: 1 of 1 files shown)


nikophil:
thanks @justpilot

PR #722, opened 2024-11-22
fix: bug with one to many

nikophil:
Fixes a bug where using "inversed one-to-many" was creating too many entites in the inverse side of the association:

this was failing, because 3 categories were created (only when using "non-cascade-entities")
```php
    public function one_to_many(FactoryCollection|array $contacts): void
    {
        $category = $this->categoryFactory()::createOne([
            'contacts' => [
                $this->contactFactory(),
                $this->contactFactory()
            ],
        ]);

        $this->categoryFactory()::repository()->assert()->count(1);        
    }
```

### DIFF
--- a/src/FactoryCollection.php
+++ b/src/FactoryCollection.php
@@ -23,10 +23,51 @@ final class FactoryCollection implements \IteratorAggregate
 {
     /**
      * @param Factory<T>                      $factory
-     * @phpstan-param \Closure():iterable<Attributes> $items
+     * @phpstan-param \Closure():iterable<Attributes>|\Closure():iterable<Factory<T>> $items
      */
     private function __construct(public readonly Factory $factory, private \Closure $items)
     {
+    }
+
+    /**
+     * @phpstan-assert-if-true non-empty-list<Factory<T>> $potentialFactories
+     *
+     * @internal
+     */
+    public static function accepts(mixed $potentialFactories): bool
+    {
+        if (!is_array($potentialFactories) || count($potentialFactories) === 0 || !array_is_list($potentialFactories)) {
+            return false;
+        }
+
+        if (!$potentialFactories[0] instanceof ObjectFactory) {
+            return false;
+        }
+
+        foreach ($potentialFactories as $potentialFactory) {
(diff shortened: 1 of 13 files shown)


src/FactoryCollection.php
```diff
@@ -23,12 +23,47 @@ final class FactoryCollection implements \IteratorAggregate
...
+            return false;
+        }
+
+        if (!$potentialFactories[0] instanceof Factory) {
+            return false;
+        }
+
+        foreach ($potentialFactories as $potentialFactory) {
+            if (!$potentialFactory instanceof Factory
+                || $potentialFactory::class() !== $potentialFactories[0]::class()) {
```
nikophil, line 49: in this loop, I want to check that all the objects in the list are a factory targeting the same class.

I'm wondering if we can go one step forward, and check if one class `is_a()` the other one, we would allow using inheritance here. No sure it is widely used :shrug: 
```php
        foreach ($potentialFactories as $potentialFactory) {
            if (!$potentialFactory instanceof ObjectFactory
                // allow inheritance
                || !(
                        is_a($potentialFactory::class(), $potentialFactories[0]::class(), true)
                        && is_a($potentialFactories[0]::class(), $potentialFactory::class(), true)
                )
            ) {
                return false;
            }
        }
```
WDYT?

tests/Integration/ORM/CascadeEntityFactoryRelationshipTest.php
```diff
@@ -27,15 +27,15 @@ final class CascadeEntityFactoryRelationshipTest extends EntityFactoryRelationsh
      */
     public function ensure_to_one_cascade_relations_are_not_pre_persisted(): void
     {
-        $contact = $this->contactFactory()
```
nikophil, line 30: sorry for all this noise, I had to migrate from "non static" to "static" for all these "factory" methods, to satisfy PHPUnit 11

src/FactoryCollection.php
kbond: Let's leave as-is for now and adjust if someone hits this.

kbond (approved):

PR #728, opened 2024-12-02
feat: add `#[WithStory]` attribute

### LINKED ISSUE #430: Add #`[WithStory]` attribute
```php
#[WithStory(MyStory::class)] // will be called for the test case
public function testSomething(): void

#[WithStory(MyStory::class)] // will be called once for the test class
class SomeTest extends TestCase {}
```

nikophil:
fixes #430

- **feat: introduce WithStory attribute**
- **feat: apply WithStory attribute at class level**
- **feat: allow WithStory attribute on parent class**

I think that only an attribute related to stories is useful, I don't see any practical usage of an attribute related to factories

TODO:
- [x] story service should work
- [x] ~only load story once if it is added to a class? Maybe this will be conditioned by a boolean~ won't be made

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1608,35 +1608,26 @@ PHPUnit Data Providers
 
 It is possible to use factories in
 `PHPUnit data providers <https://phpunit.readthedocs.io/en/9.3/writing-tests-for-phpunit.html#data-providers>`_.
-Their usage depends on which Foundry version you are running:
+Their usage depends on whether you're using Foundry's `PHPUnit Extension`_ or not.:
 
-Data Providers with Foundry ^2.2
-................................
+With PHPUnit Extension
+......................
 
-From version 2.2, Foundry provides an extension for PHPUnit.
-You can install it by modifying you ``phpunit.xml.dist``:
+.. versionadded::  2.2
 
-.. configuration-block::
+    The ability to call ``Factory::create()`` in data providers was introduced in Foundry 2.2.
 
-    .. code-block:: xml
-    
-        <phpunit>
-            <extensions>
-                <bootstrap class="Zenstruck\Foundry\PHPUnit\FoundryExtension"/>
-            </extensions>
-        </phpunit>
-    
 .. warning::
 
-    This PHPUnit extension requires at least PHPUnit 11.4.
+    You will need at least PHPUnit 11.4 to call ``Factory::create()`` in your data providers.
 
-Using this extension will allow to use your factories in your data providers the same way you're using them in tests.
-Thanks to it, you can:
+Thanks to Foundry's `PHPUnit Extension`_, you'll be able to use your factories in your data providers the same way
+you're using them in tests. Thanks to it, you can:
     * Call ``->create()`` or ``::createOne()`` or any other method which creates objects in unit tests
     (using ``PHPUnit\Framework\TestCase``) and functional tests (``Symfony\Bundle\FrameworkBundle\Test\KernelTestCase``)
     * Use `Factories as Services`_ in functional tests
     * Use `faker()` normally, without wrapping its call in a callable
-    
+
 ::
 
     use App\Factory\PostFactory;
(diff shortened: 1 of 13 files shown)


src/Attribute/WithStory.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Zenstruck\Foundry\Attribute;
+
+use Zenstruck\Foundry\Story;
+
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
+final class WithStory
```
nikophil, line 10: or:
- `#[Story]`
- `#[WithStoryLoaded]`
- `#[UseStory]`
- `#[LoadStory]`

kbond: I like `WithStory`

kbond:
> only load story once if it is added to a class? Maybe this will be conditioned by a boolean

So this would only load the story before the first test in a class?

nikophil:
That's what I have in mind, but I'm afraid it's would be too much complicated, we have to hack into dama and reset database for this, but I'll give this a chance in another PR

kbond:
This is soooo cool, I love it!

> but I'm afraid it's would be too much complicated,

Yeah, this feels complex and I'm worried about breaking test isolation. Kind of a sub-global state...

nikophil:
> This is soooo cool, I love it!

super nice :heart: 
sadly, if you're still not using PHPUnit 10, you won't be able to use this feature :confused: 

> Yeah, this feels complex and I'm worried about breaking test isolation. Kind of a sub-global state...

yes I think I won't even try, I've checked how dama works and it would require some ugly stuff to make this work. 

BTW, the PR is RFR now, I've added some docs as well.

docs/index.rst
```diff
@@ -1610,33 +1610,16 @@ It is possible to use factories in
 `PHPUnit data providers <https://phpunit.readthedocs.io/en/9.3/writing-tests-for-phpunit.html#data-providers>`_.
 Their usage depends on which Foundry version you are running:
 
-Data Providers with Foundry ^2.2
-................................
+Data Providers with Foundry ^2.2 and PHPUnit ^11.4
```
kbond: Below, you have the extension requires 10+.

I think we should remove the version constraints from these subtitles and use a versionadded + warning for required phpunit versions

nikophil: I've updated the docs, WDYT?

docs/index.rst
```diff
@@ -2181,6 +2197,29 @@ Please, enable it with:
...
+            </extensions>
+        </phpunit>
+
+This extension provides the following features:
+- support for the `#[WithStory] Attribute`_
+- ability to use ``Factory::create()`` in `PHPUnit Data Providers`_ (along with PHPUnit ^11.4)
+
+.. note::
+
+    The extension was introduced in Foundry 2.2, and is only compatible with PHPUnit ^10.
```
```suggestion
.. versionadded:: 2.2

    The PHPUnit extension was introduced in Foundry 2.2.

.. warning::

    The PHPUnit extension is only compatible with PHPUnit 10+.
```

docs/index.rst
```diff
@@ -2168,6 +2155,35 @@ Objects can be fetched from pools in your tests, fixtures or other stories:
     ProvinceStory::getRandomRange('be', 1, 4); // between 1 and 4 random Province|Proxy's from "be" pool
     ProvinceStory::getPool('be'); // all Province|Proxy's from "be" pool
 
+#[WithStory] Attribute
+~~~~~~~~~~~~~~~~~~~~~~
+
+You can use the ``#[WithStory]`` attribute to load stories in your tests:
```
kbond, line 2169:
```suggestion
.. versionadded:: 2.3

    The `#[WithStory]` attribute was added in Foundry 2.3.

You can use the ``#[WithStory]`` attribute to load stories in your tests:
```

kbond:
Can you move the versionadded/warnings to directly below the subtitle? I feel like people might get excited about the feature and start using without reading below. :)

kbond (approved):

PR #729, opened 2024-12-04
chore: remove error handler hack

### LINKED ISSUE #711: Change in symfony exception handler
Hi!

In yesterday releases (7.1.6, 6.4.13, 5.4.45) symfony fixed it's exception/error handler registration issue with phpunit.
see: https://github.com/symfony/symfony/pull/58372

Now the workaround introduced after my previous issue (https://github.com/zenstruck/foundry/issues/640) is the one that triggers phpunit's `THE ERROR HANDLER HAS CHANGED!` warning :D  

if I remove the call to KernelHelper::shutdownKernel the warning is gone.

```
env:
phpunit/phpunit                                11.4.2
symfony/framework-bundle                       7.1.6
zenstruck/foundry                              1.38.3
```

nikophil:
fixes #711


composer.json
```diff
@@ -40,6 +40,7 @@
         "symfony/dotenv": "^6.4|^7.0",
         "symfony/maker-bundle": "^1.55",
         "symfony/phpunit-bridge": "^6.4|^7.0",
+        "symfony/runtime": "^6.4|^7.0",
```
nikophil, line 43: I've added runtime as a dev requirement because the fix [only works with runtime present](https://github.com/symfony/symfony/pull/58372/files#r1774895155) - -but it's not used anywhere :shrug:

PR #730, opened 2024-12-05
fix: Fix the parameter name of the first and last methods

marien-probesys:
This PR fixes the documentation of the `Persistent*ObjectFactory` `first()` and `last()` methods.

I discovered the issue when calling `MyFactory::first(sortedField: 'field')` with the error:

```
Error: Unknown named parameter $sortedField
```

Indeed, the parameter is now named `sortBy`.

Then I realized that the field needed to be renamed also in the `MakeFactoryPHPDocMethod` class.

Finally, I also fixed the parameter name in the `*RepositoryDecorator::last()` method to be more coherent (the name was correct in the `first()` method).


nikophil:
nice! thank you @marien-probesys

PR #731, opened 2024-12-05
Typo in Immutable section

franckranaivo:
I think the doc meant Factories instead of Factory's.


nikophil:
Thanks @franckranaivo

PR #732, opened 2024-12-06
fix: detect if relation is oneToOne

nikophil:
This PR fixes a last bug before we can release 2.3: some edge case was buggy, when the a many to one was referencing a "self referencing" class


src/ORM/OrmV2PersistenceStrategy.php
```diff
@@ -29,17 +29,18 @@ public function relationshipMetadata(string $parent, string $child, string $fiel
     {
         $metadata = $this->classMetadata($parent);
 
-        $association = $this->getAssociationMapping($parent, $field);
+        $association = $this->getAssociationMapping($parent, $child, $field);
```
nikophil, line 32: I think that at some point I'll refacto this method which is becoming ugly

kbond (approved):

PR #738, opened 2024-12-10
feat: provide current factory to after/before hooks

### LINKED ISSUE #677: Cannot access actual factory instance in after/before Instantiate hooks
I recently had a pretty sneaky bug. Given this code:
```php
final class SomeFactory extends PersistentObjectFactory
{
    {
        return $this->beforeInstantiate(
            function (array $attributes): array {
                if ($this->isPersisting()) {
                    // do some stuff which interacts with db
                } else {
                    // do some stuff without db
                }

                return $attributes;
            }
        );
    }
}

// in some test:
SomeFactory::new()->withoutPersisting()->create();
```

We're entering in the `if()` because `$this` here represents the factory returned by `SomeFactory::new()` and not the last instance on which we call `create()`. This is really unlucky, I had hard time to figure what was going on...

My first guess to fix this was to bind the closure to the actual instance used in `create()`:
```php
// SomeFactory
public function create(callable|array $attributes = []): object
{
    // ...

    foreach ($this->beforeInstantiate as $hook) {
        try {
	        $hook = \Closure::fromCallabale($hook)->bindTo($this);
		} catch(\Exception) {
            // the closure is static and cannot be bound to anything
        }

        $parameters = $hook($parameters, static::class());

        // ...
    }
}
```

it works nicely, but it would not work outside from a factory, so this is a no-go.

Another solution would be to the current factory as a new parameter to all our hooks:
```php
    /**
     * @param callable(Parameters,class-string<T>, static):Parameters $callback
     */
    final public function beforeInstantiate(callable $callback): static;

    /**
     * @final
     *
     * @param callable(T,Parameters, static):void $callback
     */
    public function afterInstantiate(callable $callback): static;

    /**
     * @param callable(T, Parameters, static):void $callback
     */
    final public function afterPersist(callable $callback): static;
```

WDYT?

nikophil:
fixes #677

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -552,20 +552,24 @@ they were added.
     use Zenstruck\Foundry\Proxy;
 
     PostFactory::new()
-        ->beforeInstantiate(function(array $attributes): array {
+        ->beforeInstantiate(function(array $attributes, string $class, static $factory): array {
             // $attributes is what will be used to instantiate the object, manipulate as required
+            // $class is the class of the object being instantiated
+            // $factory is the factory instance which creates the object
             $attributes['title'] = 'Different title';
 
             return $attributes; // must return the final $attributes
         })
-        ->afterInstantiate(function(Post $object, array $attributes): void {
+        ->afterInstantiate(function(Post $object, array $attributes, static $factory): void {
             // $object is the instantiated object
             // $attributes contains the attributes used to instantiate the object and any extras
+            // $factory is the factory instance which creates the object
         })
-        ->afterPersist(function(Post $object, array $attributes) {
+        ->afterPersist(function(Post $object, array $attributes, static $factory) {
             // this event is only called if the object was persisted
             // $object is the persisted Post object
             // $attributes contains the attributes used to instantiate the object and any extras
+            // $factory is the factory instance which creates the object
         })
 
         // multiple events are allowed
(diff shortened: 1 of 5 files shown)


docs/index.rst
```diff
@@ -552,20 +552,24 @@ they were added.
     use Zenstruck\Foundry\Proxy;
 
     PostFactory::new()
-        ->beforeInstantiate(function(array $attributes): array {
+        ->beforeInstantiate(function(array $attributes, string $class, static $factory): array {
             // $attributes is what will be used to instantiate the object, manipulate as required
+            // $class is the class of the object being instantiated
```
nikophil, line 557: I'm not sure why we're passing the $class here :thinking:

tests/Integration/Persistence/FactoryWithHooksInInitializeTest.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Zenstruck\Foundry\Tests\Integration\Persistence;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Zenstruck\Foundry\Test\Factories;
+use Zenstruck\Foundry\Tests\Fixture\Factories\WithHooksInInitializeFactory;
+
+/**
+ * @author Nicolas Philippe <nikophil@gmail.com>
+ */
+final class FactoryWithHooksInInitializeTest extends KernelTestCase
```
nikophil, line 14: I did not find any other class to put this test (we need a `KernelTestCase`, so that `->isPersisting()` can return true)

docs/index.rst
kbond: Me either... I think it was in case someone had some kind of beforeInstantiate object/service that could act on many different objects?

kbond (approved):

PR #742, opened 2024-12-11
feat: schedule objects for insert right after instantiation

### LINKED ISSUE #537: Should object be scheduled for insert before afterInstantiate call ?
Hi, 

@nikophil and I having an issue in one of my tests concerning the way `afterInstantiate` is handled. 

I struggle to understand what is meant to happen but my intuition is pointing to the moment where the "root" object is scheduled for insertion and the moment `afterInstantiate` callbacks are called.

We cannot use `afterPersist` callbacks because some of our tests are preformed without persistence.

To my understanding the process should be the following 

```mermaid
flowchart LR
    A[new] --> B(beforeInstantiate) --> B
    B --> C[schedule for persistence]
    C --> D(afterInstantiate) --> D
    D --> E[persist]
    E --> F(afterPersist) --> F
```

But when we used this process in a factory like this : 

```php

final class PostFactory extends ModelFactory
{
    public function published(): static 
    {
        return $this->afterInstantiate(
            static function (Post $post): void {
                PublishFactory::new([
                    'post' => $post,
                ])->create();
            }
        );
    }
}

```

Doctrine complains about not knowing the post 

 
>Doctrine\ORM\ORMInvalidArgumentException : A new entity was found through the relationship 'App\Entity\Publish#post' >that was not configured to cascade persist operations for entity: ef8013ee-2f5b-43e2-81bb-d9e617fb8973. To solve this >issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the >mapping for example @ ManyToOne(..,cascade={"persist"}).

Ending with this ugly workaround that seems to do the trick 

```php

final class PostFactory extends ModelFactory
{
    public function published(): static 
    {
        return $this->afterInstantiate(
            static function (Post $post): void {
              // Workaround doctrine cascade 
               (new Proxy($post))->save();

                PublishFactory::new([
                    'post' => $post,
                ])->create();
            }
        );
    }
}

```

Is this the intended behavior that the root object is not yet scheduled for insert in this case ?

### LINKED ISSUE #724: bug: infinite loop when 2 factories refer to each other in their `default()` methods
Since 2.x on one of my projects, I have a segfault when I try to create a complex `PersistentProxyObject` with a lot of dependencies. 
On 1.x no issues at all.

I dig a lot to find what's wrong but I have no clue. 
Reproducer is here : https://github.com/simondaigre/foundry-repro/tree/repro-segfault

nikophil:
:warning: merge https://github.com/zenstruck/foundry/pull/738 before

fixes https://github.com/zenstruck/foundry/issues/537
fixes https://github.com/zenstruck/foundry/issues/724


simondaigre:
Minor comment : before this PR, when I use `UserFactory::find(['customer' => $customer]);` on Factory and when there is no results, the error was `RuntimeException: Could not find "App\Entity\User" object.`

And now it's `RuntimeException: No "App\Entity\User" object found for "array".`, which is unclear.

nikophil:
@simondaigre is it related to this PR or Foundry 2? :thinking: 

if not related to this PR, but to foundry 2, could you please open another issue?

kbond (approved):
Nice solution!

PR #744, opened 2024-12-12
fix: define FactoryCollection type more precisely

### LINKED ISSUE #609: Named `FactoryCollection` constructors don't have proper type declarations
The constructor of `FactoryCollection` has a `@phpstan-param Factory<TObject> $factory`. Using this constructor is deprecated, and named constructors like `set`, `range` etc. should be used instead. However, these named constructors are missing this annotation. Also, the annotation is only for PHPStorm, not for Psalm.

This leads to Psalm errors like

```
Argument 1 of Zenstruck\Foundry\FactoryCollection::set expects Zenstruck\Foundry\Factory<object>, but My\Factory\Class provided (see https://psalm.dev/004)
```

nikophil:
fixes #609 

ping @janopae

This PR will create a BC in static analysis to all users which type hint `FactoryCollection` and which does not ignore generics errors

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -101,24 +101,24 @@ abstract class Factory
     abstract public function create(array|callable $attributes = []): mixed;
 
     /**
-     * @return FactoryCollection<T>
+     * @return FactoryCollection<T, static>
      */
     final public function many(int $count): FactoryCollection
     {
-        return FactoryCollection::many($this, $count);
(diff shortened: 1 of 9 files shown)


kbond (approved):
> This PR will create a BC in static analysis to all users which type hint FactoryCollection

I'm ok with this.

PR #753, opened 2024-12-16
[Doc] Added some docs to clarify default attributes and fixed some syntax issues

javiereguiluz:
Today I started using Foundry and faced some issues related to instantiation + defaults. Kevin helped me solve that (thanks!) but I thought we could tweak docs a bit to make that more clear. Thanks!

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -11,7 +11,10 @@ Foundry supports ``doctrine/orm`` (with `doctrine/doctrine-bundle <https://githu
 ``doctrine/mongodb-odm`` (with `doctrine/mongodb-odm-bundle <https://github.com/doctrine/DoctrineMongoDBBundle>`_)
 or a combination of these.
 
-Want to watch a screencast 🎥 about it? Check out https://symfonycasts.com/foundry
+.. admonition:: Screencast
+    :class: screencast
+
+    Want to watch a screencast 🎥 about it? Check out `https://symfonycasts.com/foundry`__
 
 .. warning::
 
@@ -251,6 +254,8 @@ This command will generate a ``PostFactory`` class that looks like this:
             // ...
         }
 
+.. _defaults:
+
 In the ``defaults()``, you can return an array of all default values that any new object
 should have. `Faker`_ is available to easily get random data:
 
@@ -259,12 +264,21 @@ should have. `Faker`_ is available to easily get random data:
     protected function defaults(): array
     {
         return [
-            // Symfony's property-access component is used to populate the properties
-            // this means that setTitle() will be called or you can have a $title constructor argument
+            // use the built-in Fake integration to generate good random values...
             'title' => self::faker()->unique()->sentence(),
             'body' => self::faker()->sentence(),
+
+            // ...or generate the values yourself if you prefer
+            'createdAt' => new \DateTimeImmutable('today'),
         ];
     }
+
+These default values are applied to both the **constructor arguments** and the
+**properties** of the objects. For example, defining a default value for ``title``
+will first attempt to set a constructor argument called ``$title``. If that doesn't
+exist, the `PropertyAccess <https://symfony.com/doc/current/components/property_access.html>`_
+component will be used to call the ``setTitle()`` method or directly set the public
+``$title`` property.
 
 .. tip::
 
@@ -666,6 +680,12 @@ You can customize the instantiator in several ways:
         })
     ;
 
+.. warning::
+
+    The ``instantiateWith()`` method completely overrides the default instantiation
+    and object hydration system. Any attributes defined in the ``defaults()``
+    method will be ignored.
+
 You can customize the instantiator globally for all your factories (can still be overruled by factory instance
 instantiators):
 
@@ -843,8 +863,8 @@ The ``defaults()`` method is called everytime a factory is instantiated (even if
 creating it). Sometimes, you might not want your value calculated every time. For example, if you have a value for one
 of your attributes that:
 
- - has side effects (i.e. creating a file or fetching a random existing entity from another factory)
- - you only want to calculate once (i.e. creating an entity from another factory to pass as a value into multiple other factories)
+* has side effects (i.e. creating a file or fetching a random existing entity from another factory)
+* you only want to calculate once (i.e. creating an entity from another factory to pass as a value into multiple other factories)
 
 You can wrap the value in a ``LazyValue`` which ensures the value is only calculated when/if it's needed. Additionally,
 the LazyValue can be `memoized <https://en.wikipedia.org/wiki/Memoization>`_ so that it is only calculated once.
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -259,13 +264,22 @@ should have. `Faker`_ is available to easily get random data:
     protected function defaults(): array
     {
         return [
-            // Symfony's property-access component is used to populate the properties
-            // this means that setTitle() will be called or you can have a $title constructor argument
+            // use the built-in Fake integration to generate good random values...
```
```suggestion
            // use the built-in Faker integration to generate good random values...
```

docs/index.rst
```diff
@@ -666,6 +680,12 @@ You can customize the instantiator in several ways:
         })
     ;
 
+.. warning::
+
+    The ``instantiateWith()`` method completely overrides the default instantiation
+    and object hydration system. Any attributes defined in the ``defaults()``
+    method will be ignored.
```
kbond: Hmm, they aren't ignored, they're made available in the `array $attributes` callable argument. Of course they could have been overridden by `->with()` also.

nikophil: this is not true: all attributes defined in `defaults()` and all other "states" (added thanks to the `->with()` method) are always used... in a different fashion, depending on the given instantiator:

- `Instantiator::withConstructor()` / `Instantiator::namedConstructor()` /  `Instantiator::use(callable(...))` => the instantiator uses the right attributes for instantiation (with reflection, depending on the constructor/callable parameters) - and then the rest of the attributes is passed to the hydrator.
- `Instantiator::withoutContructor()` => the object is created with `\ReflectionClass::newInstanceWithoutConstructor()`, and all of the attributes are passed to the hydrator
- `->instantiateWith(callable(...))` => all of the attributes are passed to the callable, and the user has the responsability to instantiate/hydrate the object. (I've actually never used this option, but I know it is used in the wild)

And for information, the "hydration" phase has two main behaviors:
- with the normal behavior, the property accessor is used
- when a property is "forced" (you can do that globally, per factory or per property), reflection is used

docs/index.rst
```diff
@@ -259,13 +264,22 @@ should have. `Faker`_ is available to easily get random data:
...
+            'createdAt' => new \DateTimeImmutable('today'),
         ];
     }
 
+These default values are applied to both the **constructor arguments** and the
+**properties** of the objects. For example, defining a default value for ``title``
+will first attempt to set a constructor argument called ``$title``. If that doesn't
+exist, the `PropertyAccess <https://symfony.com/doc/current/components/property_access.html>`_
+component will be used to call the ``setTitle()`` method or directly set the public
+``$title`` property.
```
nikophil: FYI, this is not always true (see my other comment): reflection can also be used when a property is "forced". That being said, I don't know if it is relevant to say this here :sweat_smile:

nikophil (changes requested):
Hi @javiereguiluz 

thanks for this! One of your statements is actually wrong, I tried to explain a little bit more what's happening, so we can find a correct way of documenting things

javiereguiluz:
Thanks for the review folks. I committed some fixes.

docs/index.rst
```diff
@@ -259,13 +264,23 @@ should have. `Faker`_ is available to easily get random data:
...
         ];
     }
 
+These default values are applied to both the **constructor arguments** and the
+**properties** of the objects. For example, defining a default value for ``title``
+will first attempt to set a constructor argument called ``$title``. If that doesn't
+exist, the `PropertyAccess <https://symfony.com/doc/current/components/property_access.html>`_
+component will be used to call the ``setTitle()`` method or directly set the public
+``$title`` property. If all these methods fail, PHP Reflection will be used
+to set the value of the property.
```
kbond, line 282:
```suggestion
``$title`` property.
```

It would actually throw an exception. Setting via reflection is only used if enabled (it isn't by default).

kbond (commented):
Just a small thing.

https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#instantiation has more details about when/how reflection is used.

docs/index.rst
```diff
@@ -666,6 +681,14 @@ You can customize the instantiator in several ways:
         })
     ;
 
+.. warning::
+
+    The ``instantiateWith()`` method fully replaces the default instantiation
+    and object hydration system. Attributes defined in the ``defaults()`` method,
+    as well as any states defined with the ``with()`` method, **will not be
+    applied automatically**. However, they are available as arguments to the
+    ``instantiateWith()`` callable.
```
nikophil, line 690: sorry but I'm still confused :sweat_smile:  

what do you mean by "will not be applied automatically"? I feel this is confusing. 

If you call something like: `$factory->instantiateWith(Instantiator::withoutConstructor())`, the attributes are "applied automatically". They are ineeded passed to the callable, which in that case is the class `Instantiator`, but this all happen under the hood.

What was your initial problem that led you to make this change in the doc?

nikophil (changes requested):

docs/index.rst
javiereguiluz: The original problem was this:

(1) I have this entity, the only one in my app with constructor arguments:

```php
class User
{
    // ...

    public function __construct(string $uuid)
    {
        $this->uuid = $uuid;
    }
}
```

(2) I wanted to apply some default values AND instantiate the entity correctly. This doesn't work:

```php
final class UserFactory extends PersistentProxyObjectFactory
{
    public function __construct()
    {
    }

    {
        return User::class;
    }

    protected function defaults(): array|callable
    {
        return [
            'accessToken' => Ulid::generate(),
            'username' => self::faker()->userName(),
            'createdAt' => new \DateTimeImmutable('now'),
            'updatedAt' => new \DateTimeImmutable('now'),
        ];
    }
}
```

(3) So, after reading the docs, I did this:

```php
final class UserFactory extends PersistentProxyObjectFactory
{
    public function __construct()
    {
    }

    {
        return User::class;
    }

    {
        return $this
            ->instantiateWith(function (array $attributes) {
                return (new User(Uuid::v4()))
                    ->setAccessToken($attributes['accessToken'])
                    ->setUsername($attributes['username'])
                    ->setCreatedAt($attributes['createdAt'])
                    ->setUpdatedAt($attributes['updatedAt'])
                ;
            });
        ;
    }
}
```

(4) The previous code works, but then I cannot do this without adding more method calls in the `initialize()` method:

```php
UserFactory::createOne(['some_property' => 'some_value']);
```

(5) What I want is: define some default values AND pass constructor arguments

(6) The solution? The docs explain it but I totally overlooked it. You can do this:

```php
protected function defaults(): array|callable
{
    return [
        'uuid' => Uuid::v4(),
        'accessToken' => Ulid::generate(),
        'username' => self::faker()->userName(),
        'createdAt' => new \DateTimeImmutable('now'),
        'updatedAt' => new \DateTimeImmutable('now'),
    ];
}
```

In defaults you can pass **both** constructor arguments and properties!!

------

So, in this PR I want to make that feature **super clear**, so nobody skips it like I did.

kbond: > what do you mean by "will not be applied automatically"? I feel this is confusing.

When using a totally custom callback, the hydrator is not used. I think Javier expected it to just instantiate the object but still run it through the hydrator.

At one point I did have a PR to split these two concepts.

nikophil: ok, I understand better now! thanks for the explanations

you tried the version `instantiateWith(callable(...))`, which is the one I just never use. This one should only be used in rare cases when Foundry is not able to instantiate/hydrate the object, or if you want total control. Nonetheless, it is not rare to use  `instantiateWith()` with an `Instantiator` object as an argument, for which the statement "attributes will not be applied automatically" is false.

I think the solution for the docs, would be to separate the last special case from the code block above your warning, and add the warning only for this case

> In defaults you can pass both constructor arguments and properties!!

if this is not crystal clear in the docs, this may be highlighted!

nikophil:
I'm closing this, in favor of https://github.com/zenstruck/foundry/pull/765

Issue #752, opened 2024-12-16
Unable to create entity using 2.3.1


mvhirsch:
I'm trying to update to v2.3.0 (or v2.3.1) and facing this issue:

```
1) App\Tests\AppBundle\Admin\DDS\DdsItemAdminTest::showIdFilterIsApplied
Doctrine\DBAL\Exception\UniqueConstraintViolationException: An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '21' for key 'UNIQ_DDAA1CDA99AD6306'

/path/to/projectprotas/orf/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php:62
/path/to/projectprotas/orf/vendor/doctrine/dbal/src/Connection.php:1939
/path/to/projectprotas/orf/vendor/doctrine/dbal/src/Connection.php:1881
/path/to/projectprotas/orf/vendor/doctrine/dbal/src/Statement.php:194
/path/to/projectprotas/orf/vendor/doctrine/dbal/src/Statement.php:249
/path/to/projectprotas/orf/vendor/doctrine/orm/src/Persisters/Entity/BasicEntityPersister.php:285
/path/to/projectprotas/orf/vendor/doctrine/orm/src/UnitOfWork.php:1185
/path/to/projectprotas/orf/vendor/doctrine/orm/src/UnitOfWork.php:447
/path/to/projectprotas/orf/vendor/doctrine/orm/src/EntityManager.php:414
/path/to/projectprotas/orf/vendor/zenstruck/foundry/src/Persistence/PersistenceManager.php:105
/path/to/projectprotas/orf/vendor/zenstruck/foundry/src/Persistence/PersistenceManager.php:73
/path/to/projectprotas/orf/vendor/zenstruck/foundry/src/Persistence/PersistentObjectFactory.php:211
/path/to/projectprotas/orf/vendor/zenstruck/foundry/src/Persistence/PersistentProxyObjectFactory.php:41
/path/to/projectprotas/orf/vendor/zenstruck/foundry/src/Persistence/PersistentObjectFactory.php:286
/path/to/projectprotas/orf/vendor/zenstruck/foundry/src/Persistence/PersistentObjectFactory.php:214
/path/to/projectprotas/orf/vendor/zenstruck/foundry/src/Persistence/PersistentProxyObjectFactory.php:41
[...]

PR #755, opened 2024-12-16
fix: use a "placeholder" for inversed one-to-one

### LINKED ISSUE #749: Null value for inversed one-to-one relation (v2.3.1)
I'm upgrading Foundry on my project to use the v2.

However, it seems like the inversed relation is not auto filled, resulting in a null value (and violating the constraint on insertion), which was not the case before.

I thought [this issue](https://github.com/zenstruck/foundry/issues/655) would fix ours but it's not the case (since the latest version including the fix is not working).

For the reproducer, I created an Account and AccountSetting entity.

When trying to create an Account with a specified AccountSetting, the account property of the accountSetting is not filled (it is null).
```php
\App\Factory\AccountFactory::createOne([
  'accountSetting' => \App\Factory\AccountSettingFactory::new(),
]);
```

[Link to the reproducer test](https://github.com/florian-bailly/foundry-inversed-oneone-null/blob/main/tests/SampleTest.php)

nikophil:
:warning: https://github.com/zenstruck/foundry/pull/754 should be merged first

fixes https://github.com/zenstruck/foundry/issues/749 (and maybe https://github.com/zenstruck/foundry/issues/752)

~I still need to add a non-regression test~ DONE :blush:

### DIFF
--- a/src/Persistence/PersistentObjectFactory.php
+++ b/src/Persistence/PersistentObjectFactory.php
@@ -22,7 +22,7 @@ use Zenstruck\Foundry\ObjectFactory;
 use Zenstruck\Foundry\Persistence\Exception\NotEnoughObjects;
 use Zenstruck\Foundry\Persistence\Exception\RefreshObjectFailed;
 
-use function Zenstruck\Foundry\get;
+use function Zenstruck\Foundry\set;
 
 /**
  * @author Kevin Bond <kevinbond@gmail.com>
@@ -34,10 +34,13 @@ use function Zenstruck\Foundry\get;
  */
 abstract class PersistentObjectFactory extends ObjectFactory
 {
-    private bool $persist;
+    private PersistMode $persist;
 
     /** @phpstan-var list<callable(T, Parameters, static):void> */
     private array $afterPersist = [];
+
+    /** @var list<callable(T):void> */
+    private array $tempAfterInstantiate = [];
 
     /** @var list<callable(T):void> */
     private array $tempAfterPersist = [];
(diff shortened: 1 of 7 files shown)


src/Persistence/PersistentObjectFactory.php
```diff
@@ -34,11 +34,14 @@
...
 abstract class PersistentObjectFactory extends ObjectFactory
 {
-    private bool $persist;
+    private PersistMode $persist = PersistMode::PERSIST;
 
     /** @phpstan-var list<callable(T, Parameters, static):void> */
     private array $afterPersist = [];
 
+    /** @var list<callable(T):void> */
+    private array $tempAfterInstantiate = [];
```
nikophil, line 43: I don't like this solution... :shrug:

src/Persistence/PersistentObjectFactory.php
```diff
@@ -277,20 +296,21 @@ protected function normalizeParameter(string $field, mixed $value): mixed
...
-                    $value->create([$inverseField => $object]);
-                    $pm->refresh($object);
-                    $oldObj = get($inversedObject, $inverseField);
-                    delete($inversedObject);
-                    if ($oldObj) {
-                        delete($oldObj); // @phpstan-ignore argument.templateType
-                    }
+                // we need to handle the circular dependency involved by inversed one-to-one relationship:
+                // a placeholder object is used, which will be replaced by the real object, after its instantiation
+                $inversedObject = $value->withoutPersistingButScheduleForInsert()
```
nikophil, line 301: we're creating the "child" object here, but it needs the object created by the current factory, and we're passing it an empty "placeholder". So we should not persist now, or it try to save the place holder and it will break. 

Otherwise, we should "schedule for insert" the "child" object, as well as all other objects that would be created by the factory `$value`. 

Hence this "persist" mode stuff

I tried the other way around, meaning creating a placeholder for the "child" object, but I didn't manage to make it work

kbond (approved):

PR #758, opened 2024-12-17
[Docs fix] The incorrect "static" parameter type

GrinWay:
In PHP language there is no such a paramter type as "static"

I just removed it in the docs


nikophil:
this is wrong, [it has been added in PHP 8.0](https://wiki.php.net/rfc/static_return_type)

PR #759, opened 2024-12-17
[Docs fix] Factory::_real() instead Factory::object()

GrinWay:
Relying on docs I was trying to call such a method as "Factory::object()" but there was no this method.

Maybe the "Factory::_real()" was ment


nikophil:
hi @GrinWay 

thanks

PR #760, opened 2024-12-17
[Docs fix] Proxy::_save() instead of Proxy::save()

GrinWay:
Changed an incorrect function name in the docs

$post->save(); // persist the Post 
to
$post->_save(); // persist the Post


nikophil:
thanks @GrinWay

PR #762, opened 2024-12-17
tests: validate PSR-4 in CI

nikophil:
It's been several times that I detect and fix few classes that does not respect psr 4 in the `tests` directory.
Moreover, PHPStorm seems to be a little be buggy here... so I'm adding a check in the CI.


kbond:
Cool, yeah, I've noticed this too - just with tests... I've had to manually map this ditectory

PR #763, opened 2024-12-17
tests: transform "migrate" testsuite into "reset database" testsuite

nikophil:
This was my initial goal, when I split the test suite: to have a "reset database" testsuite, and a main test suite.

Another cool thing is that I've moved all the tests related to "global story" here. Each time I have some complex stuff to debug in the main testsuite, I always deactivate the global stories, because it creates some noise (extra pass in `create()` method, for example)

something I'm wondering is maybe we can now remove some permutations for the main testsuite in the CI. maybe remove sqlite and pgsql permutations, always use dama... WDYT? 

todo: add to the "reset database" matrix `--prefer-lowest` permutation


.gitignore
```diff
@@ -14,5 +14,4 @@
 /docker/.makefile
 /.env.local
 /docker-compose.override.yaml
-/tests/Fixture/MigrationTests/Migrations/
```
nikophil, line 17: generated migrations are now in `./var`

kbond (approved):
> maybe remove sqlite and pgsql permutations, always use dama

I think it's important to keep at least one sqlite permutation. For 3rd party bundle test suites (and I use for some SymfonyCasts tutorials) it's just easier to use.

Issue #710, opened 2024-10-28
Bug: doctrine identity problem when using `$this` in a method of proxified objects


KDederichs:
Hey,

I just updated to 2.2 from 1.38 and I found an issue where foundry doesn't correctly handle objects that are created by proxy objects.

Given that you have a class `Foo`:
```php
class Foo
{

    #[OneToMany(mappedBy: 'foo', targetEntity: FooBar::class, cascade: ['persist', 'remove'], fetch: 'EAGER', orphanRemoval: true)]
    private Collection $fooBarCollection;

    public function addBarToFoo(Bar $bar): self 
    {
        $fooBar = new FooBar();
        $fooBar->setBar($bar)->setFoo($this);

        return $this.
    }

    public function createABarAndAddToFoo(): self 
    {
        $bar = new Bar():
        return $this->addBarToFoo($bar);
    }
}

class Bar
{
    //something
}

class FooBar {
    
    #[ManyToOne(targetEntity: Foo::class, inversedBy: 'fooBarCollection')]
    private Foo $foo;

    #[ManyToOne(targetEntity: Bar::class, cascade: ['persist'], fetch: 'EAGER')]  
    private Bar $bar;

    public function setFoo(Foo $foo): self
    {
        $this->foo = $foo;
    
        return $this;
    }

    public function setBar(Bar $bar): self
    {
        $this->bar = $bar;

        return $this;
    }
}
```

When you create a new `Foo` using `FooFactory::createOne()` and then call `createABarAndAddToFoo()` and `_save()` on the created `Foo` object this will result in an error like this:

```
Doctrine\ORM\ORMInvalidArgumentException: A new entity was found through the relationship 'App\Entity\FooBar#foo' that was not configured to cascade persist operations for entity: XXX. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}).
```

This error obviously should happen since the `Foo` object has been created by the factory earlier and thus exists/is persisted.
This was not an issue in the `1.x` version of Foundry.

You can 'fix' it by prefixing the `createABarAndAddToFoo` call with a `_real()` but I'd rather have it work correctly :)

nikophil:
Hello,

please, could you share your factories?

KDederichs:
https://github.com/KDederichs/sf-reproducer/tree/foundy/proxy_save here's a reproducer for that, just default factories as they come out of the box

[...]

PR #764, opened 2024-12-17
chore: add test to check if big int datatype messes with the proxy

KDederichs:
related to #710


nikophil:
Aaaand tests are passing 😅

(The failure comes from a namespace problem)

KDederichs:
yeah I'm trying to figure out if it's a Postgres thing... but something's wrong with those, can't figure out what atm :D

KDederichs:
There you go, it only happens when you use an UUID as primary key... figures (no clue how to fix the other tests, not sure what's wrong with them)

.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           - {php: 8.3, symfony: '*', phpunit: 9, database: none}
           - {php: 8.3, symfony: '*', phpunit: 11, database: mysql|mongo}
           - {php: 8.3, symfony: '*', phpunit: 11, database: pgsql|mongo}
+          - {php: 8.3, symfony: '*', phpunit: 9, database: pgsql}
```
nikophil, line 34: seems unneeded

tests/Fixture/Factories/BigIntProxyFactory.php
```diff
@@ -0,0 +1,34 @@
...
+
+use Zenstruck\Foundry\Tests\Fixture\Entity\BigIntEntity;
+use Zenstruck\Foundry\Persistence\PersistentProxyObjectFactory;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ *
+ * @extends PersistentProxyObjectFactory<BigIntEntity>
+ */
+final class BigIntProxyFactory extends PersistentProxyObjectFactory
```
nikophil: FYI, you can use [anonymous factories](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#anonymous-factories)

nikophil (commented):
this is very strange, that it only fails when uuid is involved :shrug:

nikophil:
ok then, it has nothing do to with uuid. If you do this change:
```diff
- self::assertEquals($entity->getId(), $entity->getId());
+ self::assertEquals($entity->getBigIntVal(), $entity->getBigIntVal());
```
it will fail regardeless you use uuid or not.

And it receives an int from the database, so it thinks the original data is an int... this looks like a little quirk in the ORM.

I'm afraid there is nothing we can do in Foundry. I'll close this PR, and keep the issue for the record.

KDederichs:
Yeah I suspect it might be a doctrine issue, cause it's supposed to load that as string according to them, I guess I'll bring it up there and see what they say about it.

PR #765, opened 2024-12-18
[Doc] Added some docs to clarify default attributes and fixed some syntax issues

nikophil:
here is a proposal based on @javiereguiluz 's [PR](https://github.com/zenstruck/foundry/pull/753) - I've separated the special method `instantiateWith(callable(...))`, so that it's clear that it works in a different way than the others


kbond (approved):

javiereguiluz:
I couldn't review this earlier, but I just did and your changes are very nice. Nicolas, thanks a lot for taking care of this!

PR #766, opened 2024-12-18
feat: throw an error if Factories trait is not used in a KernelTestCase

### LINKED ISSUE #750: Using factories in a `KernelTestCase` without `Factories` trait should throw
Hi , 
 I have created a test class . I use foundry methods to first create and then search for the created users and create connections between them. However I see that entities that are searched are not managed. I then need to "re-search" them in order for them to become managed. 

```php
<?php

namespace App\Tests\Repository;

use App\Entity\Connection;
use App\Entity\User;
use App\Factory\UserFactory;
use App\Repository\ConnectionRepository;
use App\Tests\Fixtures\UserFixtures;
use Doctrine\ORM\EntityManager;
use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;

class ConnectionRepositoryTest extends KernelTestCase
{
    private ?EntityManager $entityManager;
    private ConnectionRepository $repository;
    use UserFixtures;

    {
        $kernel = self::bootKernel();

        $this->entityManager = $kernel->getContainer()
            ->get('doctrine')
            ->getManager();

        $this->repository = $this->entityManager->getRepository(Connection::class);
    }

    public function testDeleteConnection()
    {

    }

    public function testCreateConnection()
    {

        $this->createUsers();

        $a = $this->entityManager->contains($this->mainUser->_real());
        $b = $this->entityManager->contains(UserFactory::find(['id' => $this->mainUser->getId()])->_real());

        // need to
        $userA = $this->entityManager->getRepository(User::class)->find($this->mainUser->getId());
        $userB = $this->entityManager->getRepository(User::class)->find($this->users[0]->getId());

        $c = $this->entityManager->contains($userA);

        $this->repository->createConnection($userA, $userB);

        self::assertNotNull($this->repository->findOneBy(['user' => $userA, 'connectedToUser' => $userB]));
        self::assertNotNull($this->repository->findOneBy(['connectedToUser' => $userB, 'user' => $userA]));

    }
}

```

The fixture class is simple 
```php
<?php

namespace App\Tests\Fixtures;

use App\Factory\UserFactory;

trait UserFixtures
{
    public array $users;
    /**
     * @var \App\Entity\User|mixed|object|(object&\Zenstruck\Foundry\Persistence\Proxy)|\Zenstruck\Foundry\Persistence\Proxy
     */
    public mixed $mainUser;

    public function createUsers()
    {
        $this->mainUser = UserFactory::createOne(['roles' => ['ROLE_USER']]);

        $this->users = UserFactory::createMany(10, ['roles' => ['ROLE_USER']]);

    }
}
```

Here are the result of debugging while testing 
image

If I pass _real() values to the repository like below 
```php
 public function testCreateConnection()
    {

        $this->createUsers();

        $a = $this->entityManager->contains($this->mainUser->_real());
        $b = $this->entityManager->contains(UserFactory::find(['id' => $this->mainUser->getId()])->_real());

        // need to
        $userA = $this->entityManager->getRepository(User::class)->find($this->mainUser->getId());
        $userB = $this->entityManager->getRepository(User::class)->find($this->users[0]->getId());

        $c = $this->entityManager->contains($userA);

        $this->repository->createConnection($this->mainUser->_real(), $this->users[0]->_real());

        self::assertNotNull($this->repository->findOneBy(['user' => $userA, 'connectedToUser' => $userB]));
        self::assertNotNull($this->repository->findOneBy(['connectedToUser' => $userB, 'user' => $userA]));

    }
```
I get the error 
`
* A new entity was found through the relationship 'App\Entity\Connection#connectedToUser' that was not configured to cascade persist operations for entity: App\Entity\User@818. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}). If you cannot find out which entity causes the problem implement 'App\Entity\User#__toString()' to get a clue.
 * A new entity was found through the relationship 'App\Entity\Connection#user' that was not configured to cascade persist operations for entity: App\Entity\User@986. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}). If you cannot find out which entity causes the problem implement 'App\Entity\User#__toString()' to get a clue.

`
If I pass entities passed after finding them thru entities , then the test passes
image

Is there a workaround to making entities managed upon searching using Foundry method ? 

Thanks and regards
Ashvini

nikophil:
fixes #750 

This "feature" was actually very challenging! :exploding_head:

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -7,6 +7,13 @@ $file = __DIR__.'/.php-cs-fixer.temp.php';
     \file_get_contents('https://raw.githubusercontent.com/zenstruck/.github/main/.php-cs-fixer.dist.php')
 );
 
+$finder = PhpCsFixer\Finder::create()
+    ->in([__DIR__.'/src', __DIR__.'/tests'])
+
+    // we want to keep the traits in "wrong order"
+    ->notName('KernelTestCaseWithBothTraitsInWrongOrderTest.php')
+;
+
 try {
     return require $file;
 } finally {
--- a/src/Exception/FactoriesTraitNotUsed.php
+++ b/src/Exception/FactoriesTraitNotUsed.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the zenstruck/foundry package.
+ *
+ * (c) Kevin Bond <kevinbond@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Zenstruck\Foundry\Exception;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Zenstruck\Foundry\Test\Factories;
+
+/**
+ * @author Nicolas PHILIPPE <nikophil@gmail.com>
+ */
+final class FactoriesTraitNotUsed extends \LogicException
+{
+    /**
+     * @param class-string<KernelTestCase> $class
+     */
+    private function __construct(string $class)
+    {
+        parent::__construct(
+            \sprintf('You must use the trait "%s" in "%s" in order to use Foundry.', Factories::class, $class)
+        );
+    }
+
+    public static function throwIfComingFromKernelTestCaseWithoutFactoriesTrait(): void
+    {
+        $backTrace = \debug_backtrace(\DEBUG_BACKTRACE_IGNORE_ARGS); // @phpstan-ignore ekinoBannedCode.function
+
+        foreach ($backTrace as $trace) {
+            if (
+                '->' === ($trace['type'] ?? null)
+                && isset($trace['class'])
+                && KernelTestCase::class !== $trace['class']
+                && \is_a($trace['class'], KernelTestCase::class, allow_string: true)
+                && !(new \ReflectionClass($trace['class']))->hasMethod('_bootFoundry')
+            ) {
+                self::throwIfClassDoesNotHaveFactoriesTrait($trace['class']);
+            }
+        }
+    }
+
+    /**
+     * @param class-string<KernelTestCase> $class
+     */
+    public static function throwIfClassDoesNotHaveFactoriesTrait(string $class): void
+    {
+        if (!(new \ReflectionClass($class))->hasMethod('_bootFoundry')) {
+            // throw new self($class);
+            trigger_deprecation(
+                'zenstruck/foundry',
+                '2.4',
+                'In order to use Foundry, you must use the trait "%s" in your "%s" tests. This will throw an exception in 3.0.',
+                KernelTestCase::class,
+                $class
+            );
+        }
+    }
+}
--- a/tests/Integration/ForceFactoriesTraitUsage/KernelTestCaseWithBothTraitsInWrongOrderTest.php
+++ b/tests/Integration/ForceFactoriesTraitUsage/KernelTestCaseWithBothTraitsInWrongOrderTest.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the zenstruck/foundry package.
+ *
+ * (c) Kevin Bond <kevinbond@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Zenstruck\Foundry\Tests\Integration\ForceFactoriesTraitUsage;
+
+use PHPUnit\Framework\Attributes\RequiresPhpunit;
+use PHPUnit\Framework\Attributes\Test;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Zenstruck\Foundry\Test\Factories;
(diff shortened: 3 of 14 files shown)


src/Test/ResetDatabase.php
```diff
@@ -34,7 +35,10 @@ public static function _resetDatabaseBeforeFirstTest(): void
 
         ResetDatabaseManager::resetBeforeFirstTest(
             static fn() => static::bootKernel(),
-            static fn() => static::ensureKernelShutdown(),
+            static function () {
+                Configuration::shutdown();
+                static::ensureKernelShutdown();
+            },
```
nikophil, line 41: I'm wondering if we should shutdown foundry on kernel shutdown, WDYT?

.php-cs-fixer.dist.php
```diff
@@ -7,6 +7,11 @@
     \file_get_contents('https://raw.githubusercontent.com/zenstruck/.github/main/.php-cs-fixer.dist.php')
 );
 
+$finder = PhpCsFixer\Finder::create()
+    ->in([__DIR__.'/src', __DIR__.'/tests'])
+    ->notPath('ForceFactoriesTraitUsage')
```
nikophil, line 12: for some reasons, I need to keep tests with prefix `test` and not with annotation `@test` in the end to end tests

tests/Integration/ForceFactoriesTraitUsage/KernelTestCaseWithBothTraitsInWrongOrderTest.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Integration\ForceFactoriesTraitUsage;
+
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Zenstruck\Foundry\Test\Factories;
+use Zenstruck\Foundry\Test\ResetDatabase;
+use Zenstruck\Foundry\Tests\Fixture\Factories\Object1Factory;
+
+final class KernelTestCaseWithBothTraitsInWrongOrderTest extends KernelTestCase
+{
+    use ResetDatabase, Factories;
```
nikophil, line 26: in my different tests, I noticed that sometimes, the order of these traits could have an impact, so I wanted to test with both cases

tests/Integration/ForceFactoriesTraitUsage/kernel_test_case_with_only_reset_database_trait.phpt
```diff
@@ -0,0 +1,38 @@
+--TEST--
+phpunit --version
+--FILE--
+<?php
+$_ENV['KERNEL_CLASS'] = 'Zenstruck\Foundry\Tests\Fixture\TestKernel';
+$_SERVER['argv'][] = '--do-not-cache-result';
+$_SERVER['argv'][] = '--no-configuration';
+$_SERVER['argv'][] = __DIR__ . '/../../Fixture/ForceFactoriesTraitUsage/KernelTestCaseWithOnlyResetDatabaseTraitTest.php';
```
nikophil, line 8: for this specific case, where only the `ResetDatabase` trait is used, the error occurs during the "before test class" hook, so the error cannot be caught in a regular test . Hence this low level ".phpt" test

kbond (approved):

nikophil:
@kbond I think I won't throw directly an error, but add a deprecation instead: Foundry kinda works without the trait in few situations... WDYT?

src/Exception/FactoriesTraitNotUsed.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        if (!(new \ReflectionClass($class))->hasMethod('_bootFoundry')) {
+            //            throw new self($class);
+            trigger_deprecation(
+                'zenstruck/foundry',
+                '2.4',
+                'In order to use Foundry, you must use the trait "%s" in your "%s" tests. This will throw an exception in 3.0.',
+                KernelTestCase::class,
+                $class
+            );
```
nikophil, line 64: at first I built everything to throw an error, but now I think we should not throw, it's a little bit too much violent I think :laughing: :boxing_glove: 

But since everything is ready to throw, I propose to keep it like this in the exception

tests/Integration/ForceFactoriesTraitUsage/KernelTestCaseWithBothTraitsInWrongOrderTest.php
```diff
@@ -0,0 +1,45 @@
...
+namespace Zenstruck\Foundry\Tests\Integration\ForceFactoriesTraitUsage;
+
+use PHPUnit\Framework\Attributes\RequiresPhpunit;
+use PHPUnit\Framework\Attributes\Test;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Zenstruck\Foundry\Test\Factories;
+use Zenstruck\Foundry\Test\ResetDatabase;
+use Zenstruck\Foundry\Tests\Fixture\Factories\Object1Factory;
+
+#[RequiresPhpunit('>=11.0')]
```
nikophil, line 23: only PhpUnit 11 handles correctly deprecation, so all tests in this namespace will only be ran with phpunit 11

PR #781, opened 2025-01-06
test: assert updates are implicitly persisted

nikophil:
fix of a subtle bug introduced by https://github.com/zenstruck/foundry/pull/773


kbond (approved):
Interesting solution!

PR #787, opened 2025-01-15
fix: Ensure that nested factories can be created even if class already exists

### LINKED ISSUE #786: [make:factory] Problems with `--test` flag
I'm honestly know sure, if I'm misunderstanding something, doing it wrong or it a bug. 

Lets say I have the entity setup from documentation. https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#same-entities-used-in-these-docs

A post and a comment. 

If I create factories with the `bin/console make:factory --test Post` flag it complains that the `CommentFactory` is not present and as if it should create it for me. If I said yes, it creates it's in the `src/Factory` directory even if I use the `--test` flag. 

If it's already is created in `tests/Facoty` is still complains that it's not present as it looks for it in `src/Factory/CommentFactory.php` but when I answer yes to create if fails, as it's already created.

Is this something that I misunderstand from concept or is this a bug? 

Thanks for your help.

tomasnorre:
This is intended to resolve #786, which isn't the case yet. Right now it's only a test proving the problem described in the issue. 

As I don't know foundry that well, I have a little difficulties, with the naming as I don't know the terminology used in foundry. So inputs are welcome, but we get to a review state at some point anyway. :) 

Update: There is something not right in my test. There order should be different, and then it works. Back to the IDE..

### DIFF
--- a/src/Maker/Factory/FactoryGenerator.php
+++ b/src/Maker/Factory/FactoryGenerator.php
@@ -97,6 +97,9 @@ final class FactoryGenerator
             }
         }
 
+        // Should never do this if file/class already exists.
+        // A class_exists($factoryClass), return false, I don't know if it's not part of the autoloader, that the class is not found.
+        // I suspect that it's part of the problem.
         $generator->generateClass(
             $factoryClass,
             __DIR__.'/../../../skeleton/Factory.tpl.php',
--- a/tests/Integration/Maker/MakeFactoryTest.php
+++ b/tests/Integration/Maker/MakeFactoryTest.php
@@ -129,10 +129,17 @@ final class MakeFactoryTest extends MakerTestCase
         }
 
         $tester = $this->makeFactoryCommandTester();
-        $tester->execute(['class' => Contact::class, '--test' => true]);
-        $tester->execute(['class' => Address::class, '--test' => true]);
+        
+        $tester->execute(['class' => Contact::class, '--test' => true]);    
+        
         $this->assertFileExists(self::tempFile('tests/Factory/ContactFactory.php'));
         $this->assertFileExists(self::tempFile('tests/Factory/AddressFactory.php'));
+        
+        // here, we're faking multiple calls to `bin/console make:factory`
+        // so it's perfectly acceptable to manually load the newly created class
+        require_once self::tempFile('tests/Factory/AddressFactory.php';
+        
+        $tester->execute(['class' => Address::class, '--test' => true]);
     }
 
     /**
--- a/tests/Integration/Maker/MakeFactoryTest.php
+++ b/tests/Integration/Maker/MakeFactoryTest.php
@@ -129,16 +129,15 @@ final class MakeFactoryTest extends MakerTestCase
         }
 
         $tester = $this->makeFactoryCommandTester();
-        
-        $tester->execute(['class' => Contact::class, '--test' => true]);    
-        
+
+        $tester->execute(['class' => Contact::class, '--test' => true]);
+
         $this->assertFileExists(self::tempFile('tests/Factory/ContactFactory.php'));
         $this->assertFileExists(self::tempFile('tests/Factory/AddressFactory.php'));
         
         // here, we're faking multiple calls to `bin/console make:factory`
         // so it's perfectly acceptable to manually load the newly created class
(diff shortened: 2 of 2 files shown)


src/Maker/Factory/FactoryGenerator.php
```diff
@@ -97,6 +97,9 @@ function(string $newClassName) use ($factoryClass) {
             }
         }
 
+        // Should never do this if file/class already exists.
+        // A class_exists($factoryClass), return false, I don't know if it's not part of the autoloader, that the class is not found.
+        // I suspect that it's part of the problem.
         $generator->generateClass(
```
tomasnorre, line 103: How I read the code, we need a guard, so this is not executed, when the class already exits. And it does second time the `$tester->execute(['class' => Address::class, '--test' => true]);` is called.

tests/Integration/Maker/MakeFactoryTest.php
```diff
@@ -118,6 +119,22 @@ public function can_create_factory_in_test_dir(): void
...
+    {
+        if (!\getenv('DATABASE_URL')) {
+            self::markTestSkipped('doctrine/orm not enabled.');
+        }
+
+        $tester = $this->makeFactoryCommandTester();
+        $tester->execute(['class' => Contact::class, '--test' => true]);
+        $tester->execute(['class' => Address::class, '--test' => true]);
+        $this->assertFileExists(self::tempFile('tests/Factory/ContactFactory.php'));
+        $this->assertFileExists(self::tempFile('tests/Factory/AddressFactory.php'));
```
nikophil: I think you can change your code to this:
```suggestion
        $tester = $this->makeFactoryCommandTester();
        
        $tester->execute(['class' => Contact::class, '--test' => true]);    
        
        $this->assertFileExists(self::tempFile('tests/Factory/ContactFactory.php'));
        $this->assertFileExists(self::tempFile('tests/Factory/AddressFactory.php'));
        
        // here, we're faking multiple calls to `bin/console make:factory`
        // so it's perfectly acceptable to manually load the newly created class
        require_once self::tempFile('tests/Factory/AddressFactory.php';
        
        $tester->execute(['class' => Address::class, '--test' => true]);
```

src/Maker/Factory/FactoryGenerator.php
nikophil: (see my other comment about autoloading)

I don't see how this could fix your initial problem :thinking: 

My thoughts about your problem is that by adding `--test`, you'd expect Foundry to check the existence of the factories in `tests` directory, but we don't do this, since this check is done thanks to `foundry.factory` tag, and this tag is not added implicitly for the services in `tests` directory. (for the record, the tag is added automatically for all classes loaded in the container, [which extend `Factory` class](https://github.com/zenstruck/foundry/blob/2.x/src/ZenstruckFoundryBundle.php#L195-L196)).

I'm wondering if you do something like this, it would not resolve this problem:
```yaml
# config/packages/zenstruck_foundry.yaml
when@dev: &dev
    zenstruck_foundry: ~
    services:
        Tests\Fixtures\Factories:
            resource: '../../tests/Fixtures/Factories'

when@test: *dev
```
this is kinda a bit strange, because, it would load some tests classes in `dev` environment. But doing `bin/console make:factory --test` uses `dev` environment :shrug: And of course, we could not add this kind of strangeness to Foundry :sweat_smile: 

I think there is not an easy solution. Maybe we should deprecate `--test` option and tell people that if they want to create factories in `tests` dir, they should use `--env=test`. Or maybe we should add a configuration node in which users can declare where they store their factories, so that we can tag them (or maybe no need to tag, at least add them in another way to the `FactoryMap`)

any thoughs @kbond?

nikophil (changes requested):

src/Maker/Factory/FactoryGenerator.php
nikophil: Hi @tomasnorre 

did you try my suggestion where those factories in `tests` are autoloaded in your `dev` environment? we talked a little bit about this with @kbond and think this could be an easy fix for this, that we could promote in the documentation + the recipes.

tomasnorre: I didn't get to it yet. Will try it out.

nikophil:
I'm closing this stale PR

PR #790, opened 2025-01-18
feat: introduce Foundry events

nikophil:


src/Hooks/AsAfterInstantiateFoundryHook.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * (c) Kevin Bond <kevinbond@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Zenstruck\Foundry\Hooks;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
```
nikophil: in a first time, I think this is enough.
Maybe in a second time, if someone needs it, we could implement `TARGET_METHOD` and `IS_REPEATABLE`, but this will complexify a lot the DI

src/ZenstruckFoundryBundle.php
```diff
@@ -192,13 +202,46 @@ public function configure(DefinitionConfigurator $definition): void
...
+        $container->registerForAutoconfiguration(Story::class)->addTag('foundry.story');
+
+        $hookValidator = static function(\ReflectionClass $reflector, string $hookEventClass, object $attribute): void {
+            if (
+                !$reflector->hasMethod('__invoke')
+                || 1 !== \count($parameters = $reflector->getMethod('__invoke')->getParameters())
+                || !($typeFirstParameter = $parameters[0]->getType())
+                || (string) $typeFirstParameter !== $hookEventClass
+            ) {
+                throw new InvalidArgumentException(\sprintf('In order to be declared "#[%s]", the class "%s" must have an "__invoke" method with one unique parameter of type "%s".', (new \ReflectionClass($attribute))->getShortName(), $reflector->getName(), $hookEventClass));
```
nikophil, line 216: I wondered if I should introduce an interface, but, that would mean 3 interface

not sure what's the  best :shrug:

src/ZenstruckFoundryBundle.php
```diff
@@ -294,6 +337,22 @@ public function process(ContainerBuilder $container): void
...
+        // hooks
+        $iterators = [];
+        foreach ($container->findTaggedServiceIds('foundry.hook') as $id => $tags) {
+            $index = isset($tags[0]['class'])
+                ? HooksRegistry::hookClassSpecificIndex($tags[0]['hook'], $tags[0]['class'])
+                : $tags[0]['hook'];
+
+            $iterators[$index] ??= [];
+            $iterators[$index][] = new Reference($id);
+        }
```
nikophil: we're kinda missing a priority here. But that's not straightforward to do, because "global hook" and "hook specific to a class" are not in the same iterator.

I propose to leave this like this, and see if someone ever asks for it :)

tests/Fixture/Hooks/AddressAfterInstantiateHook.php
```diff
@@ -0,0 +1,34 @@
...
+     */
+    public function __invoke(AfterInstantiate $hook): void
+    {
+        if ($hook->factory instanceof AddressFactoryWithHook) {
+            $hook->object->setCity(
+                'AddressAfterInitialisationHook '.$hook->object->getCity()
+            );
+        }
+    }
+}
```
nikophil: here is an example of a hook.

another syntax is possible:
```php
#[AsFoundryHook(event: AfterInstantiate::class, class: Address::class)]
final class AddressAfterInstantiateHook
{
    public function __invoke(AfterInstantiate $hook): void;
}
```
it would be clear then, that you should pass a `AfterInstantiate` to the `__invoke()` method

tests/Integration/Migration/ResetDatabaseWithMigrationTest.php
```diff
@@ -45,16 +45,9 @@ public function it_generates_valid_schema(): void
         $application = new Application(self::bootKernel());
         $application->setAutoExit(false);
 
-        $exit = $application->run(
-            new ArrayInput(['command' => 'doctrine:schema:validate', '-v' => true]),
-            $output = new BufferedOutput()
-        );
-
-        // The command actually fails, because of a bug in doctrine ORM 3!
-        // https://github.com/doctrine/migrations/issues/1406
```
nikophil, line 54: the bug is fixed :tada: 
https://github.com/doctrine/migrations/issues/1406#issuecomment-2596839269

tests/Fixture/Hooks/ObjectAfterInstantiateHook.php
```diff
@@ -0,0 +1,30 @@
...
+final class ObjectAfterInstantiateHook
+{
+    /**
+     * @param AfterInstantiate<EntityForHooks> $hook
+     */
+    public function __invoke(AfterInstantiate $hook): void
+    {
+        $hook->object->name = "{$hook->object->name}\nObjectAfterInstantiateHook";
+    }
+}
```
nikophil: here is an example of a hook.

another syntax is possible:
```php
#[AsFoundryHook(event: AfterInstantiate::class(, class: EntityForHooks::class))]
final class AddressAfterInstantiateHook
{
    public function __invoke(AfterInstantiate $hook): void;
}
```
it would be clear then, that you should pass a `AfterInstantiate` to the `__invoke()` method.

I also wondered if we should not introduce an interface, but, that would mean 3 interfaces. And an attribute AND an interface seems redundant.

WDYT?

kbond: I'd really prefer we didn't require the event argument - could be inferred from the type-hint.

src/Event/AfterInstantiate.php
```diff
@@ -0,0 +1,34 @@
...
+
+use Zenstruck\Foundry\Factory;
+use Zenstruck\Foundry\ObjectFactory;
+
+/**
+ * @author Nicolas PHILIPPE <nikophil@gmail.com>
+ *
+ * @phpstan-import-type Parameters from Factory
+ */
+final class AfterInstantiate
```
nikophil: I hesitated to use `@template T` here, but this would force users to use it in listeners, and they will be tempted to do something like this:
```php
final class AddressEventListener
{
    /** @param BeforeInstantiate<Address> */
    #[AsEventListener]
    public function beforeInstantiate(BeforeInstantiate $event): void
    {
        if (Address::class !== $event->objectClass) {
            return;
        }

        // ...
    }
```
but phpstan will break on the if...

tests/Fixture/Events/FoundryEventListener.php
```diff
@@ -0,0 +1,53 @@
...
+use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
+use Zenstruck\Foundry\Event\AfterInstantiate;
+use Zenstruck\Foundry\Event\AfterPersist;
+use Zenstruck\Foundry\Event\BeforeInstantiate;
+use Zenstruck\Foundry\Tests\Fixture\Entity\EntityForEventListeners;
+
+final class FoundryEventListener
+{
+    #[AsEventListener]
+    public function beforeInstantiate(BeforeInstantiate $event): void
```
nikophil, line 25: I'm wondering if we cannot still introduce an attribute `#[BeforeInstantiate(Address::class)]`?

the DX will be a little bit better

src/Event/AfterInstantiate.php
kbond: Yeah, let's leave off for now

src/Event/AfterPersist.php
```diff
@@ -0,0 +1,34 @@
...
+/*
+ * This file is part of the zenstruck/foundry package.
+ *
+ * (c) Kevin Bond <kevinbond@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Zenstruck\Foundry\Event;
```
kbond:
```suggestion
namespace Zenstruck\Foundry\Persistence\Event;
```

?

kbond (commented):
Looks good. Could you add some doc's?

src/Event/AfterPersist.php
nikophil: I moved other ones to `Zenstruck\Foundry\Object\Event` as well

PR #797, opened 2025-01-24
chore(ci): misc improvments in CI permutations

### LINKED ISSUE #783: SQLite and ResetDatabase trait: deletion of SQL database file fails
I'm using the ResetDatabase with Doctrine/SQLite backend. 

Because ResetDatabase did not seem to do anything, I found this code to be executed in BaseOrmResetter while debugging: 

```
if (DoctrineOrmVersionGuesser::isOrmV3() && $dbPath && $fs->exists($dbPath)) {
    (new Filesystem())->remove($dbPath);
}
```

Remove is never called, because DoctrineOrmVersionGuesser::isOrmV3() returns false. 

Code for isOrmV3: 

```
public static function isOrmV3(): bool
{
        return \class_exists(FieldMapping::class);
}
```

What can I do, to make it work? Is it some kind of bug or is it an configuration error?

### LINKED ISSUE #796: Improve CI matrix
- add to the "reset database" matrix --prefer-lowest permutation
- remove some permutation after merging https://github.com/zenstruck/foundry/pull/763

nikophil:
fixes #796 
fixes #783

- remove some tests permutations for main test suite:
	- only test with dama when possible (all the "without dama" permutations are only needed for "reset database" mechanism)
	- only keep one permutation with php 8.1
	- only one permutation with phpunit 9 and 10
- add test with lowest deps in reset database test suite

on `2.x` branch, we currently have 31 permutations for reset database testsuite and 48 for main testsuite, running in 9m50

and now we have 35 permutations for reset database testsuite and 24 for main testsuite, running in 4m21 :blush:


src/ORM/ResetDatabase/BaseOrmResetter.php
```diff
@@ -68,7 +67,7 @@ final protected function dropAndResetDatabase(Application $application): void
                 // let's only drop the .db file
 
                 $dbPath = $connection->getParams()['path'] ?? null;
-                if (DoctrineOrmVersionGuesser::isOrmV3() && $dbPath && (new Filesystem())->exists($dbPath)) {
```
nikophil, line 71: this test on orm version seems to not be needed anymore

tests/Fixture/Entity/EdgeCases/InversedOneToOneWithNonNullableOwning/InverseSide.php
```diff
@@ -20,7 +20,7 @@
  * @author Nicolas PHILIPPE <nikophil@gmail.com>
  */
 #[ORM\Entity]
-#[ORM\Table('inversed_one_to_one_with_non_nullable_owning_inverse_side')]
+#[ORM\Table('inversed_one_to_one_non_nullable_owning_inverse_side')]
```
nikophil, line 23: the name was too long for PG sequences :sweat_smile:

PR #798, opened 2025-01-26
feat: skip readonly properties on entities when generating factories

### LINKED ISSUE #769: [Maker] Don't try to set read only properties
Little bit of an DX improvement:

Every time I generate new factories using the make:facrory command it'll try to set `createdAt` by default even though that value has no setter (it's set in the constructor)

It would be awesome if the maker could skip over those read only properties so I don't have to go through all the factories manually and delete it :D

KDederichs:
Fixes #769 by checking if the property is writable or settable via constructor 

Things that are missing:

- [x] Test cases (as soon as I figure out how those maker tests work)
- [x] `always_force_property` (I guess that needs to be injected there since it's only present in Hydrator and Instantiator)

If you got some pointers for those two points, I'd gladly take those

### DIFF
--- a/src/Maker/Factory/FactoryGenerator.php
+++ b/src/Maker/Factory/FactoryGenerator.php
@@ -32,9 +32,6 @@ final class FactoryGenerator
     public const PHPSTAN_PATH = '/vendor/phpstan/phpstan/phpstan';
     public const PSALM_PATH = '/vendor/vimeo/psalm/psalm';
 
-    /** @var string[]|true */
-    private array|bool $forceProperties = [];
-
     /** @param \Traversable<int, DefaultPropertiesGuesser> $defaultPropertiesGuessers */
     public function __construct(
         private ?PersistenceManager $persistenceManager,
@@ -42,6 +39,7 @@ final class FactoryGenerator
         private \Traversable $defaultPropertiesGuessers,
         private FactoryClassMap $factoryClassMap,
         private NamespaceGuesser $namespaceGuesser,
+        private bool $forceProperties = false
     ) {
     }
 
@@ -164,13 +162,5 @@ final class FactoryGenerator
             \file_exists($this->kernel->getProjectDir().self::PSALM_PATH) => MakeFactoryData::STATIC_ANALYSIS_TOOL_PSALM,
             default => MakeFactoryData::STATIC_ANALYSIS_TOOL_NONE,
         };
-    }
-
-    public function alwaysForce(string ...$properties): self
-    {
-        $clone = clone $this;
-        $clone->forceProperties = $properties ?: true;
-
-        return $clone;
     }
 }
--- a/src/Maker/Factory/FactoryGenerator.php
+++ b/src/Maker/Factory/FactoryGenerator.php
@@ -32,6 +32,9 @@ final class FactoryGenerator
     public const PHPSTAN_PATH = '/vendor/phpstan/phpstan/phpstan';
     public const PSALM_PATH = '/vendor/vimeo/psalm/psalm';
 
+    /** @var string[]|true */
+    private array|bool $forceProperties = [];
+
     /** @param \Traversable<int, DefaultPropertiesGuesser> $defaultPropertiesGuessers */
     public function __construct(
         private ?PersistenceManager $persistenceManager,
@@ -150,6 +153,7 @@ final class FactoryGenerator
             $this->staticAnalysisTool(),
             $persisted ?? false,
             $makeFactoryQuery->addPhpDoc(),
+            $this->forceProperties
         );
     }
 
@@ -160,5 +164,13 @@ final class FactoryGenerator
             \file_exists($this->kernel->getProjectDir().self::PSALM_PATH) => MakeFactoryData::STATIC_ANALYSIS_TOOL_PSALM,
             default => MakeFactoryData::STATIC_ANALYSIS_TOOL_NONE,
         };
+    }
+
+    public function alwaysForce(string ...$properties): self
+    {
+        $clone = clone $this;
+        $clone->forceProperties = $properties ?: true;
+
+        return $clone;
     }
 }
--- a/src/Maker/Factory/MakeFactoryData.php
+++ b/src/Maker/Factory/MakeFactoryData.php
@@ -158,7 +158,7 @@ final class MakeFactoryData
     public function getDefaultProperties(): array
     {
         $defaultProperties = $this->defaultProperties;
-        $clazz = $this->object->getName();
+        $class = $this->object->getName();
 
         /**
          * If forceProperties is not set we filter out properties that can not be set because they're either readonly or have no setter.
@@ -166,12 +166,12 @@ final class MakeFactoryData
          *
          * We do this here because we need to get the class of the Entity which only seems to be accessible here.
          */
-        $defaultProperties = array_filter($defaultProperties, function (string $propertyName) use ($clazz): bool {
+        $defaultProperties = array_filter($defaultProperties, function (string $propertyName) use ($class): bool {
             if (true === $this->forceProperties) {
                 return true;
             }
 
-            return self::propertyInfo()->isWritable($clazz, $propertyName) || self::propertyInfo()->isInitializable($clazz, $propertyName);
+            return self::propertyInfo()->isWritable($class, $propertyName) || self::propertyInfo()->isInitializable($class, $propertyName);
         }, ARRAY_FILTER_USE_KEY);
 
         \ksort($defaultProperties);
--- a/src/Maker/Factory/MakeFactoryData.php
+++ b/src/Maker/Factory/MakeFactoryData.php
@@ -38,8 +38,6 @@ final class MakeFactoryData
     private array $defaultProperties = [];
     /** @var list<MakeFactoryPHPDocMethod> */
     private array $methodsInPHPDoc;
-    /** @var string[]|true */
-    private array|bool $forceProperties;
 
     public function __construct(
         private \ReflectionClass $object,
@@ -48,8 +46,7 @@ final class MakeFactoryData
         private string $staticAnalysisTool,
         private bool $persisted,
         bool $withPhpDoc,
-        /** @var string[]|true $forceProperties */
-        array|bool $forceProperties
+        private bool $forceProperties
     ) {
         $this->uses = [
             $this->getFactoryClass(),
@@ -69,7 +66,6 @@ final class MakeFactoryData
         }
 
         $this->methodsInPHPDoc = $withPhpDoc ? MakeFactoryPHPDocMethod::createAll($this) : [];
-        $this->forceProperties = $forceProperties;
     }
 
     // @phpstan-ignore-next-line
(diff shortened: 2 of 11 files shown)


KDederichs:
@nikophil What would be the preferred way to get the `always_force_properties` into the `MakeFactoryData`?

My thought would be inject it into `FactoryGenerator` (`.zenstruck_foundry.maker.factory.generator`) with the ContainerBuilder and then use that to get that into the `MakeFactoryData` via setter.

Any thoughts on this?

nikophil:
Hi! Sorry I currently don't have access to a computer...
Yeah I'd do it like this I think, and pass it to its constructor or to `getDefaultProperties()`

KDederichs:
Hmm any insights on how to actually add a testcase for that? I don't think it ever loads any of the config options in the tests.

Another good option for this I just thought of was to not connect it to that flag in the config but to a new command line argument.
I think that might be even better tbh, cause that way the user can decide on a class by class basis if they want to skip those or not.

src/Maker/Factory/MakeFactoryData.php
```diff
@@ -62,6 +65,14 @@ public function __construct(
...
 
         $this->methodsInPHPDoc = $withPhpDoc ? MakeFactoryPHPDocMethod::createAll($this) : [];
+        $reflectionExtractor = new ReflectionExtractor();
+        $this->propertyInfo = new PropertyInfoExtractor(
+            [],
+            [],
+            [],
+            [$reflectionExtractor],
+            [$reflectionExtractor]
+        );
```
nikophil: few things about this: 

- I'm not opposed to use property info component (I do prefer this solution to the `newInstanceWithoutConstructor` stuff), but we should make the dependency explicit in `composer.json` 
- I think we can only use `ReflectionExtractor`
- please, store this in a static property:
```php
private static (PropertyAccessExtractorInterface&PropertyInitializableExtractorInterface)|null $propertyInfo;

private static function propertyInfo(): PropertyAccessExtractorInterface&PropertyInitializableExtractorInterface
{
    return self::$propertyInfo ??= new ReflectionExtractor();
}
```

src/Maker/Factory/MakeFactoryData.php
```diff
@@ -154,6 +165,15 @@ public function addDefaultProperty(string $propertyName, string $defaultValue):
...
     {
         $defaultProperties = $this->defaultProperties;
+        $clazz = $this->object->getName();
+
+        foreach ($defaultProperties as $propertyName => $propertyDefault) {
+            if ($this->propertyInfo->isWritable($clazz, $propertyName) || $this->propertyInfo->isInitializable($clazz, $propertyName)) {
+                continue;
+            }
+            unset($defaultProperties[$propertyName]);
+        }
```
nikophil: please, use `array_filter`

nikophil (commented):
hi @KDederichs 

thanks for this PR :+1: 

> Another good option for this I just thought of was to not connect it to that flag in the config but to a new command line argument.
> I think that might be even better tbh, cause that way the user can decide on a class by class basis if they want to skip those or not.

I have some doubt that people will ever use such niche option, but I'm not opposed to add it :)
What do you think of doing both? If `always_force_properties` is enabled, we will never skip any properties, and if the option is added, we will not skip those properties for this factory. Using the option AND having the configuration enabled would be redundant but won't trigger any warning nor error.

> Hmm any insights on how to actually add a testcase for that? I don't think it ever loads any of the config options in the tests.

We don't have any mechanism for this yet. You can check what's done in `zenstruck/messenger-test` (see [here](https://github.com/zenstruck/messenger-test/blob/1.x/tests/Fixture/Kernel.php#L49) and [here](https://github.com/zenstruck/messenger-test/blob/1.x/tests/InteractsWithBusTest.php#L45)) about this. Or if you're not willing to do this, I can merge your PR without this test, and I'll add it myself

src/Maker/Factory/MakeFactoryData.php
KDederichs: My IDE is screaming at me that that union type is a 8.2 feature but minimum version is 8.1 :)

nikophil: and your IDE is right :)

well, actually not union type, but "[Disjunctive Normal Form Types](https://wiki.php.net/rfc/dnf_types)"

just use `ReflectionExtractor` everywhere :shrug:

KDederichs: It's actually working as return type but not as variable type apparently, I just annotated it and set it to mixed for the variable. Seems to work ok

KDederichs:
Well not quite sure what PHPStan wants from me but I think that's how the always_force_property stuff should work.
I also figured out how to test it, turn's out `.zenstruck_foundry.maker.factory.generator` doesn't seem to be a thing.

If you know how to best inject that setting into there let me know

src/Maker/Factory/MakeFactoryData.php
nikophil: Yes, the return type is not a "DNF" (the ones using union and intersection, but just simple intersection

src/Maker/Factory/FactoryGenerator.php
```diff
@@ -32,6 +32,9 @@ final class FactoryGenerator
     public const PHPSTAN_PATH = '/vendor/phpstan/phpstan/phpstan';
     public const PSALM_PATH = '/vendor/vimeo/psalm/psalm';
 
+    /** @var string[]|true */
+    private array|bool $forceProperties = [];
```
nikophil: `always_force_properties` is [a bool which can be true/false](https://github.com/zenstruck/foundry/blob/2.x/src/ZenstruckFoundryBundle.php#L81-L84). You cannot globally chose to force some specific properties.

please, add this to the constructor, using property promotion, and with `false` as default. We won't use the `methodCall`/`returnsClone` stuff her

src/Maker/Factory/FactoryGenerator.php
```diff
@@ -161,4 +165,12 @@ private function staticAnalysisTool(): string
...
         };
     }
+
+    public function alwaysForce(string ...$properties): self
+    {
+        $clone = clone $this;
+        $clone->forceProperties = $properties ?: true;
+
+        return $clone;
+    }
```
nikophil: please remove this method

src/Maker/Factory/MakeFactoryData.php
```diff
@@ -43,6 +48,8 @@ public function __construct(
         private string $staticAnalysisTool,
         private bool $persisted,
         bool $withPhpDoc,
+        /** @var string[]|true $forceProperties */
+        array|bool $forceProperties
```
nikophil:
```suggestion
        private bool $forceProperties
```

(BTW, I'm not sure, but I think that PHPStan does not understand because you should either use `@param` above the `__construct` method, or use `@var` here but without the name of the property)

src/ZenstruckFoundryBundle.php
```diff
@@ -341,6 +341,9 @@ private function configureInstantiator(array $config, ContainerBuilder $containe
             $container->getDefinition('.zenstruck_foundry.instantiator')
                 ->addMethodCall('alwaysForce', returnsClone: true)
             ;
+            $container->getDefinition('.zenstruck_foundry.maker.factory.generator')
+                ->addMethodCall('alwaysForce', returnsClone: true)
+            ;
```
nikophil: The error `You have requested a non-existent service ".zenstruck_foundry.maker.factory.generator".` happens because at this time, the service does not exist yet: see where is called `configureInstantiator()`, and where is called `$configurator->import('../config/makers.php');` few lines later.

Please, move this stuff at the end of the if: `if (isset($bundles['MakerBundle'])) {`  and you can set it like this:

```suggestion
            $container->getDefinition('.zenstruck_foundry.maker.factory.generator')
                ->setArgument('$forceProperties', $config['instantiator']['always_force_properties'])
            ;
```

I think this is enough if you add the default `false` is the constructor of `FactoryGenerator`

nikophil (changes requested):

tests/Fixture/TestKernel.php
```diff
@@ -35,6 +35,13 @@ protected function configureContainer(ContainerBuilder $c, LoaderInterface $load
     {
         parent::configureContainer($c, $loader);
 
+        $path = \sprintf('%s/config/%s.yaml', __DIR__, $this->getEnvironment());
+        if (file_exists($path)) {
+            $loader->load($path);
+        }
+
+
+
```
nikophil: please remove extra lines

tests/Fixture/TestKernel.php
```diff
@@ -35,6 +35,13 @@ protected function configureContainer(ContainerBuilder $c, LoaderInterface $load
     {
         parent::configureContainer($c, $loader);
 
+        $path = \sprintf('%s/config/%s.yaml', __DIR__, $this->getEnvironment());
+        if (file_exists($path)) {
+            $loader->load($path);
+        }
```
nikophil:
```suggestion
        if ($this->getEnvironment() !== 'test') {
            $loader->load(\sprintf('%s/config/%s.yaml', __DIR__, $this->getEnvironment()));
        }
```
this way we will throw if the file does not exist

KDederichs: the env isn't test for some reason, it's dev not sure if that's intended

nikophil: Not intended, maybe you can set it in the constructor, to be sure of what we need to exclude in the if

src/ZenstruckFoundryBundle.php
KDederichs: I can't for the life of me get this to work....

I tried setting it a million different ways but I think once that's service is initialized and loaded you can't seem to change that value anymore

nikophil: big :facepalm: sorry, this "force `test` environment in constructor" stuff was a mistake

maybe we can force it with `APP_ENV=test` in  `.env`?

nikophil: but the problem still exists... something is strange here :thinking:

nikophil: ok, I found the culprit!
you should not call `self::bootKernel();` directly, but pass the option through `makeFactoryCommandTester()` (and still, not fix the env in the constructor, this was a mistake)

src/Maker/Factory/MakeFactoryData.php
```diff
@@ -154,6 +158,16 @@ public function addDefaultProperty(string $propertyName, string $defaultValue):
     public function getDefaultProperties(): array
     {
         $defaultProperties = $this->defaultProperties;
+        $clazz = $this->object->getName();
+
+        $defaultProperties = array_filter($defaultProperties, function (string $propertyName) use ($clazz): bool {
```
nikophil: please, add a small comment to explain what/why we're doing this here

src/ZenstruckFoundryBundle.php
KDederichs: Oh yeah that does it, didn't even see that it booted the Kernel down there...

nikophil (approved):
nice! this was harder than expected!

src/Maker/Factory/MakeFactoryData.php
```diff
@@ -154,6 +158,22 @@ public function addDefaultProperty(string $propertyName, string $defaultValue):
     public function getDefaultProperties(): array
     {
         $defaultProperties = $this->defaultProperties;
+        $clazz = $this->object->getName();
```
nikophil:
```suggestion
        $class = $this->object->getName();
```

KDederichs:
Yeah many little hidden pitfalls, I did not expect this either

PR #801, opened 2025-02-02
feat: validate objects

nikophil:
Credits to @COil for the idea https://www.strangebuzz.com/en/blog/validating-your-data-fixtures-with-the-alice-symfony-bundle :smile:

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -12,7 +12,5 @@ foreach ($argv ?? [] as $i => $arg) {
         break;
     }
 }
-
 $application = new Application(new TestKernel($_ENV['APP_ENV'], true));
-
 $application->run();
--- a/config/services.php
+++ b/config/services.php
@@ -5,7 +5,9 @@ namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 use Faker;
 use Zenstruck\Foundry\Configuration;
 use Zenstruck\Foundry\FactoryRegistry;
+use Zenstruck\Foundry\Object\Event\AfterInstantiate;
 use Zenstruck\Foundry\Object\Instantiator;
+use Zenstruck\Foundry\Object\ValidationListener;
 use Zenstruck\Foundry\StoryRegistry;
 
 return static function (ContainerConfigurator $container): void {
@@ -34,7 +36,13 @@ return static function (ContainerConfigurator $container): void {
             service('.zenstruck_foundry.persistence_manager')->nullOnInvalid(),
             service('event_dispatcher'),
             '%env(default:zenstruck_foundry.faker.seed:int:FOUNDRY_FAKER_SEED)%',
+            param('.zenstruck_foundry.validation_enabled'),
+            abstract_arg('validation_available'),
         ])
         ->public()
+
+        ->set('.zenstruck_foundry.validation_listener', ValidationListener::class)
+        ->args([service('validator')])
+        ->tag('kernel.event_listener', ['event' => AfterInstantiate::class, 'method' => '__invoke'])
     ;
 };
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1199,7 +1199,7 @@ Validate your objects
 Foundry can validate your objects automatically after they are instantiated. This can be useful to
 ensure that your objects are in a valid state before they are used in your tests.
 
-You can either enable validation them globally:
+You can either enable validation globally:
 
 .. configuration-block::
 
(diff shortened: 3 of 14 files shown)


bin/console
```diff
@@ -6,5 +6,11 @@ use Zenstruck\Foundry\Tests\Fixture\TestKernel;
...
 require_once __DIR__ . '/../tests/bootstrap.php';
 
-$application = new Application(new TestKernel('test', true));
+foreach ($argv ?? [] as $i => $arg) {
+    if (($arg === '--env' || $arg === '-e') && isset($argv[$i + 1])) {
+        $_ENV['APP_ENV'] = $argv[$i + 1];
+        break;
+    }
+}
+$application = new Application(new TestKernel($_ENV['APP_ENV'], true));
```
nikophil, line 15: this is easier for debugging

config/services.php
```diff
@@ -33,7 +35,13 @@
...
             service('.zenstruck_foundry.persistence_manager')->nullOnInvalid(),
             service('event_dispatcher'),
+            param('.zenstruck_foundry.validation_enabled'),
+            abstract_arg('validation_available'),
         ])
         ->public()
+
+        ->set('.zenstruck_foundry.validation_listener', ValidationListener::class)
+        ->args([service('validator')])
+        ->tag('kernel.event_listener', ['event' => AfterInstantiate::class, 'method' => '__invoke'])
```
nikophil, line 46: I was wondering if I should listen `AfterPersist` event, and trigger validation on this event for persisting factories, but it feels wrong to validate after persisting

src/Object/ValidationListener.php
```diff
@@ -0,0 +1,30 @@
...
+    public function __invoke(AfterInstantiate $event): void
+    {
+        if (!$event->factory->validationEnabled()) {
+            return;
+        }
+
+        $violations = $this->validator->validate($event->object);
+
+        if ($violations->count() > 0) {
+            throw new ValidationFailedException($event->object, $violations);
```
nikophil, line 30: I'm wondering if we should throw this exception or another one from Foundry's namespace.  This ones is kinda ugly:
```
Symfony\Component\Validator\Exception\ValidationFailedException: Object(Zenstruck\Foundry\Tests\Fixture\Entity\EntityForValidation).name:
    This value should not be blank. (code c1051bb4-d103-4f74-8988-acbcafc7fdc3)
```

src/ZenstruckFoundryBundle.php
```diff
@@ -87,6 +89,16 @@ public function configure(DefinitionConfigurator $definition): void
                             ->example('my_instantiator')
                             ->defaultNull()
                         ->end()
+                        ->arrayNode('validation')
+                            ->info('Automatically validate the objects created.')
+                            ->canBeEnabled()
```
nikophil, line 100: it seems logical to add this config under `instantiator` conf, but this could be moved outisde from it.

tests/Fixture/Entity/EntityForValidation.php
```diff
@@ -0,0 +1,29 @@
...
+use Doctrine\ORM\Mapping as ORM;
+use Zenstruck\Foundry\Tests\Fixture\Model\Base;
+use Symfony\Component\Validator\Constraints as Assert;
+
+#[ORM\Entity]
+class EntityForValidation extends Base
+{
+    public function __construct(
+        #[ORM\Column()]
+        #[Assert\NotBlank()]
```
COil, line 25: Hello @nikophil, what about validation groups?

nikophil: hi @COil  :wave: 

yeah good call, I'll add them as a parameter of `withValidation()` method. Or maybe a new method `withValidationGroups()`

nikophil: done

docs/index.rst
```diff
@@ -1193,6 +1193,38 @@ You can even create associative arrays, with the nice DX provided by Foundry:
     // will create ['prop1' => 'foo', 'prop2' => 'default value 2']
     $array = SomeArrayFactory::createOne(['prop1' => 'foo']);
 
+Validate your objects
+~~~~~~~~~~~~~~~~~~~~~
+
+Foundry can validate your objects automatically after they are instantiated. This can be useful to
+ensure that your objects are in a valid state before they are used in your tests.
+
+You can either enable validation them globally:
```
```suggestion
You can either enable validation globally:
```

docs/index.rst
```diff
@@ -1193,6 +1193,38 @@ You can even create associative arrays, with the nice DX provided by Foundry:
...
+Or enable/disable it for a specific factory, or in a specific test with methods ``withValidation()`` / ``withoutValidation()``:
+
+::
+
+    class UserFactory extends ObjectFactory
+    {
+        protected function initialize(): static
+        {
+            return $this->withValidation();
+            // you can also disable validation using ->withoutValidation()
```
kbond: Can you show an example of using groups? I'm guessing this would be most useful when actually creating the factory in your test?

kbond (approved):

docs/index.rst
nikophil: I've updated the docs, and I've removed the example using `Factory::initialize()`, this would create an error in unit tests :sweat_smile: 

By the way, I've added a warning, mentioning you can only use validation in kernel tests

kbond (approved):

kbond:
Well, after the CI is green

PR #802, opened 2025-02-02
feat: introduce  attribute `#[AsFoundryHook]`

nikophil:
I finally found a -not so hacky- way to do this! :sweat_smile:

### DIFF
--- a/src/Object/Event/HookListenerFilter.php
+++ b/src/Object/Event/HookListenerFilter.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the zenstruck/foundry package.
+ *
+ * (c) Kevin Bond <kevinbond@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Zenstruck\Foundry\Object\Event;
+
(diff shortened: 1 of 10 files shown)


kbond (approved):
Clever, I love it!

src/Object/Event/HookListenerFilter.php
```diff
@@ -0,0 +1,45 @@
...
+final class HookListenerFilter
+{
+    /** @var \Closure(Event<object>): void */
+    private \Closure $listener;
+
+    /**
+     * @param array{0: object, 1: string} $listener
+     * @param class-string|null           $objectClass
+     */
+    public function __construct(array $listener, private ?string $objectClass = null)
```
kbond, line 25:
```suggestion
    public function __construct(callable $listener, private ?string $objectClass = null)
```

Can't do this directly - or perhaps this is better for improved exceptions message?

nikophil: I don't know how to do it: there is a class `ServiceClosureArgument` but it takes only one argument, so the class has to be callable
 
https://symfony.com/doc/current/service_container/service_closures.html#using-a-service-closure-in-a-compiler-pass

kbond: Ok no worries. I was just curious

PR #804, opened 2025-02-02
feat: introduce "reuse()" method

### LINKED ISSUE #794: `reuse()` method
I do think a `recycle()` method, similar to what [Laravel factories](https://laravel.com/docs/11.x/eloquent-factories#recycling-an-existing-model-for-relationships) provide could be useful.

While reading [some blog posts about it](https://medium.com/@arifulweb/simplifying-model-relationships-in-laravel-with-recycle-a-game-changer-for-my-lms-eff39cdd7a78), I figured out that one of its usage is basically the same thing than  something like `Contact::new(['address' => AddressFactory::createOne())->many(5)->create()`, and I'm not interested about this part of the feature.

What I like about it, is that it forces to reuse **deeply** the same object, for every `ManyToOne` relationship.

Given their example with `Airline`, `Flight`, and `Ticket` models, where the ticket belongs to an airline and a flight, and the flight also belongs to an airline, the following code will create only one `Airline`, and use it both for `Ticket::$airline` relationship and for `Flight::$airline` relationship:
```php
 Ticket::new()
    ->recycle(Airline::createOne())
    ->create();
```

As far as I understand how it works, `recycle()` only applies for `*ToOne` relationships.

It also accepts a collection of objects:
```php
$airlines = Airline::createMany(2);

 Ticket::new()
    ->recycle($airlines)
    ->many(5)
    ->create();
```
 
But I don't know if in this case, we should ensure the integrity of the data and always pick the same airline for both the ticket and its flight. I'm tempted to say "yes", because for me the whole point of this `recycle()` method would be to ensure integrity of the model. As far as I understand, this would differ from Laravel's `recycle()` method, where they always pick data randomly from the recycle collection.

any thought @kbond ?

nikophil:
fixes #794 

For now, `reuse()` only takes one parameter. It cannot be passed multiple times an object of the same class nor a Factory

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -535,6 +535,10 @@ If you have a collection of values that you want to distribute over a collection
 
         ->create();
 
+.. versionadded::  2.4
+
+    The ``distribute()`` method was added in Foundry 2.4.
+
 Faker
 ~~~~~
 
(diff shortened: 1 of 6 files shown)


docs/index.rst
```diff
@@ -947,6 +951,31 @@ The following assumes the ``Post`` entity has a many-to-many relationship with `
     // Example 5: create 3 Posts each with between 0 and 3 unique Tags
     PostFactory::createMany(3, ['tags' => TagFactory::new()->many(0, 3)]);
 
+Reuse objects in relationships
```
```suggestion
Reuse Objects in Relationships
```

kbond (approved):
Love this! Docs are super clear and really show the benefit.

PR #807, opened 2025-02-05
feat(faker): Improve reproducibility with faker

### LINKED ISSUE #793: Improve reproducibility with faker
When using Faker, I prefer to have full randomness from one execution to another (ie: not specifying a seed)

but, when a test fails, I would like to reproduce the problem, with the exact same data, which is not currently possible, because it used an unknown seed.

My proposal would be to always give a random seed to faker, and then display it right after phpunit's execution (thanks to our PHPUnit extension). And being able to set the seed thanks to an environment variable.

I think this would also mean to deprecate `zenstruck_foudry.faker.seed` configuration.

nikophil:
fixes #793

This PR does the following:
- deprecate `faker.seed` config
- load faker seed from the en var `FOUNDRY_FAKER_SEED`
- always set a random seed if not provided by the env var (and the legacy config)
- display the seed after PHPUnit has run


kbond (approved):
Cool!

Issue #808, opened 2025-02-07
[2.3.2] consecutive _real() calls return different objects


KDederichs:
Like the title says, since 2.3.2 calling _real() multiple times returns a different object every time.

This used to work in 2.3.1 but starting from 2.3.2 only the second element is contained in the collection, prompting me to  believe I get a new 'real' object every time I call it.

```php
        $boxSubscription->_real()->addApplicationPdf(base64_decode(SignatureFixture::BASE64));
        $boxSubscription->_real()->addKvForm(base64_decode(SignatureFixture::BASE64));
        $boxSubscription->_save();
```

Is that intended? I'd assume it should always return the same real instance of the entity

nikophil:
Hi @KDederichs 

> Is that intended?

absolutely not

I'm trying to reproduce your problem, but I don't find how
see https://github.com/zenstruck/foundry/pull/809

Could you elaborate, please?

KDederichs:
@nikophil I PR'd you a failing test case

nikophil:
perfect!

KDederichs:
Oh btw, it's the same code snipped that gave you this gem:
https://github.com/zenstruck/foundry/issues/710

Dunno if it's in any way related, just in case you're wondering why I use _real() there....
So basically atm I can not use anything higher than 2.3.1 ^^

nikophil:
#710 is a bug that we cannot fix for now, and I think we will never be able to fix. One of the stuff that makes me says that we should drop this whole proxy system that creates more problems than it fixes 🤷 

> I PR'd you a failing test case

I'm looking forward to it, I'm still not capable to reproduce your problem

KDederichs:
Oh I think there's a miscommunication here, when I said that I actually DID PR you the test case (to your test repo)
https://github.com/nikophil/foundry/pull/7

nikophil:
haha ok, that's how I did understand your sentence, but this never come up in my notifications 🤷 

I'll check this really soon

KDederichs:
yeah Github has been a bit inconsistent with notifications the last week or so I feel...
Thanks for taking a look though :)

nikophil:
@KDederichs  hum ok, it is a `OneToMany`! based on your example here, I was thinking it was an array of strings

Still, I cannot reproduce: https://github.com/zenstruck/foundry/pull/809/files

could you make my new test fail? (please do it in this repo, no problem to add another PR)

KDederichs:
[...]

PR #809, opened 2025-02-09
test: ensure `Proxy::_real()` always return same object

nikophil:
related to #808


nikophil:
I'm closing this. I think the fix would be to stop calling autorefesh in Proxy::_real()

PR #813, opened 2025-02-10
Test/real test

KDederichs:


KDederichs:
There you go @nikophil, it's somehow related to persisting things in collections via `cascade: ['persist']` yet again, sorry I always seem to run into those strange ones

nikophil:
ho yeah, I forgot to try with different cascade config

nikophil:
I'm closing this. I think the fix would be to stop calling autorefesh in  `Proxy::_real()`

PR #815, opened 2025-02-13
fix: can index one to many relationships based on "indexBy"

### LINKED ISSUE #812: [2.3.2] Creating an OneToMany in an OneToMany layer leads to collection using indexby missing index
After upgrading zenstruck/foundry to 2.3.2 in our project, some tests failed that we successful before.
Test which are using Entities which rely on OneToMany DoctrineCollections with the "indexBy" parameter were affected

With an Collection inside an Entity
```php
    #[ORM\OneToMany(targetEntity: Translation::class, mappedBy: 'category', cascade: ['persist'], indexBy: 'culture')]
    private Collection $translations;
```

the getTranslation call will return null with 2.3.2, but return the expected translation with 2.3.1

```php
    public function getTranslation(string $culture = 'en'): ?Translation
    {
        return $this->translations->get($culture);
    }
```

This seems to only occur on an Entity / Factory Setup where there is an OneToMany relationship across at least 2 entities

```php
public function testIndexedByWithTwoOneToManyLayers(): void
    {
        $space = SpaceFactory::new()->with(
            [
                'categories' => CategoryFactory::new()->with(
                    [
                        'translations' => TranslationFactory::new(['culture' => 'en'])->many(1),
                    ])->many(1),
            ]
        )->create()
        ;
        self::assertNotNull($space->getCategories()[0]->getTranslation('en'));
    }
```

I setup an reproducer project with the failing TestCase `testIndexedByWithTwoOneToManyLayers` https://github.com/tandev/foundry-index-by-doctrine-collections-reproducer/blob/main/tests/DoctrineCollectionIndexByTest.php

It seems something might have changed in regards of loading collections on such setups? Is there something to be changed in the usage of Factories on our side or is this a bug?

nikophil:
fixes #812

ping @tandev

needs to be merged in 2.x afterwards

### DIFF
--- a/src/ORM/OrmV2PersistenceStrategy.php
+++ b/src/ORM/OrmV2PersistenceStrategy.php
@@ -66,6 +66,7 @@ final class OrmV2PersistenceStrategy extends AbstractORMPersistenceStrategy
         return new InverseRelationshipMetadata(
             inverseField: $association['fieldName'],
             isCollection: $inversedAssociationMetadata->isCollectionValuedAssociation($inversedAssociation['fieldName']),
+            collectionIndexedBy: $inversedAssociation['indexBy'] ?? null
         );
     }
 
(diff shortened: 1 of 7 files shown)


kbond (approved):
TIL about `indexBy` 😅

PR #818, opened 2025-02-16
minor: deprecate auto-persist configuration

nikophil:


kbond (approved):
To clarify, the way to achieve is to use `->withoutPersisting()` on your factory?

nikophil:
@kbond yep, there are multiple ways to do this:

- indeed, use `->withoutPersisting()` on your factory
- use the helper `disable_persisting()` which disables it globally, for the current test

nikophil:
by the way, this was not tested, nor documented and I've never seen it used in the wild. It is also a small annoyance, because before accessing `PersistentObjectFactory::$persistMode` you have to check if it is actually initialized.

For all these reasons, I think the best is to deprecate it. But I'm fine to keep it if you think we should :relaxed:

kbond:
Let's deprecate and see if we hear anything

PR #822, opened 2025-02-17
fix: trigger after persist callbacks for entities scheduled for insert

### LINKED ISSUE #821: [2.3.2] afterPersist not called for FactoryCollection for collection property with inversed relationship
Since the update, it seems that afterPersist calls for an collection property with an inversed relationship are not called anymore

This can be reproduced by adding the following test cases to: `\Zenstruck\Foundry\Tests\Integration\ORM\EntityRelationship\ProxyEntityFactoryRelationshipTest`

```php
    /** @test */
    public function it_uses_after_persist_with_collection(): void
    {
        $contact = static::contactFactory()->with(
            [
                'tags' => static::tagFactory()
                    ->afterPersist(function (Tag $tag) {$tag->setName('foobar');})
                    ->many(1)])
            ->create()
        ;

        self::assertEquals('foobar', $contact->getTags()->first()->getName());
    }

    /** @test */
    public function it_uses_after_persist_with_collection_with_inversed_relationship(): void
    {
        $contact = static::categoryFactory()
            ->with(
                [
                    'contacts' => static::contactFactory()
                        ->afterPersist(function (Contact $contact) {$contact->setName('foobar');})
                        ->many(1)])
            ->create()
        ;

        self::assertEquals('foobar', $contact->getContacts()->first()->getName());
    }
```

```
There was 1 failure:

1) Zenstruck\Foundry\Tests\Integration\ORM\EntityRelationship\ProxyEntityFactoryRelationshipTest::it_uses_after_persist_with_collection_with_inversed_relationship
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'foobar'
+'harum'

```

nikophil:
Because we're now "scheduling for insert" some entities, and not directly saving them, we should take care of their after persist callbacks differently. The proposal of this PR is to give this responsibility to `PersistenceManager` 

fixes #821

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -25,13 +25,6 @@ use Faker;
  */
 abstract class Factory
 {
-    /**
-     * Memoization of normalized parameters.
-     *
-     * @internal
-     * @var Parameters|null
-     */
-    protected ?array $normalizedParameters = null;
     /** @phpstan-var Attributes[] */
     private array $attributes;
 
@@ -208,7 +201,7 @@ abstract class Factory
      */
     protected function normalizeParameters(array $parameters): array
     {
-        return $this->normalizedParameters = \array_combine(
+        return \array_combine(
             \array_keys($parameters),
             \array_map($this->normalizeParameter(...), \array_keys($parameters), $parameters)
         );
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -38,7 +38,6 @@ parameters:
         - identifier: property.readOnlyByPhpDocDefaultValue
           paths:
             - src/Object/Hydrator.php
-            - src/Factory.php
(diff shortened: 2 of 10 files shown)


src/Factory.php
```diff
@@ -25,13 +25,6 @@
  */
 abstract class Factory
 {
-    /**
-     * Memoization of normalized parameters.
-     *
-     * @internal
-     * @var Parameters|null
-     */
-    protected ?array $normalizedParameters = null;
```
nikophil, line 34: this is not needed anymore, because we're scheduling entities for insert from an "after instantiate" callback, so we can pass the attributes from there

kbond (approved):

PR #823, opened 2025-02-18
docs: Make sure we add links on separate lines

Nyholm:
This is just a minor. I did not understand it was two links at first.


nikophil:
thanks @Nyholm

PR #825, opened 2025-02-19
doc: fixes linking to object-proxy

mvhirsch:


nikophil (approved):

docs/index.rst
```diff
@@ -1733,7 +1733,7 @@ you're using them in tests. Thanks to it, you can:
 
 .. warning::
 
-    Because Foundry is relying on its `Proxy mechanism <object-proxy>`_, when using persistence,
+    Because Foundry is relying on its `Proxy mechanism <https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#object-proxy>`_, when using persistence,
```
nikophil: I think, in `.rst` an internal link is written this way:
```suggestion
    Because Foundry is relying on its :ref:`Proxy mechanism <object-proxy>`, when using persistence,
```

nikophil (changes requested):
hi @mvhirsch 
thanks for this. Actually, a change is needed please

mvhirsch:
I'm not familiar with rst, but well, let's fix all of them, right @nikophil ? E.g.:

https://github.com/zenstruck/foundry/blob/fd44762755347afa76afc94c21abf00e552dda4b/docs/index.rst#L8

nikophil:
good idea, thanks!

mvhirsch:
Here we go, @nikophil :-)

nikophil:
thanks @mvhirsch 

I'm wondering if the `:ref:` will work inside a code block, I'll fix it if it does not

nikophil (approved):

PR #834, opened 2025-02-26
feat: introduce `mapEach()` method

nikophil:
I'm bringing back this `mapEach()` method. 

On the project I'm working on, we have tons of factories with "withers" which prevents to use sequences and fluent usage of the factories when collections are involved.

I'm opened to any better api, or naming :sweat_smile:

### DIFF
--- a/src/FactoryCollection.php
+++ b/src/FactoryCollection.php
@@ -187,6 +187,29 @@ final class FactoryCollection implements \IteratorAggregate
         );
     }
 
+    /**
+     * @param list<mixed>|list<list<mixed>> $values
+     *
+     * @return self<T, TFactory>
+     */
+    public function mapEach(callable $callback, array $values): self // @phpstan-ignore missingType.callable (cannot properly type the callable)
+    {
+        $factories = $this->all();
+
+        if (count($factories) !== count($values)) {
+            throw new \InvalidArgumentException('Number of values must match number of factories.');
+        }
+
+        return new self(
+            $this->factory,
+            static fn() => \array_map(
+                static fn(Factory $f, mixed $value) => is_array($value) ? $callback($f, ...$value) : $callback($f, $value),
+                $factories,
+                $values
+            )
+        );
+    }
+
     public function getIterator(): \Traversable
     {
         return new \ArrayIterator($this->all());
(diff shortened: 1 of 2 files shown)


smnandre:
`withEach` ?  Or maybe even `eachWith` ?

kbond (commented):
Would you mind adding the docs for this with a *real* example? It'd help me understand the feature more I think.

aegypius:
> `withEach` ? Or maybe even `eachWith` ?

I agree `mapEach` seem to be a little redundant why not a simple `map(callable callback): self` like our beloved [PHP\Ds Sequence::map(...)](https://www.php.net/manual/fr/ds-sequence.map.php)

src/FactoryCollection.php
```diff
@@ -187,6 +187,29 @@ public function distribute(string $field, array $values): self
         );
     }
 
+    /**
+     * @param list<mixed>|list<list<mixed>> $values
+     *
+     * @return self<T, TFactory>
+     */
+    public function mapEach(callable $callback, array $values): self // @phpstan-ignore missingType.callable (cannot properly type the callable)
```
silasjoisten, line 195:
```suggestion
    public function map(callable $callback, array $values): self // @phpstan-ignore missingType.callable (cannot properly type the callable)
```

smnandre:
Maybe then it would be better to use the same signature as `array_map` (the variadic values) ? 

```php
*
 * @return self<T, TFactory>
 */
function map(callable $callback, array... $arguments): self
```

Instead of calling  `map($callable, [[1, 2, 3], [a, b, c]])` , one could do `map(callable, [1,2,3], [4,5,6]) `

nikophil:
yeah this could be a good idea... I still need to work on this one and think of a good real life example to convince @kbond :laughing: 

I'm not really sure `map()` is a good name, the main difference is that this method effectively applies a callback on each element of a `FactoryCollection`, but it will actually pass a different item as a parameter of the callback on each iteration.

It's a mix between a "map()" and the new [`distribute()`](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#distribute-values-over-a-collection) method recently introduced.

BTW, I'm still not 100% convince of the API of this method. But I really would like something to call so-called "state method" on each element of a `FactoryCollection`

Issue #835, opened 2025-02-28
Creating entities with One2One and One2Many relationships fails if one2one is before one2many


mmarton:
Good day @nikophil !

You've helped me before with my edge cases about one2one relationships in the past (#655), and I'm here again with a special case.  :)

There is this special case that happens in the PersistentObjectFactory with one2one relations:
https://github.com/zenstruck/foundry/blob/719710a76d4fcfebf26b124989ae846502ee406e/src/Persistence/PersistentObjectFactory.php#L310

If my entity has a one2one and a many2one this "placeholder entity" may be flushed by doctrine.
This happens when the one2one relation is defined before the one2many, making the order of keys in the default function matter. 

I've created a small reproducer to this issue:
https://github.com/mmarton/foundry-entity-relations-issue

best to you

nikophil:
hey @mmarton 

cool thanks for the reproducer, I'll have a look soon

did this happen in a recent version?

mmarton:
yes, everything is on the latest version.
- foundry: 2.3.6
- phpunit: 12.0.5
- symfony: 7.2.4
- doctrine orm: 3.3.2

edit: I misunderstood the question. The issue was introduced with the 2.3.0 version.

mmarton:
An extra addition to the issue:

If I move my one2one relation as the last entry of the array and call 

ProductFactory::new()->withOptionalEntity['optionalentity' => OptionalEntityFactory::new()])->create()

 to add a new optional relation that is not present in the default() array, then the same issue happens.

nikophil:
hey @mmarton 

out of curiosity, do you use php 8.4?

For some reasons, I'm struggling to fix this in every possible combinations of "cascade persist" relationships. I think the only real complete fix will use php 8.4 proxy objects

mmarton:
Interesting. Yes the project is already on 8.4, like the reproducer I've created

nikophil:
just found a very nice workaround :)

https://github.com/zenstruck/foundry/pull/836

I'd be glad if you try this before we merge. I've already tested it in your reproducer, and everything works.

mmarton:
I've tested your fix, and interestingly it fixes my reproducer, but fails in my project with the same thing: it tries to save an incomplete entity. I have no idea why.

nikophil:
😬

nikophil:
I don't manage to reproduce it within our testsuite without a reproducer 😞 

[...]

PR #836, opened 2025-03-02
fix: flush only once per calls to `Factory::create()` in userland

nikophil:
Ok, I think it's time for this now :sweat_smile: 

It's been a while that I wanted to achieve this: we're calling too many `flush()`, because we're flushing every time an entity is created, although, the "Doctrine way" would be to only flush once per "root factory".

This is actually possible thanks to all the work that have been done around "[schedule for insert](https://github.com/zenstruck/foundry/pull/742)" stuff. 

A problem that could have happened because of this, is with "after persist" callbacks : not calling `save()` for every entity would mean not calling their "after persist callbacks", but I was pleasantly surprised that this is an already solved problem, thanks to [this PR](https://github.com/zenstruck/foundry/pull/822)  :tada: 

What triggered me to do this now, was [this issue](https://github.com/zenstruck/foundry/issues/835) raised by @mmarton

given this code:
```php
ProductFactory::createOne([
    'stock' => StockFactory::new(), // this is an "inverse OneToOne"
    'category' => CategoryFactory::new(), // this is a regular "ManyToOne"
])
```
there is a problem with the "[placeholder solution](https://github.com/zenstruck/foundry/blob/2.x/src/Persistence/PersistentObjectFactory.php#L310)" for inverse `OneToOne` : at some point, because we flush right after the `category` is created, and because `stock` is already persisted (the doctrine way), it tries to save the "placeholder" :shrug: 

I've tried a lot of different things, but nothing works with every combination possible of "cascade persist" stuff. But the problem is naturally fixed when we deffer the `flush()` call: we can safely remove the placeholder from the UoW, before any flush attempt is made.

Another solution would be to use PHP 8.4 proxies, because the placeholder could be the real object, but that's more work, and it will only fix the problem for the last PHP version. Moreover both of these solutions are not mutually exclusive, and I'm planning to work on this at some point,(maybe in the same time when I'll use PHP 8.4 proxies for data provider as well).

It comes also with a performance boost: 
locally, without Mongo and with DAMA enabled, the command `phpunit --exclude-group maker` runs in 10s now and in 15s before :runner: :zap: 
FYI, before the fix, the whole testsuite was performing 3040 calls to `flush()`, and now, only 1300

Surprisingly, there is no performance boost in the testsuite without dama. 

Finally, I think this should be released in `2.3.x`, because the 2.4 release will be full of new features, which are not 100% ready.

:warning: EDIT:

I've discovered yet another problem with the "placeholder" stuff for inversed one-to-one, when not using auto-generated ids: in case of `cascade: persist`, the placeholder was passed to `$em->persist()`, and it was emitting a `EntityMissingAssignedId` because the id is generated in the constructor of the entity, but the "placeholder" is created with `newInstanceWithoutConstructor()`.

Then I decided to even deffer `$em->persist()` calls. This has the main advantage that we can do all the weird stuff we want, and only after that, doctrine joins the party. With the solution, I think we will finally be done with all that "inverse one to one" madness :tada: :sweat_smile:


kbond (approved):
This is great!

nikophil:
After a lot of digging, I finally got something working on my different projects, I added a lot of test cases to prevent a lot of edge cases I encountered along the way :sweat_smile: 

@mmarton @tandev @aegypius I'd really appreciate if you could test this branch and report any bug, before I realease this PR which changed quite a lot of things :sweat_smile: :pray:

src/ORM/OrmV2PersistenceStrategy.php
nikophil, line 1: I just cleaned a little bit how we get "inversedRelationshipMetadata", hence these changes - I needed to have more information (we did only get "one to many" and "inversed one to one" but I now needed also "normal one to one"

src/Persistence/PersistentObjectFactory.php
```diff
@@ -364,26 +376,43 @@ protected function normalizeCollection(string $field, FactoryCollection $collect
...
+        $inverseRelationship = $persistenceManager->inverseRelationshipMetadata(static::class(), $object::class, $field);
+
+        if ($inverseRelationship instanceof OneToOneRelationship) {
+            $this->tempAfterInstantiate[] = static function(object $newObject) use ($object, $inverseRelationship) {
+                try {
+                    set($object, $inverseRelationship->inverseField(), $newObject);
+                } catch (\Throwable) {
+                }
+            };
+        }
```
nikophil, line 404:
```php
$address = static::addressFactory()
            ->afterInstantiate(
                static function(Address $a): void {
                    static::contactFactory()->create(['address' => $a]);
                }
            )->create();

self::assertNotNull($address->getContact());
```

it was a fluke that this kind of code did actually work. Not really sure why, maybe thanks to some doctrine lazy loading behavior. Now we're only flushing once here, we need to hydrate manually the inverse field

nikophil:
@kbond I'm closing this PR, to the benefit of this one: https://github.com/zenstruck/foundry/pull/855

I'll soon try towork again on "flush once", after 2.4 is released

PR #854, opened 2025-03-27
feat: add `force()` helper

nikophil:
Here is a small feature I'll need for the future fix for inverse one to one I told you about

I think it is also a nice small addition to the Foundry's API, from time to time I had the need for this

### DIFF
--- a/src/ForceValue.php
+++ b/src/ForceValue.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the zenstruck/foundry package.
+ *
+ * (c) Kevin Bond <kevinbond@gmail.com>
+ *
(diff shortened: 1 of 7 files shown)


src/ForceValue.php
```diff
@@ -0,0 +1,40 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Zenstruck\Foundry;
+
+/**
+ * @author NIcolas PHILIPPE <nikophil@gmail.com>
+ *
+ * @internal
```
nikophil, line 17: I think it is best to set it `@internal`, so that, the only possibility to use the feature would be to use the helper

kbond (approved):

PR #855, opened 2025-03-28
fix: handle "inverse one to one" without "placeholder" solution

nikophil:
This PR will hopefully fix every problem with "inverse one to one", in a simpler way than the "placeholder" solution which introduced a lot of glitches.

now, the inverse one to one is resolved this way:
```
- is the field nullable in the current object ?
    -> add `null` to the current field, and defer the creation of the other object after the instantiation

- otherwise, is the inverse field nullable ?
    -> create the inverse object now, but set null the inverse relationship

- otherwise, throw an exception: a bi-directional one-to-one relationship with both objects nullable introduces circular dependency
```

fixes https://github.com/zenstruck/foundry/pull/835


src/Persistence/PersistentObjectFactory.php
```diff
@@ -290,30 +291,40 @@ protected function normalizeParameter(string $field, mixed $value): mixed
...
+            if ($relationshipMetadata && !$relationshipMetadata->isCollection) {
+                $inverseField = $relationshipMetadata->inverseField;
 
-                $inversedObject = $value->withPersistMode(
+                $value = $value->withPersistMode(
                     $this->isPersisting() ? PersistMode::NO_PERSIST_BUT_SCHEDULE_FOR_INSERT : PersistMode::WITHOUT_PERSISTING
-                )
+                );
+
+                if ((new \ReflectionClass(static::class()))->getProperty($field)->getType()?->allowsNull()) {
```
nikophil, line 304: I'm wondering if checking the "nullability" based on type system only is OK: it would not work with relationships which are typed with phpdoc only...

I think it's OK like this, I don't see any valid case where a one-to-one should not be strongly typed :thinking: 

Maybe we could use `symfony/typeinfo` for this?

I'm also wondering if we should reflect this in the exception message? "if the properties of the relations are not strongly types, please do..."

src/Persistence/PersistentObjectFactory.php
```diff
@@ -290,30 +291,42 @@ protected function normalizeParameter(string $field, mixed $value): mixed
...
-                    set($inversedObject, $inverseField, $object);
-                };
+                    $this->tempAfterInstantiate[] = static function(object $object) use ($inverseObject, $inverseField) {
+                        set($inverseObject, $inverseField, $object);
+                    };
 
-                return $inversedObject;
+                    return $inverseObject;
+                } elseif ($fieldType === null || $inverseFieldType === null) {
+                    throw new \InvalidArgumentException(sprintf("Cannot handle inverse OneToOne relationship: cannot determine types of \"%s::\${$field}\" and \"%s::\${$inverseField}\", please and type to the properties.", static::class(), $value::class()));
```
nikophil, line 326: I think this is enough for now: we could actually use symfony/type-info or symfony/property-info components to read phpdoc / doctrine mapping, but I can't see a real reason to not strongly type the properties of the relationship

tests/Integration/ORM/EntityRelationship/EntityFactoryRelationshipTestCase.php
```diff
@@ -519,6 +522,137 @@ public function can_call_create_in_after_persist_callback(): void
...
+                    }),
+            ]
+        );
+
+        self::assertSame('city from after persist', $contact->getAddress()->getCity());
+    }
+
+    /** @test */
+    #[Test]
+    #[DataProvider('provideCascadeRelationshipsCombinations')]
```
kbond, line 545: Don't we need an `@dataProvider` annotation also?

kbond (approved):
👍🏻

tests/Integration/ORM/EntityRelationship/EntityFactoryRelationshipTestCase.php
nikophil: nope, the whole mechanism used to provide combinations for cascade relations does not support phpunit 9

PR #852, opened 2025-03-25
Add support for persistence 4

Chris53897:
for 2.3.x branch

PR #856, opened 2025-03-28
minor: allow `doctrine/persistence` 4.x

kbond:
Same as #852 but for 1.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "require": {
         "php": ">=8.0",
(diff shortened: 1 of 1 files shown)


kbond:
@nikophil Can you merge this and release a new 1.x _after_ you create the next 2.3.x release?

If 1.x is released first, `composer require foundry` could install 1.x (if the app is already locked at `doctrine/persistence` 4+).

nikophil:
released in 1.38.5

PR #865, opened 2025-04-03
fix: performance problem with reuse

nikophil:
now we're checking all the attributes of the current class only if the factory has some "reuse" attributes

second performance improvement is to not use property info for this. I'd rather use `symfony/type-info`, but it is not  available for php  8.1 :shrug:


kbond (approved):
Failing CI job is weird, it shouldn't have run at all...

PR #869, opened 2025-04-11
minor: use `mt_rand` instead of `random_int`

kbond:
Per Slack discussion with @nikophil.

Ref: https://github.com/zenstruck/.github/commit/4bd01625145da5494ab4bf145f64013658d09712

### DIFF
--- a/src/FactoryCollection.php
+++ b/src/FactoryCollection.php
@@ -111,7 +111,7 @@ final class FactoryCollection implements \IteratorAggregate
             throw new \InvalidArgumentException('Min must be less than max.');
         }
 
-        return new self($factory, static fn() => \array_fill(0, \random_int($min, $max), []));
+        return new self($factory, static fn() => \array_fill(0, \mt_rand($min, $max), []));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


kbond:
@nikophil, there are a few other places in the code base we use `random_int` - should these be switched also?

nikophil:
yeah, I think we can replace all of them


I recently learnt that `random_int()` should be chosen over `mt_rand()` for security concerns. But in our case, we don't care about security, so we can blindly replace to `mt_rand()`

Maybe [this one](https://github.com/zenstruck/foundry/blob/2.x/src/Configuration.php#L69) should not be replaced, because at this time the seed is still not fixed.

kbond:
> Maybe [this one](https://github.com/zenstruck/foundry/blob/2.x/src/Configuration.php#L69) should not be replaced, because at this time the seed is still not fixed.

I thought the same thing, I've changed the others.

nikophil:
you should rebase in order to have the benchmark pass :sweat_smile:

nikophil (approved):

kbond:
you beat me to it :)

nikophil:
I think I will allow +/- 10% in the bechmark :sweat_smile: the results are too versatile to use only +/- 5%

(this one is failing because `CategoryFactoryBench` for bench test `bench_create_many` with `createMany(10)` is 5% slower in this PR than in 2.x, although none of them uses any random.

I think an assertion with 10% is coherent with the option `--retry-threshold=10`

I'm merging your PR, I'll fix this in another one

PR #871, opened 2025-04-13
perf: revert validation / `#[AsFoundryHook]` / global event system

nikophil:
Because of huge performance problems, I roll back those PRs:
- https://github.com/zenstruck/foundry/pull/801
- https://github.com/zenstruck/foundry/pull/802
- https://github.com/zenstruck/foundry/pull/790

Maybe we'll reintroduce it in the future if we find a way that do not create performance problems


kbond (approved):
😞

smnandre:
Does it happen when debug=0 ? 

I had problems previously with traceable event dispatcher and the profiler

kbond:
> Does it happen when debug=0 ?
I thought the same thing - @nikophil, did you ever test this?

nikophil:
I did try that yes, and but it wasn't enough

smnandre:
We had massive problems with events & TwgComponent (and still have some problems with Live) ..  i'll check my notes to see if some of the causes apply here.

PR #877, opened 2025-04-15
Fix: missing parameter when using custom `faker` service

### LINKED ISSUE #876: Breaking Change after updating from `2.3.10` to `2.4.0`
After updating the package to 2.4.0 i encounter following error:

```bash
$> symfony composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Nothing to install, update or remove
Generating autoload files
149 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
infection/extension-installer: No extensions found
phpstan/extension-installer: Extensions installed

Run composer recipes at any time to see the status of your Symfony recipes.

Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!
!!  In EnvVarProcessor.php line 128:
!!
!!    Invalid env fallback in "default:zenstruck_foundry.faker.seed:int:FOUNDRY_F
!!    AKER_SEED": parameter "zenstruck_foundry.faker.seed" not found.
!!
!!
!!
Script @auto-scripts was called via post-install-cmd
```

silasjoisten:
closes #876 

I think it does not hurt to always have the seed parameter set.

### DIFF
--- a/src/ZenstruckFoundryBundle.php
+++ b/src/ZenstruckFoundryBundle.php
@@ -348,6 +348,8 @@ final class ZenstruckFoundryBundle extends AbstractBundle implements CompilerPas
      */
     private function configureFaker(array $config, ContainerBuilder $container): void
     {
+        $container->setParameter('zenstruck_foundry.faker.seed', $config['seed']);
+
         if ($config['service']) {
             $container->setAlias('.zenstruck_foundry.faker', $config['service']);
 
@@ -359,7 +361,5 @@ final class ZenstruckFoundryBundle extends AbstractBundle implements CompilerPas
         if ($config['locale']) {
             $definition->addArgument($config['locale']);
         }
-
-        $container->setParameter('zenstruck_foundry.faker.seed', $config['seed']);
     }
 }


nikophil (approved):

nikophil:
hi @silasjoisten 

thanks for this fix, I'll release it as soon as the CI passes :)

nikophil:
benchmark fails, but that's OK, it's a really new CI check that we need to fine-tune

nikophil:
By the way, we don't test the case when `zenstruck_foundry.faker.service` is provided, we need to do this at some point

nikophil:
released ;)

https://github.com/zenstruck/foundry/releases/tag/v2.4.1

silasjoisten:
> By the way, we don't test the case when `zenstruck_foundry.faker.service` is provided, we need to do this at some point

let me try to add a test here :)

silasjoisten:
> released ;)

Thank you ❤️

nikophil:
> let me try to add a test here :)

yeah be my guest :)

we have some other tests that use different configuration of the bundle, see for example how is done `FakerSeedSetFromEnvVarKernelTest::faker_seed_can_be_set_by_environment_variable()`

we're passing `self::bootKernel(['environment' => 'faker_seed_env_var']);` and you have a file `tests/Fixture/config/faker_seed_env_var.yaml`   where you can add the custom config

thanks!

PR #878, opened 2025-04-15
chore(tests): Adds `ZenstruckFoundryBundleTest`

silasjoisten:
This will test the default configuration values for now. and offers a bridge to test symfony configurations.

### DIFF
--- a/tests/Unit/ZenstruckFoundryBundleTest.php
+++ b/tests/Unit/ZenstruckFoundryBundleTest.php
@@ -14,9 +14,8 @@ declare(strict_types=1);
 namespace Zenstruck\Foundry\Tests\Unit;
 
 use PHPUnit\Framework\Attributes\Test;
-use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\IgnoreDeprecations;
 use PHPUnit\Framework\TestCase;
-use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\Configurator\DefinitionConfigurator;
 use Symfony\Component\Config\Definition\Loader\DefinitionFileLoader;
@@ -38,8 +37,6 @@ use Zenstruck\Foundry\ZenstruckFoundryBundle;
  */
 final class ZenstruckFoundryBundleTest extends TestCase
 {
-    use ExpectDeprecationTrait;
-
     private ZenstruckFoundryBundle $bundle;
     private ContainerBuilder $container;
     private ContainerConfigurator $configurator;
@@ -85,11 +82,9 @@ final class ZenstruckFoundryBundleTest extends TestCase
      * @group legacy
      */
     #[Test]
-    #[Group('legacy')]
+    #[IgnoreDeprecations]
     public function faker_seed_value_overridden(): void
     {
-        self::expectDeprecation('Since zenstruck/foundry 2.4: The "faker.seed" configuration is deprecated and will be removed in 3.0. Use environment variable "FOUNDRY_FAKER_SEED" instead.');
-
         $config = self::buildConfiguration([['faker' => ['seed' => $expected = 1234]]]);
 
         $this->bundle->loadExtension($config,  $this->configurator, $this->container);
--- a/tests/Unit/ZenstruckFoundryBundleTest.php
+++ b/tests/Unit/ZenstruckFoundryBundleTest.php
@@ -32,6 +32,7 @@ use Zenstruck\Foundry\ZenstruckFoundryBundle;
 
 final class ZenstruckFoundryBundleTest extends TestCase
 {
+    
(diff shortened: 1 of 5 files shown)


tests/Unit/ZenstruckFoundryBundleTest.php
```diff
@@ -0,0 +1,85 @@
...
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Config\Definition\Builder\TreeBuilder;
+use Symfony\Component\Config\Definition\Configurator\DefinitionConfigurator;
+use Symfony\Component\Config\Definition\Loader\DefinitionFileLoader;
+use Symfony\Component\Config\Definition\Processor;
+use Symfony\Component\Config\FileLocator;
+use Zenstruck\Foundry\ORM\ResetDatabase\ResetDatabaseMode;
+use Zenstruck\Foundry\ZenstruckFoundryBundle;
+
+final class ZenstruckFoundryBundleTest extends TestCase
```
nikophil, line 38: I'm not against to at this test to our test suite, but I'm not sure it could assert the bug you fixed won't happen again :thinking: It does not use a container, nor use the loads the extension

I think we should do an integration test for this, see my comment here https://github.com/zenstruck/foundry/pull/877#issuecomment-2804119865

silasjoisten: I know that will come in a different PR :) i am just getting started 😄

nikophil: ok then! :ok_hand:

nikophil: BTW, don't hesitate to do all those tests in the same PR :)

silasjoisten: Ahhh okay sure!

nikophil:
this PR should now target 2.5.x please :)

silasjoisten:
@nikophil i added some more tests ;)

nikophil:
hey @silasjoisten 

thanks for this! I don't know why your last commit did not trigger the CI, but there are still some failures in the tests.

nikophil:
in the case where you are using the deprecated configuration `faker.seed`, you'd need to add `@group legacy` (for phpunit 9 + symfony/phppunit-bridge) and `#[IgnoreDeprecations]` (for phpunit >=10)

nikophil:
(actually, it is expected to break on deprecation with phpunit > 10 [but it does not](https://github.com/zenstruck/foundry/actions/runs/14509743551/job/40705647207?pr=878#step:7:142), I'll fix that soon)

silasjoisten:
I actually wanted to expect that this deprecation is triggered but i don't know how, since the native phpunit methods such as `expectDeprecation` are deprecated. That was my goal for the previously failing test case

nikophil:
You can check how it is done here: https://github.com/zenstruck/foundry/blob/2.x/tests/Integration/ForceFactoriesTraitUsage/KernelTestCaseWithoutFactoriesTraitTestCase.php#L94

But I would advise that you should not waste your time doing this :grin: 

Because, this method only exist in phpunit > 10 I guess

nikophil:
any chance you introduce an integration test, the way I described [here](https://github.com/zenstruck/foundry/pull/877#issuecomment-2804119865)?

I can also do it, no problem

silasjoisten:
> Because, this method only exist in phpunit > 10 I guess

Yea initially then i figured.

And initially i have chosen this unit test because they are considered the cheap tests.

silasjoisten:
@nikophil i added the test but i cant run this locally somehow any tip?

nikophil:
there are some explanations in the README about how to run the tests locally:

the best is to use docker to have the databases availble.

then, you can run the tests by using `./phpunit`

in order that the tests run uber-fast, you can create a `.env.local` file as following:
```env
MONGO_URL="" # if mongo is disabled, tests run faster
USE_DAMA_DOCTRINE_TEST_BUNDLE="1" # dama makes your tests faster as well
PHPUNIT_VERSION="11"
```
and then you can exclude the "maker" tests, since you're not working on it, and they are quite slow:
```terminal
./phpunit  --exclude-group maker
```

silasjoisten:
@nikophil i cant make the integration test work. The Service i want to reference does not exist in the container that makes the test fail. I reverted the integration test addition from my side.

nikophil:
hey @silasjoisten thanks for your work! It's super nice to have new contributors :blush: 

I took the liberty of fixing this myself, see [the last commit](https://github.com/zenstruck/foundry/commit/a478c859326fb2c8baa01f52e611dfdb8998c0f6)

declaring the service in the configuration was the missing part:
```yaml
services:
    Zenstruck\Foundry\Tests\Fixture\ExtendedGenerator: ~
```

silasjoisten:
@nikophil thanks! damn it was that simple? 😄

nikophil:
yeah, I really like how we can provide easily configuration for the container and the kernel this way!

PR #879, opened 2025-04-17
Prevent random value collisions when kernel is rebooted

HypeMC:
PR #807 introduced an interesting issue: when a fixed seed is used, `mt_srand()` generates the same sequence of random values every time it's called (see https://3v4l.org/AKjIO). As a result, if the kernel is shut down and rebooted, the random number generator is "reset".

This causes problems when using Foundry with Symfony's `KernelBrowser`. The kernel is shut down after the second request ([code](https://github.com/symfony/symfony/blob/01c26aa928ecc5dc42ce90268299c190c6b712ed/src/Symfony/Bundle/FrameworkBundle/KernelBrowser.php#L141-L157)), so any calls to Faker after that will return the same values as before.

Here's a simplified example:

```php
#[ORM\Entity]
class Test
{
    // ...

    #[ORM\Column(unique: true)]
    private ?int $example = null;

    // ...
}

final class TestFactory extends PersistentProxyObjectFactory
{
    {
        return Test::class;
    }

    protected function defaults(): array|callable
    {
        return [
            'example' => self::faker()->randomNumber(),
        ];
    }
}

class MyTest extends WebTestCase
{
    use Factories;
    use ResetDatabase;

    public function testSomething(): void
    {
        static::createClient()->request('GET', '/');

        TestFactory::createOne();

        static::getClient()->request('GET', '/');

        TestFactory::createOne(); // ← Fails here
    }
}
```

This currently results in a `UniqueConstraintViolationException`:

```
Doctrine\DBAL\Exception\UniqueConstraintViolationException: An exception occurred while executing a query: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "uniq_d87f7e0c6eec9b9f"
DETAIL:  Key (example)=(52197) already exists.
```

This PR ensures the seed isn't reapplied in the middle of a test. Since `Configuration::$instance` is set at the beginning of each test and reset afterward, we assume that if it's set, a test is in progress and we skip seeding.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -55,7 +55,9 @@ final class Configuration
         private readonly ?PersistenceManager $persistence = null,
         ?int $forcedFakerSeed = null,
     ) {
-        $this->faker->seed(self::fakerSeed($forcedFakerSeed));
+        if (null === self::$instance) {
+            $this->faker->seed(self::fakerSeed($forcedFakerSeed));
+        }
 
         $this->instantiator = $instantiator;
     }
--- a/tests/Fixture/Entity/WithUniqueColumn.php
+++ b/tests/Fixture/Entity/WithUniqueColumn.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the zenstruck/foundry package.
+ *
+ * (c) Kevin Bond <kevinbond@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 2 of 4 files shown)


nikophil:
haha I kinda like the concept of "interesting issue" :sweat_smile: 

you're totally right about the problem. That's strange I didn't encountered it in my test project.

thanks though @HypeMC 

(don't take care of the "benchmark" workflow which is failing, it's pretty new, we still need to fine tune it)

nikophil:
released https://github.com/zenstruck/foundry/releases/tag/v2.4.2

Issue #838, opened 2025-03-07
[Feature Request] Call Repository method for saving


vincentchalamon:
## Description

I have a Factory which create a Foo Entity with a custom Repository as following:

```php
#[ORM\Entity(repositoryClass: FooRepository::class)]
class Foo
{
    // ...
}

final class FooRepository extends ServiceEntityRepository
{
    // ...

    public function save(Foo $foo): void
    {
        // ...
    }
}
```

I need to call `FooRepository::save` when the Entity is saved, to perform some custom behavior.

## Problem Statement

Currently, the `PersistentObjectFactory::create` method calls `Configuration::persistence::save`, which is internal so it cannot easily be overridden (and I guess it doesn't intend to be). Additionally, this method is overridden by `PersistentProxyObjectFactory::create` which is final.

Therefore, it's not possible to override the save process of an Entity in a Factory without duplicating the whole `PersistentObjectFactory::create` method and disabling the proxy feature (which I found useful).

## Alternative Solution

A solution would be to use [Foundry hooks](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#hooks) (`afterInstanciate` and `afterPersist`), but this would require to duplicate the custom behavior code from the Repository::save method, or change my code design (move this custom behavior to a public method, for instance, but it would be inappropriate)

## Proposal

Allow to **optionally** configure a Repository method to save the Entity would solve my issue:

```php
final class FooFactory extends PersistentProxyObjectFactory
{
    // ...

    #[\Override]
    protected function initialize(): static
    {
        return $this
            // The repository might be injected in the callback
            // but I thought to maybe let the developer handle the persist himself/herself
            // in case of very-custom-persistence-not-using-the-repository (yes it is a valid word 😆)
            ->persistWith(fn (Foo $foo): void => self::repository()->save($foo));
    }
}
```

If this custom persistence callback is not set, the default behavior would continue (`$om->persist()`).

Of course, this persistence method would be call only if the persistence is enabled 😃

![Image](https://github.com/user-attachments/assets/1cd23275-1595-4782-bd6f-ecebeec63350)

dgoosens:
sounds like an excellent idea @vincentchalamon !

nikophil:
hi @vincentchalamon 👋 

does this `save()` method call `$em->flush()`?

[...]

Issue #695, opened 2024-09-30
Custom persister logic?


mpdude:
I'd like to use Foundry for a legacy code base where objects are not persisted through Doctrine ORM, but other means.

Currently, my factories for such legacy model objects call `$this->withoutPersisting()` in their `initialize()` method. All clients using these factories call `create()` or similar and then have to take care of persisting the objects (when necessary) themselves.

Is there a standard way or hook that I could use to centralize custom persistence code in the factories?

nikophil:
Hi

Basically, what I'd do would be something like this:
```php
abstract class LegacyObjectFactory extends ObjectFactory
{
    public function __construct(
        private LegacyPersister $legacyPersister
    ){}

    public function initialize(): static
    {
        return $this->afterInstatiate(
            function(object $data): void {$this->legacyPersister->save($data)}
        );
    }
}
```

~Note that you won't be able to use these factories in a data provider, because they will need Symfony to be booted (using a constructor in your factory makes it a "service factory")~ not true anymore since https://github.com/zenstruck/foundry/releases/tag/v2.2.0

in the future, we may offer an extension point dedicated for this, but there's still work to do before that

sirpilan:
Got a similar problem which would be resolved by this.

I am using a `prePersist` `EntityListener` on my Entity. In my case i automatically add the current user as owner.

When creating fixtues now:

`MyEntity::createOne(['owner' => SOME_OWNER])`
`MyEntity::createOne(['owner' => ANOTHER_OWNER])`

`owner` will allways be overwritten in my `prePersist` - so i cant create fixtues i want to have.

A Custom persister-logic would help here too. Or is there another way to avoid this behavior?

nikophil:
Hi @sirpilan 

maybe you can change your listener and not override when a value already exists? :sweat_smile: 

I'm not sure what would be the benefits of a custom persister in your case?  Maybe the best option would be to be able to disable doctrine's listeners from the factories :thinking: [A previous attempt exists](https://github.com/zenstruck/foundry/pull/216), but the PR was never finished (and, by the way, it should now target branch `2.x`, so I'm not sure something from this PR can be reused)

sirpilan:
> maybe you can change your listener and not override when a value already exists? 😅

Well, I need the user at the entity i create explicitly here: `MyEntity::createOne(['owner' => SOME_OWNER])`
[...]

PR #882, opened 2025-04-19
chore: decouple from framework bundle

nikophil:
[This discussion](https://phpc.social/@Crell/114354438064545079) made me realize that we're forcing the installation of `symfony/framework-bundle` and `doctrine/persistence` although Foundry could run without them.

This would allow to use Foundry as a standalone library.

At some point I would really like to provide an abstraction in order to allow custom persistence (see https://github.com/zenstruck/foundry/issues/695). This is already kinda doable using [this trick](https://github.com/zenstruck/foundry/issues/838#issuecomment-2705884373), but a proper persistence abstraction could be better. I even think it should not be very complex to do (as long as we don't handle relationships :sweat_smile:  - BTW that's what we do with Mongo)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,10 +17,8 @@
     ],
     "require": {
         "php": ">=8.1",
-        "doctrine/persistence": "^2.0|^3.0|^4.0",
         "fakerphp/faker": "^1.23",
         "symfony/deprecation-contracts": "^2.2|^3.0",
-        "symfony/framework-bundle": "^6.4|^7.0",
         "symfony/property-access": "^6.4|^7.0",
         "symfony/property-info": "^6.4|^7.0",
         "symfony/var-exporter": "^6.4.9|~7.0.9|^7.1.2",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -76,6 +76,10 @@
             "symfony/runtime": false
         }
     },
+    "conflict": {
+        "doctrine/persistence": "<2.0",
+        "symfony/framework-bundle": "<6.4"
```
nikophil, line 81: at some point in would be nice to add more conflicts, mainly for doctrine libs

kbond (approved):
Looks good!

kbond:
Did the CI not run for this?

nikophil:
I don't know why, sometimes it fails to run... it happens sometimes. I've rebased without touching the workflow file, and it runs now

nikophil:
lol https://github.com/zenstruck/foundry/actions/runs/14549732210/job/40819482826?pr=882

why does that pop up now? :thinking:

kbond:
Maybe it was retroactively added to 6.4?

nikophil:
> Maybe it was retroactively added to 6.4?

don't know, but it did not show up in the other PR where I re-enabled failing on deprecs

PR #885, opened 2025-04-27
feat: enable "flush once" with configuration

nikophil:
Here is a PR where we can control whether "flush once" is enabled or not

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1174,6 +1174,27 @@ The ``flush_after()`` function forwards the callback’s return, in case you nee
         TagFactory::createOne(),
     ]);
 
+Flush once
+~~~~~~~~~~
+
+Foundry used to call ``ObjectManager::flush()`` for every entity (or ODM document) created. This could have a performance
+downside. Since 2.5, Foundry is able to only call ``flush()`` once per call of ``PersistentObjectFactory::create()`` in
+userland. "Flush once" can change some behaviors in a subtle way, so it can be enabled by configuration. Not enabling it
+is deprecated, but you can migrate at your own pace.
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        zenstruck_foundry:
+            persistence:
+                flush_once: true
+
+.. versionadded::  2.5
+
+    Flush once capability was introduced in Foundry 2.5.
+
+
 Not-persisted objects factory
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/src/ZenstruckFoundryBundle.php
+++ b/src/ZenstruckFoundryBundle.php
@@ -96,6 +96,15 @@ final class ZenstruckFoundryBundle extends AbstractBundle implements CompilerPas
                     ->info('Stories or invokable services to be loaded before each test.')
                     ->scalarPrototype()->end()
                 ->end()
+                ->arrayNode('persistence')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->booleanNode('flush_once')
+                            ->info('Flush only once per call of `PersistentObjectFactory::create()` in userland.')
+                            ->defaultFalse()
+                        ->end()
+                    ->end()
+                ->end()
                 ->arrayNode('orm')
                     ->addDefaultsIfNotSet()
                     ->children()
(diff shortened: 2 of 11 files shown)


docs/index.rst
```diff
@@ -1174,6 +1174,27 @@ The ``flush_after()`` function forwards the callback’s return, in case you nee
         TagFactory::createOne(),
     ]);
 
+Flush once
+~~~~~~~~~~
+
+Foundry used to call ``ObjectManager::flush()`` for every entity (or ODM document) created. This could have a performance
+downside. Since 2.5, Foundry is able to only call ``flush()`` once per call of ``PersistentObjectFactory::create()`` in
+userland. "Flush once" can change some behaviors in a subtle way, so it can be enabled by configuration. Not enabling it
+is deprecated, but you can migrate at your own pace.
```
nikophil, line 1183: don't hesitate to fix my english :see_no_evil: :sweat_smile:

src/ZenstruckFoundryBundle.php
```diff
@@ -94,6 +94,15 @@ public function configure(DefinitionConfigurator $definition): void
...
+                ->arrayNode('persistence')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->booleanNode('flush_once')
+                            ->info('Flush only once per call of `PersistentObjectFactory::create()` in userland.')
+                            ->defaultFalse()
+                        ->end()
+                    ->end()
+                ->end()
                 ->arrayNode('orm')
```
nikophil, line 108: I'm wondering if at some point, it would not be accurate to move all `orm` and `mongo` config under `persistence` node :thinking:

tests/Integration/ORM/EntityRelationship/StandardEntityFlushOnceDisabledFactoryRelationshipTest.php
```diff
@@ -0,0 +1,32 @@
...
+use PHPUnit\Framework\Attributes\RequiresPhpunit;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+/**
+ * @author Nicolas PHILIPPE <nikophil@gmail.com>
+ * @requires PHPUnit >=11.4
+ */
+#[RequiresPhpunit('>=11.4')]
+#[IgnoreDeprecations]
+final class StandardEntityFlushOnceDisabledFactoryRelationshipTest extends StandardEntityFactoryRelationshipTest
```
nikophil, line 26: I still wanted to test with "multiple flushes" but not wanted to introduce a new test permutation, I think this way we test enough things

tests/Integration/ORM/EntityRelationship/StandardEntityFlushOnceDisabledFactoryRelationshipTest.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    protected static function bootKernel(array $options = []): KernelInterface
+    {
+        return parent::bootKernel(['environment' => 'disable_flush_once'] + $options);
```
nikophil, line 30: I still wanted to test with "flush once" disabled, but not wanted to introduce a new test permutation, I think this way we test enough things

src/ZenstruckFoundryBundle.php
kbond: Yep, that would make sense.

src/ZenstruckFoundryBundle.php
```diff
@@ -284,6 +293,12 @@ public function loadExtension(array $config, ContainerConfigurator $configurator
         $configurator->import('../config/in_memory.php');
 
         $container->registerForAutoconfiguration(InMemoryRepository::class)->addTag('foundry.in_memory.repository');
+
+        if ($config['persistence']['flush_once'] === false) {
```
kbond, line 297: So what's the plan if some users have issues with making `true`? Try and fix the use-case? If we get too many issues, allow false?

kbond (approved):
Looks good! We'll have to update the recipe to enable by default.

src/ZenstruckFoundryBundle.php
nikophil: I might do this some day (before Foundry 3 :sweat_smile:)

src/ZenstruckFoundryBundle.php
nikophil: > Try and fix the use-case

yes

> If we get too many issues, allow false?

why not, as you can see, the change to have "flush once" is really small. But I'd would really prefer to support only one way to do this :crossed_fingers: 

I think that 95% of the time this will be OK. Some of the "bugs" could not be fixed, mainly because they won't be bugs but subtle changes in the behavior. the example with "soft-delete" I told you about is a perfect example. But it was quite complex to understand what was happening :grimacing:

PR #887, opened 2025-04-28
feat: add doctrine adapter repository for in-memory

nikophil:
I got a problem here: originally I wanted to extract an interface from `RepositoryDecorator`, but we've defined `RepositoryDecorator` to be a `@template I of ObjectRepository`, but "in-memory" repository might not extends `ObjectRepository` (well, actually the should not extend it :sweat_smile:), so we have some conflicts here.

Thus, those method prototypes are problematic:
```php
     // src/Persistence/PersistentObjectFactory.php

    /**
     * @return RepositoryDecorator<T,ObjectRepository<T>>
     */
    public static function repository(): ObjectRepository;

    // src/Persistence/RepositoryDecorator.php

    /**
     * @return ObjectRepository<T>
     */
    public function inner(): ObjectRepository
```

hence this "Doctrine adapter in-memory repository"


kbond (commented):
> but "in-memory" repository might not extends ObjectRepository

`DoctrineInMemoryObjectRepository` does - are you referring to a user's custom in-memory repos?

nikophil:
> DoctrineInMemoryObjectRepository does - are you referring to a user's custom in-memory repos?

no, I was explaining _why_ I needed to do this... this feels odd: we're using "in-memory" stuff, but still need to introduce something related to doctrine... This is because we are still coupled to Doctrine here.

in the future, I'll decouple this, and introduce our own interface.

by the way, I've renamed this class `InMemoryDoctrineObjectRepositoryAdapter`, which is more acurate

kbond (approved):

PR #892, opened 2025-05-04
tests: add test permutation with multiple migration configuration files

### LINKED ISSUE #889: FrozenDependencies exception when using multiple migration configurations with ResetDatabase in mode: migrate
I'm encountering an issue when using ResetDatabase trait with multiple Doctrine migration configuration .yaml files in mode: migrate.

I have multiple .yaml migration configuration files — one for each entity manager and corresponding database:
```yaml
zenstruck_foundry:
    orm:
        reset:
            mode: migrate
            migrations:
                configurations:
                    - './config/migrations/database1.yaml'
                    - './config/migrations/database2.yaml'
```
Each configuration specifies a different em and migrations_paths

## The Problem
When the first migration is executed, it runs successfully. However, on the second configuration, I get Doctrine `FrozenDependencies` exception: `The dependencies are frozen and cannot be edited anymore.`

Full error message from my logger:
```json
{
  "message": "Error thrown while running command \"doctrine:migrations:migrate --configuration='./config/migrations/database2.yaml' --no-interaction\". Message: \"The dependencies are frozen and cannot be edited anymore.\"",
  "context": {
    "exception": {
      "class": "Doctrine\\Migrations\\Exception\\FrozenDependencies",
      "message": "The dependencies are frozen and cannot be edited anymore.",
      "file": "/app/vendor/doctrine/migrations/src/Exception/FrozenDependencies.php:13"
    }
  }
}
```

It appears that the internal container dependencies for migrations are frozen after the first migration, so when it attempts to run the second, it fails due to immutability.

I expected Foundry to sequentially run all listed migration configurations (each with its own entity manager) during database reset in test environment.

Is there a way to reset multiple databases using migration mode with separate config files for each?
Or does this require some workarounds?

nikophil:
fixes https://github.com/zenstruck/foundry/issues/889

BTW, I've figured that the CI was not totally testing the migration stuff :grimacing: 

but now it is fixed!

needs https://github.com/doctrine/migrations/pull/1528

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
         "doctrine/common": "^3.2.2",
         "doctrine/doctrine-bundle": "^2.10",
         "doctrine/doctrine-migrations-bundle": "^2.2|^3.0",
+        "doctrine/migrations": "dev-feat/unfroze-dependencies-after-command as 3.9.5",
         "doctrine/mongodb-odm": "^2.4",
         "doctrine/mongodb-odm-bundle": "^4.6|^5.0",
         "doctrine/orm": "^2.16|^3.0",
@@ -64,6 +65,12 @@
             "src/symfony_console.php"
         ]
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/nikophil/migrations"
+        }
+    ],
     "autoload-dev": {
         "psr-4": {
             "Zenstruck\\Foundry\\Tests\\": ["tests/"],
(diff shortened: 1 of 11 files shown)


composer.json
```diff
@@ -33,6 +33,7 @@
         "doctrine/common": "^3.2.2",
         "doctrine/doctrine-bundle": "^2.10",
         "doctrine/doctrine-migrations-bundle": "^2.2|^3.0",
+        "doctrine/migrations": "dev-feat/unfroze-dependencies-after-command as 3.9.5",
```
nikophil, line 36: https://github.com/doctrine/migrations/pull/1528

PR #893, opened 2025-05-05
feat: replace proxy mechanism with PHP 8.4 lazy objects

nikophil:
relates to https://github.com/zenstruck/foundry/issues/899

I think we'll release this in Foundry 2.6


src/Persistence/Proxy/CreatedObjectsTracker.php
```diff
@@ -0,0 +1,51 @@
...
+            )
+        );
+    }
+
+    public static function reset(): void
+    {
+        self::$buffer = [];
+    }
+
+    public static function countObjects(): int
```
nikophil: those two "count" methods are here for the sake of the tests, but I think we can keep them

src/Persistence/Proxy/CreatedObjectsTracker.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public static function proxifyObjects(): void
+    {
+        self::$buffer = array_values(
+            array_map(
+                static function (WeakReference $weakRef) {
+                    $object = $weakRef->get() ?? throw new \LogicException('Object cannot be null.');
+                    $clone = clone $object;
+                    (new \ReflectionClass($object))->resetAsLazyProxy($object, fn() => refresh($clone));
```
nikophil: cloning the object is even easier in order to be able to refresh it!

as far as I can tell, it does not create any "identity problem"

src/Persistence/Proxy/CreatedObjectsTracker.php
```diff
@@ -0,0 +1,51 @@
...
+use function Zenstruck\Foundry\Persistence\refresh;
+
+final class CreatedObjectsTracker
+{
+    /** @var list<\WeakReference<object>> */
+    private static $buffer = [];
+
+    public static function add(object $object): void
+    {
+        self::$buffer[] = \WeakReference::create($object);
```
nikophil: using `WeakReference` allows to not track too much objects

kbond: Not WeakMap?

config/persistence.php
```diff
@@ -18,4 +20,10 @@
                 tagged_iterator('.foundry.persistence.schema_resetter'),
             ])
     ;
+
+    if (PHP_VERSION_ID >= 80400) {
+        $container->services()->set('.foundry.proxy.kernel_terminate_listener', KernelTerminateListener::class)
+            ->tag('kernel.event_listener', ['event' => TerminateEvent::class, 'method' => '__invoke'])
```
kbond: Command.terminate too? Maybe make the listener more generic so we can add more events as we find them.

src/Persistence/Proxy/CreatedObjectsTracker.php
nikophil: How I understand it, `WeakMap` is used when you need the object to be the _key_ in the dictionnary

what I'd need is a `WeakList`, but this does not exist, so I'm dealing with a list of `WeakReference`

kbond: Ah, got it!

config/persistence.php
nikophil: yeah I was planning to do that! Maybe also `WorkerMessageHandledEvent`? or `WorkerStoppedEvent`?

kbond: Let's use kernel.reset (on ObjectTracker directly) and remove the need for a listener

tests/Integration/ORM/ProxyPHP84Test.php
```diff
@@ -0,0 +1,94 @@
...
+use PHPUnit\Framework\Attributes\Depends;
+use PHPUnit\Framework\Attributes\RequiresPhp;
+use PHPUnit\Framework\Attributes\Test;
+use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use Zenstruck\Foundry\Persistence\Proxy\CreatedObjectsTracker;
+use Zenstruck\Foundry\Test\Factories;
+use Zenstruck\Foundry\Tests\Fixture\Factories\Entity\Contact\ContactFactory;
+use Zenstruck\Foundry\Tests\Fixture\Factories\Entity\GenericEntityFactory;
+
+final class ProxyPHP84Test extends WebTestCase
```
kbond: Let's make sure we test:
1. An object was deleted
2. What happens if you re-proxify over and over

kbond (commented):
When called without args, let's make the `refresh()` function proxify everything.

src/Persistence/PersistentObjectFactory.php
```diff
@@ -454,6 +456,10 @@ final protected function initializeInternal(): static
         return parent::initializeInternal()
             ->afterInstantiate(
                 static function(object $object, array $parameters, PersistentObjectFactory $factoryUsed): void {
+                    if (PHP_VERSION_ID >= 80400) {
```
kbond: We need to make sure we don't track objects creates with `PersistentProxyObjectFactory`

nikophil:
> When called without args, let's make the refresh() function proxify everything.

this feels weird, `refresh()` function always returns something:
```php
function refresh(object &$object): object
{
    return Configuration::instance()->persistence()->refresh($object);
}
```

maybe we could add a new function?

kbond:
> this feels weird, refresh() function always returns something:

Hmm yeah, forgot about that. maybe a `refresh_all()`

nikophil:
`refresh_all()` done :+1:

nikophil:
OK CI is now green :slightly_smiling_face: 

I've deprecated `PersistentProxyObjectFactory` for PHP 8.4 and handled the deprecations in our testsuite 

I've listed [in an issue](https://github.com/zenstruck/foundry/issues/899) what needs to be done before releasing this

Benchmark is showing a small performance downgrade, which seems totally legit, because we're performing more requests than before with `PersistentObjectFactory`, but I'm pretty sure that "new Proxy 8.4 Factory" VS the old `PersistentProxyObjectFactory` (which basically everybody uses)  is equivalent in performance, or maybe better
```
+--------------------------------+-------------------+-----+------+-----+-----------------+------------------+----------------+
| benchmark                      | subject           | set | revs | its | mem_peak        | mode             | rstdev         |
+--------------------------------+-------------------+-----+------+-----+-----------------+------------------+----------------+
| ContactFactoryBench            | bench_create      |     | 10   | 5   | 17.728mb +0.24% | 2.038ms +0.61%   | ±5.23% +32.84% |
| ContactFactoryBench            | bench_create_many | 1   | 10   | 5   | 17.730mb +0.24% | 2.185ms +7.07%   | ±3.36% +45.21% |
| ContactFactoryBench            | bench_create_many | 10  | 10   | 5   | 18.610mb +0.49% | 10.736ms +2.39%  | ±1.80% +32.57% |
| ContactFactoryBench            | bench_create_many | 50  | 10   | 5   | 22.730mb +1.06% | 46.839ms -0.83%  | ±1.03% +1.36%  |
| CategoryFactoryBench           | bench_create      |     | 10   | 5   | 17.945mb +0.33% | 5.206ms +6.12%   | ±1.53% -13.05% |
| CategoryFactoryBench           | bench_create_many | 1   | 10   | 5   | 17.957mb +0.33% | 5.490ms +2.20%   | ±3.03% -15.99% |
| CategoryFactoryBench           | bench_create_many | 10  | 10   | 5   | 21.388mb +1.03% | 38.583ms +0.33%  | ±0.67% -46.50% |
| CategoryFactoryBench           | bench_create_many | 50  | 10   | 5   | 36.158mb +2.20% | 182.886ms -0.06% | ±0.47% -49.97% |
| PersistentDocumentFactoryBench | bench_create      |     | 10   | 5   | 16.240mb +0.24% | 746.168μs -1.00% | ±2.78% +83.50% |
| PersistentDocumentFactoryBench | bench_create_many | 1   | 10   | 5   | 16.245mb +0.24% | 774.627μs -0.07% | ±1.48% -11.57% |
| PersistentDocumentFactoryBench | bench_create_many | 10  | 10   | 5   | 16.310mb +0.74% | 4.824ms -1.42%   | ±4.77% +13.25% |
| PersistentDocumentFactoryBench | bench_create_many | 50  | 10   | 5   | 17.943mb +1.07% | 22.993ms +0.52%  | ±1.01% -45.65% |
+--------------------------------+-------------------+-----+------+-----+-----------------+------------------+----------------+
```

config/persistence.php
```diff
@@ -18,4 +22,13 @@
                 tagged_iterator('.foundry.persistence.schema_resetter'),
             ])
     ;
+
+    if (PHP_VERSION_ID >= 80400) {
+        $container->services()->set('.foundry.persistence.objects_tracker', ObjectsTracker::class)
+            ->tag('kernel.reset', ['method' => 'refresh'])
+            ->tag('kernel.event_listener', ['event' => TerminateEvent::class, 'method' => 'refresh'])
+            ->tag('kernel.event_listener', ['event' => ConsoleTerminateEvent::class, 'method' => 'refresh'])
+            ->tag('kernel.event_listener', ['event' => WorkerMessageHandledEvent::class, 'method' => 'refresh']) // @phpstan-ignore class.notFound
```
nikophil, line 31: I have some doubts that this is needed, but I think it adds almost zero overhead, so let's keep it for now

config/persistence.php
```diff
@@ -18,4 +22,13 @@
                 tagged_iterator('.foundry.persistence.schema_resetter'),
             ])
     ;
+
+    if (PHP_VERSION_ID >= 80400) {
+        $container->services()->set('.foundry.persistence.objects_tracker', PersistedObjectsTracker::class)
+            ->tag('kernel.reset', ['method' => 'refresh'])
+            ->tag('kernel.event_listener', ['event' => TerminateEvent::class, 'method' => 'refresh'])
+            ->tag('kernel.event_listener', ['event' => ConsoleTerminateEvent::class, 'method' => 'refresh'])
```
nikophil, line 30: `CommandTester` does not trigger `ConsoleTerminateEvent`, nor it resets the kernel, so this _might_ not be useful.

But you mentioned the `ConsoleTerminateEvent` is triggered with zenstruck/console-test`, so let's keep it this way

config/persistence.php
```diff
@@ -18,4 +22,13 @@
                 tagged_iterator('.foundry.persistence.schema_resetter'),
             ])
     ;
+
+    if (PHP_VERSION_ID >= 80400) {
+        $container->services()->set('.foundry.persistence.objects_tracker', PersistedObjectsTracker::class)
+            ->tag('kernel.reset', ['method' => 'refresh'])
+            ->tag('kernel.event_listener', ['event' => TerminateEvent::class, 'method' => 'refresh'])
```
nikophil, line 29: For the record, `TerminateEvent` is needed since the kernel is only refreshed before the SECOND call to `$client->request()`

phpstan.neon
```diff
@@ -69,6 +69,11 @@ parameters:
         - identifier: missingType.callable
           path: tests/Fixture/Maker/expected/
 
+        # we're currently running static analysis with PHP 8.3
```
nikophil, line 72: I tried to use php 8.4 for the static analysis workflow, but psalm is complaining for stuff I don't want to fix now :sweat_smile:  - maybe at some point I'll split the workflows

phpunit
```diff
@@ -4,7 +4,8 @@ set -o errexit
 set -o nounset
 
 check_phpunit_version() {
-    INSTALLED_PHPUNIT_VERSION=$(composer info phpunit/phpunit | grep versions |  cut -c 14-)
+    COMPOSER_BIN=$(which composer)
+    INSTALLED_PHPUNIT_VERSION=$(php -d xdebug.mode=off ${COMPOSER_BIN} info phpunit/phpunit | grep versions |  cut -c 14-)
```
nikophil, line 8: this has nothing to deal with this fixture, but when using step-by-step debugger, xdebug was stopping this command

phpunit
```diff
@@ -3,8 +3,10 @@
 set -o errexit
 set -o nounset
 
+COMPOSER_BIN="php -d xdebug.mode=off $(which composer)"
```
nikophil, line 6: this has nothing to deal with this feature, but when using step-by-step debugger, xdebug was stopping this command

src/Persistence/PersistenceManager.php
```diff
@@ -157,6 +159,11 @@ public function refresh(object &$object, bool $force = false): object
             return $object->_refresh();
         }
 
+        if (\PHP_VERSION_ID >= 80400 && ($reflector = new \ReflectionClass($object))->isUninitializedLazyObject($object)) {
+            /** @var T $object */
+            $object = $reflector->initializeLazyObject($object);
```
nikophil, line 164: this is basically the same thing that the line before, but for new proxy system

src/Persistence/PersistenceManager.php
```diff
@@ -165,26 +172,45 @@ public function refresh(object &$object, bool $force = false): object
...
-
-            return $object;
-        }
-
         if ($strategy->isEmbeddable($object)) {
             return $object;
         }
 
-        $id = $om->getClassMetadata($object::class)->getIdentifierValues($object);
+        if (!$strategy->contains($object)) {
```
nikophil, line 179: I had to refactor how the refresh was done, this fixes two things:
- if the object is not currently manage by the orm/odm, we need to perform a "merge"
- some stuff had to be made to handle deleted objects. I also manually set to null ids fields that are nullable. Not sure that's the right approach :shrug:

src/Persistence/PersistentProxyObjectFactory.php
```diff
@@ -22,6 +22,19 @@
...
 {
+    public function __construct()
+    {
+        parent::__construct();
+
+        if (\PHP_VERSION_ID >= 80400) {
+            trigger_deprecation(
+                'zenstruck/foundry',
+                '2.6',
+                'Proxy usage is deprecated in PHP 8.4. Use directly PersistentObjectFactory, Foundry now leverages the native PHP lazy system to auto-refresh objects.',
```
nikophil, line 33: I think we will need this message to link to the documentation when it will be read

src/Persistence/Proxy/PersistedObjectsTracker.php
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+namespace Zenstruck\Foundry\Persistence\Proxy;
+
+use Zenstruck\Foundry\Configuration;
+
+/**
+ * @internal
+ */
+final class PersistedObjectsTracker
```
nikophil, line 19: finally I found a name I do like :+1:

src/Persistence/Proxy/PersistedObjectsTracker.php
```diff
@@ -0,0 +1,78 @@
...
+     * This buffer of objects needs to be static to be kept between two kernel.reset events.
+     *
+     * @var list<\WeakReference<object>>
+     */
+    private static $buffer = [];
+
+    public function refresh(): void
+    {
+        self::proxifyObjects();
+    }
```
nikophil, line 31: `proxifyObjects()` being now a private method, this non-static method is the only entrypoint to trigger the refresh

src/Persistence/Proxy/PersistedObjectsTracker.php
```diff
@@ -0,0 +1,78 @@
...
+
+                    if ($reflector->isUninitializedLazyObject($object)) {
+                        return \WeakReference::create($object);
+                    }
+
+                    $clone = clone $object;
+                    $reflector->resetAsLazyProxy($object, function() use ($clone, $reflector) {
+                        Configuration::instance()->persistence()->refresh($clone, allowRefreshDeletedObject: true);
+
+                        return $reflector->initializeLazyObject($clone);
```
nikophil, line 67: I had hard times dealing with deleted objects, PHP always told me that "Lazy proxy factory must return a non-lazy object" even if I was pretty certain that the object returned was not lazy :thinking: 

the only way I found to fix this was to call `initializeLazyObject()`. Maybe that's because of Doctrine which also deals with proxies? I don't know...

Based on the [documentation](https://www.php.net/manual/en/reflectionclass.initializelazyobject.php), calling this method, even on "non lazy" objects is harmless

src/Persistence/Proxy/PersistedObjectsTracker.php
```diff
@@ -0,0 +1,78 @@
...
+                static function(\WeakReference $weakRef) {
+                    $object = $weakRef->get() ?? throw new \LogicException('Object cannot be null.');
+
+                    $reflector = new \ReflectionClass($object);
+
+                    if ($reflector->isUninitializedLazyObject($object)) {
+                        return \WeakReference::create($object);
+                    }
+
+                    $clone = clone $object;
```
nikophil, line 63: I'm wondering if at some point we won't be bitten by this :thinking: 

this was easier to do this instead of extracting the object ids and using them in some query, it would have required a new method in `PersistenceManager` I think

tests/Integration/Persistence/ProxyPHP84TestCase.php
```diff
@@ -0,0 +1,192 @@
...
+        self::assertFalse((new \ReflectionClass($object))->isUninitializedLazyObject($object));
+    }
+
+    /**
+     * @test
+     * @requires PHP >= 8.4
+     */
+    #[Test]
+    #[RequiresPhp('>= 8.4')]
+    public function it_can_refresh_after_command_terminated(): void
```
nikophil, line 149: this test is only here to confirm that using `CommandTester` updates the objects, but since `ConsoleTerminate` nor `kernel.reset` are triggered, the proxy mechanism is not involved at all in this test :shrug: 

this could be problematic, because if the command would call `$em->clear()` or perform raw SQL, this test would fail :disappointed: 

I think there is nothing we can do, except documenting this (ie: "you should call `refresh_all()` after using `CommandTester`", and/or modifying how `CommandTester` works, and make it trigger one of the event. WDYT?

kbond (commented):
Looking good! Thanks for working on this!

src/Persistence/PersistenceManager.php
```diff
@@ -165,26 +172,45 @@ public function refresh(object &$object, bool $force = false): object
...
-        if (!$id || !($object = $om->find($object::class, $id))) { // @phpstan-ignore parameterByRef.type
-            throw RefreshObjectFailed::objectNoLongExists();
+            if ($objectFromDb) {
+                $object = $objectFromDb;
+            } else {
+                if ($allowRefreshDeletedObject) {
+                    $identifierFields = $om->getClassMetadata($object::class)->getIdentifierFieldNames();
+                    foreach ($identifierFields as $field) {
+                        try {
+                            set($object, $field, null);
```
nikophil: @kbond any thoughts about this?

I'm resetting to null ids that can be set null (usually, it is the ones that are generated by the DB I guess)

src/Persistence/Proxy/PersistedObjectsTracker.php
smnandre: > the only way I found to fix this was to call initializeLazyObject(). Maybe that's because of Doctrine which also deals with proxies? I don't know...

Well, does this not remove the benefit of lazy objects ?

nikophil: no, because we actually are inside the factory of the proxy where we should definitely not return a lazy object.

At some point I was thinking this was a bug from PHP itself because I had this behavior:
```php
$reflector->resetAsLazyProxy($object, function() use ($clone, $reflector) {
    $object = Configuration::instance()->persistence()->refresh($clone, allowRefreshDeletedObject: true);

   // this check passes
   Assert::false($reflecto->isUninitializedLazyObject($object));

    // return here says "Lazy proxy factory must return a non-lazy object"
    // even if we just validated that this is not the case
    return $object;
}
```

src/Persistence/PersistenceManager.php
kbond: What would be the problem with doing away with this logic and always throwing an `objectNotLongerExists()` exception?

nikophil: this test would break badly :sweat_smile: 

```php
    public function test_remove_object(): void
    {
        $object = $this->factory()->create();

        $client = self::createClient();
        $client->request('GET', "/delete/{$object->id}");
        self::assertResponseIsSuccessful();

        // here, $object is a lazy proxy, so any access to any method/prop will throw `RefreshObjectFailed`
    }
```

nikophil: solved now: I've reverted reset of the ids, and implemented functions `assertPersisted()` / `assertNotPersisted()`

nikophil:
@kbond I think this PR is ready - still needs some work to be done (see https://github.com/zenstruck/foundry/issues/899), but in another PR

src/Persistence/functions.php
```diff
@@ -176,6 +201,24 @@ function enable_persisting(): void
     Configuration::instance()->persistence()->enablePersisting();
 }
 
+function assertPersisted(object $object, string $message = '{entity} is not persisted.'): void
```
kbond, line 204: I feel like we should stick with our current snake_case function naming.

In a future PR, thinking we should add some helper phpunit methods to either Factories or a new trait:

- `->save()`
- `->assertPersisted()`
- etc...

kbond (commented):
Looking good! Think we should release 2.6 before merging?

src/Persistence/functions.php
nikophil: > I feel like we should stick with our current snake_case function naming.

do you mean I should change to `_assertPersisted()`?

> In a future PR, thinking we should add some helper phpunit methods to either Factories or a new trait

yeah, why not

kbond: > do you mean I should change to _assertPersisted()?

`assert_persisted()` to match the other functions like `persistent_factory()`

nikophil: LOL ok sorry :sweat_smile: :facepalm:

Issue #898, opened 2025-05-06
[Story] Introduce `#[AsFixture]` attribute and `foundry:load-fixture` command


nikophil:
the `#[AsFixture]` attribute could only be added on stories:

```php
#[AsFixture(name: 'my-story', groups: ['my-group'])
public function MyStory extends Story
{}
```

then this command should load the story:
```terminal
$ bin/console foundry:load-story my-story
```

this command should load all the stories with `my-group`:
```terminal
$ bin/console foundry:load-story --group my-group
# or
$ bin/console foundry:load-story my-group # we should then take care of name collisions between groups and names
```

If the command is run without any parameters or options, it should allow selecting from the list of all stories with the attribute `#[AsFixture]`

even if I think stories should be independent between them, we could add a `priority` parameter as well.

A last idea could also to load all the stories in parallel, which could drastically improve the perfs

flohw:
Hi @nikophil 

A small typo on your first example (an extra square bracket, the parenthesis are not required too 🤔 )

```diff
- #[AsFixture(name: 'my-story', groups: [('my-group'])]
+ #[AsFixture(name: 'my-story', groups: [('my-group')]
```

If `priority` is added, maybe a notion of dependency should be introduced as doctrine fixtures bundle does? Or a story is used to build and define dependencies between factories et should load a complete scenario? I don't use stories, I didn't read the doc enough when I first used the bundle so I may have missed some details 😅 

Thanks, it looks promising 🙂

nikophil:
hi @flohw 

> A small typo on your first example

fixed, thanks

> If priority is added, maybe a notion of dependency should be introduced as doctrine fixtures bundle does?

I don't think I am willing to do this 😬 I'd really want this to be simple. For all complex scenarios, I'd rather redirect people to doctrine fixtures bundle.

> a story is used to build and define dependencies between factories et should load a complete scenario?

Yeah this is more or less how I think stories should behave. IMO stories should be independent units, that you could call separately. Nevertheless, this is not a hard requirement. Story can also call other stories, but when they are run in the same runtime, a story will be played only once, even if it is called several times.

> I don't use stories

They are pretty useful, though 🙂 

> it looks promising

thanks!

kbond:
Since stories can call other stories, I don't know a priority is required?

If you have StoryA that depends on StoryB which depends on StoryC, I'd just mark StoryA with AsStory - so no possibility of loading B & C with the command.

flohw:
[...]

PR #900, opened 2025-05-11
feat: Introduce `#[AsFixture]` attribute and `foundry:load-fixture` command

nikophil:
fixes https://github.com/zenstruck/foundry/issues/898

:warning:  I made a mistake with git, follow up in https://github.com/zenstruck/foundry/pull/903

### DIFF
--- a/config/persistence.php
+++ b/config/persistence.php
@@ -2,6 +2,7 @@
 
 namespace Symfony\Component\DependencyInjection\Loader\Configurator;
 
+use Zenstruck\Foundry\Command\LoadStoryCommand;
 use Zenstruck\Foundry\Persistence\PersistenceManager;
 use Zenstruck\Foundry\Persistence\ResetDatabase\ResetDatabaseManager;
 
@@ -16,6 +17,15 @@ return static function (ContainerConfigurator $container): void {
             ->args([
                 tagged_iterator('.foundry.persistence.database_resetter'),
                 tagged_iterator('.foundry.persistence.schema_resetter'),
+            ])
+
+        ->set('.zenstruck_foundry.story.load_story-command', LoadStoryCommand::class)
+            ->arg('$databaseResetters', tagged_iterator('.foundry.persistence.database_resetter'))
+            ->arg('$kernel', service('kernel'))
+            ->tag('console.command', [
+                'command' => 'foundry:load-stories',
+                'aliases' => ['foundry:load-fixtures'],
+                'description' => 'Load stories which are marked with #[AsFixture] attribute.',
             ])
     ;
 };
--- a/src/Command/LoadStoryCommand.php
+++ b/src/Command/LoadStoryCommand.php
@@ -13,10 +13,12 @@ namespace Zenstruck\Foundry\Command;
 
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Exception\InvalidArgumentException;
+use Symfony\Component\Console\Exception\LogicException;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\DependencyInjection\ServiceLocator;
 use Symfony\Component\HttpKernel\KernelInterface;
 use Zenstruck\Foundry\Configuration;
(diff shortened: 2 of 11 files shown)


src/ZenstruckFoundryBundle.php
```diff
@@ -221,84 +225,10 @@ public function loadExtension(array $config, ContainerConfigurator $configurator
...
-
-        if (false === $config['persistence']['flush_once']) {
-            trigger_deprecation('zenstruck/foundry', '2.5', 'Not setting "zenstruck_foundry.persistence.flush_once" to true is deprecated. This option will be forced to true in 3.0');
-        }
-
-        $container->setParameter('zenstruck_foundry.persistence.flush_once', $config['persistence']['flush_once']);
+        $this->configureMakers($configurator, $container, $config);
+        $this->configurePersistence($container, $configurator, $config);
+        $this->configureInMemory($configurator, $container);
+        $this->configureFixturesStory($configurator, $container);
```
nikophil: I think it is time to group those in methods :slightly_smiling_face: 

new code is only in the last method `configureFixturesStory()`

tests/Integration/Command/LoadStoryTest.php
```diff
@@ -0,0 +1,226 @@
...
+use Zenstruck\Foundry\Test\ResetDatabase;
+use Zenstruck\Foundry\Tests\Fixture\Factories\Entity\GenericEntityFactory;
+use Zenstruck\Foundry\Tests\Fixture\Stories\Fixtures\FixtureStory;
+use Zenstruck\Foundry\Tests\Fixture\Stories\Fixtures\FixtureStoryWithNameCollision;
+use Zenstruck\Foundry\Tests\Fixture\TestKernel;
+use Zenstruck\Foundry\Tests\Integration\RequiresORM;
+
+final class LoadStoryTest extends KernelTestCase
+{
+    use RequiresORM, ResetDatabase, Factories;
```
nikophil: those tests would work regardless the DBMS, so I only test it with ORM

src/Command/LoadStoryCommand.php
```diff
@@ -0,0 +1,117 @@
...
+
+        return self::SUCCESS;
+    }
+
+    private function resetDatabase(): void
+    {
+        foreach ($this->databaseResetters as $databaseResetter) {
+            $databaseResetter->resetBeforeFirstTest($this->kernel);
+        }
+    }
```
nikophil: not sure what to do with global state here :thinking: 

maybe add an option to disable it?

kbond: Hmm, good question. You could certainly have one global state for tests, and another for dev but since most users probably have the same config for each, it could be unexpected to have it added here by default.

I'm leaning to not including in the command right now: "global state is only for tests". In the future, if asked, we could add a `--load-global-state` option.

kbond (commented):
This is great!

From a DX perspective, I'd really like:
1. When you install foundry, the recipe creates an AppStory (like the doctrine-fixtures recipe does)
2. Just run `bin/console foundry:load-stories` (no arguments)

So, what do you think about: if you pass the command no args and there's only one services tagged `#[AsFixture]`, load that one?

config/persistence.php
```diff
@@ -17,5 +18,14 @@
                 tagged_iterator('.foundry.persistence.database_resetter'),
                 tagged_iterator('.foundry.persistence.schema_resetter'),
             ])
+
+        ->set('.zenstruck_foundry.story.load_story-command', LoadStoryCommand::class)
+            ->arg('$databaseResetters', tagged_iterator('.foundry.persistence.database_resetter'))
+            ->arg('$kernel', service('kernel'))
+            ->tag('console.command', [
+                'command' => 'foundry:load-stories',
+                'aliases' => ['foundry:load-fixtures'],
```
kbond:
```suggestion
                'aliases' => ['foundry:load-fixtures', 'foundry:load-story'],
```

I'd like both `foundry:load-story` and `foundry:load-fixture` to work (`foundry:load-fixture` already works).

nikophil:
> From a DX perspective, I'd really like:
> 
> - When you install foundry, the recipe creates an AppStory (like the doctrine-fixtures recipe does)
> - run bin/console foundry:load-stories (no arguments)
> 
> So, what do you think about: if you pass the command no args and there's only one services tagged #[AsFixture], load that one?

:100:

src/Command/LoadStoryCommand.php
nikophil: skipping global state is actually more complex than I expected :shrug: 

everything is public (so that it could be decorated in userland), so I can't add a new param in the existing public methods :grimacing:

nikophil: ok, actually, we're calling `BeforeFirstTestResetter` which only loads global state if dama is enabled.

but dama should be disabled when running this command (otherwise, nothing will be persisted at the end of the command)... not so complex, then :slightly_smiling_face:

Issue #894, opened 2025-05-05
[PHPStan] improve `FactoryCollection` generics annotation with `template-type`


edglev:
Looks like there is an issue with [phpstan type definition](https://github.com/zenstruck/foundry/blob/2.x/src/Persistence/PersistentProxyObjectFactory.php#L21), specifically `&Proxy<T>` part breaking extension chain resolving to `Factory<T>`.

I don't understand why `T&Proxy<T>` is a not a subtype of `object`.
Is this a problem in PhpStan itself?

Minimal reproducer https://phpstan.org/r/eece4718-641c-4e5b-bac9-e4f1430f8fda

Just as a test - removing  `&Proxy<T>` part from PersistentProxyObjectFactory makes PhpStan happy.

https://phpstan.org/r/dc467615-ce04-4f6c-8370-b6be346ec8c0

nikophil:
fixed 😄 

https://phpstan.org/r/441da82d-53cc-40ee-8b4e-4ff82b7988b6

```diff
- @param FactoryCollection<Something, SomethingFactory> $somethingFactories
+ @param FactoryCollection<Something&Proxy<Something>, SomethingFactory> $somethingFactories
```

those `FactoryCollection` with proxies are kinda hard to deal with. We're open to any good idea to simplify this :)

By the way, the only use case to type hint with `FactoryCollection` is when dealing with data provider, but since Foundry 2.2 and PHPUnit 11.4, you can directly call `YouFactory::createMany()` in your data provider. you'll need to use the phpunit extension provided by Foundry. see https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#phpunit-data-providers

edglev:
Thank you :heart:  The use case for type hints is with data providers.

edglev:
I've might have just found a way to simplify this - you can interpret the type from TFactory, example taken from here https://github.com/phpstan/phpstan/discussions/9053

https://phpstan.org/r/7a4e6c92-1d39-4470-860f-890d47aee32c

What do you think?

nikophil:
this is very cool! I didn't even know `template-type` exists!

one thing I'd like, is that the current way of documenting factories still works. We've already introduced a BC break around this few months ago, and I would not want to introduce another one now...

do you think both fashions of documenting `FactoryCollection` could be compatible?

edglev:
[...]

PR #902, opened 2025-05-12
fix: simplify FactoryCollection phpstan type resolving

edglev:
for https://github.com/zenstruck/foundry/issues/894

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -103,7 +103,7 @@ abstract class Factory
     abstract public function create(array|callable $attributes = []): mixed;
 
     /**
-     * @return FactoryCollection<T, static>
+     * @return FactoryCollection<static>
      */
     final public function many(int $count): FactoryCollection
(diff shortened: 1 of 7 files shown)


edglev:
Not sure where do these PHPStan errors come from :thinking: Works on my machine :tm: 

Anyway, the bigger issue for this PR is that I'm not sure it can be made compatible with psalm without suppressing errors

nikophil:
Hi @edglev 

thanks for trying this!

> the bigger issue for this PR is that I'm not sure it can be made compatible with psalm without suppressing errors

this sounds like a blocker :grimacing: 

if the `template-type` stuff does not exist in Psalm, I don't think we're gonna use it. sorry for your wasted time :shrug:

nikophil:
I'm closing this Pr, don't hesitate to reopen it if you manage to find a solution!

PR #903, opened 2025-05-12
feat: Introduce `#[AsFixture]` attribute and `foundry:load-fixture` command

nikophil:
fixes https://github.com/zenstruck/foundry/issues/898

(follow up of https://github.com/zenstruck/foundry/pull/900 which was merged by mistake)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1160,7 +1160,7 @@ once. To do this, wrap the operations in a ``flush_after()`` callback:
         TagFactory::createMany(200); // instantiated/persisted but not flushed
     }); // single flush
 
-The ``flush_after()`` function forwards the callback’s return, in case you need to use the objects in your tests:
+The ``flush_after()`` function forwards the callback's return, in case you need to use the objects in your tests:
 
 ::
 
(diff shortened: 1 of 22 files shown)


nikophil:
@kbond I think this PR is ready:
- global state is actually not loaded (see [here](https://github.com/zenstruck/foundry/pull/900#discussion_r2083606805))
- if only one story as fixture exists, it will be used automatically if the command is run without any argument
- I've added an `AppStory` example in the recipe (see https://github.com/symfony/recipes/pull/1415)

docs/index.rst
```diff
@@ -2436,6 +2387,49 @@ You can use the ``#[WithStory]`` attribute to load stories in your tests:
...
 If used on the class, the story will be loaded before each test method.
 
+Loading stories as fixtures in your database
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.6
+
+    Command ``foundry:load-stories`` and attribute ``#[AsFixture]`` were added in 2.6.
+
+Using command ``bin/console foundry:load-stories``, you can load stories as fixtures in your database.
```
kbond, line 2397: Should we mention that if only a single story exists (the recipe default), no command arguments are required?

kbond (approved):
Awesome, let's roll!

docs/index.rst
nikophil: done

PR #906, opened 2025-05-14
Fix data providers phpunit link

alexander-schranz:
Using the `.5` version which docs exists mostly longer as the minor versions between.


nikophil (approved):

nikophil:
thanks @alexander-schranz

PR #908, opened 2025-05-20
fix: should not use flush_after() in FactoryCollection::create()

### LINKED ISSUE #907: EntityIdentityCollision after upgrade from 2.4 to 2.5
I have two factories:

```php
<?php

declare(strict_types=1);

namespace App\Currency\Foundry\Factory;

use Zenstruck\Foundry\Persistence\PersistentProxyObjectFactory;

/**
 * @extends PersistentProxyObjectFactory<Currency>
 */
final class CurrencyFactory extends PersistentProxyObjectFactory
{
    //...
}

/**
 * @extends PersistentProxyObjectFactory<Product>
 */
final class ProductFactory extends PersistentProxyObjectFactory
{
    /**
     * @return array<string, mixed>
     */
    {
        return [
            'id' => new UuidV7(),
            'pricingCurrency' => CurrencyFactory::findOrCreate(['id' => 'eur', 'code' => 'EUR', 'name' => 'Euro']),
        ];
    }

    //...
}
```

Tests were running fine, but now I started to get:

```bash
Doctrine\ORM\Exception\EntityIdentityCollisionException: While adding an entity of class App\Entity\Currency with an ID hash of "eur" to the identity map,
another object of class App\Entity\Currency was already present for the same ID. This exception
is a safeguard against an internal inconsistency - IDs should uniquely map to
entity object instances. This problem may occur if:

- you use application-provided IDs and reuse ID values;
- database-provided IDs are reassigned after truncating the database without
clearing the EntityManager;
- you might have been using EntityManager#getReference() to create a reference
for a nonexistent ID that was subsequently (by the RDBMS) assigned to another
entity.

Otherwise, it might be an ORM-internal inconsistency, please report it.

/var/www/html/vendor/doctrine/orm/src/Exception/EntityIdentityCollisionException.php:15
/var/www/html/vendor/doctrine/orm/src/UnitOfWork.php:1544
/var/www/html/vendor/doctrine/orm/src/UnitOfWork.php:1392
/var/www/html/vendor/doctrine/orm/src/UnitOfWork.php:936
/var/www/html/vendor/doctrine/orm/src/UnitOfWork.php:1806
/var/www/html/vendor/doctrine/orm/src/UnitOfWork.php:1764
/var/www/html/vendor/doctrine/orm/src/EntityManager.php:439
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistenceManager.php:103
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistentObjectFactory.php:469
/var/www/html/vendor/zenstruck/foundry/src/ObjectFactory.php:65
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistentObjectFactory.php:206
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistentProxyObjectFactory.php:41
/var/www/html/vendor/zenstruck/foundry/src/Factory.php:66
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistentProxyObjectFactory.php:50
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistentObjectFactory.php:79
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistentProxyObjectFactory.php:68
/var/www/html/src/Marketplace/Foundry/Factory/ProductFactory.php:43
/var/www/html/vendor/zenstruck/foundry/src/Factory.php:183
/var/www/html/vendor/zenstruck/foundry/src/ObjectFactory.php:49
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistentObjectFactory.php:206
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistentProxyObjectFactory.php:41
/var/www/html/vendor/zenstruck/foundry/src/FactoryCollection.php:151
/var/www/html/vendor/zenstruck/foundry/src/FactoryCollection.php:151
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistenceManager.php:121
/var/www/html/vendor/zenstruck/foundry/src/Persistence/functions.php:160
/var/www/html/vendor/zenstruck/foundry/src/FactoryCollection.php:150
/var/www/html/vendor/zenstruck/foundry/src/Factory.php:76
```

I create two products with `ProductFactory::createMany(2)` and probably it tries for both to create the currency, because if I pre-create it before creating products, then it works.

### LINKED ISSUE #911: Broken tests after upgrade to `2.5` with new `flush once` feature
Just a heads up, I'm having an issue with the new [`flush once`](https://github.com/zenstruck/foundry/pull/873) feature released in `2.5`.  Tests work when I comment out the following lines: 

https://github.com/zenstruck/foundry/blob/f9919995669477a829fd6ae55f6effceb57faee2/src/FactoryCollection.php#L149-L153

Would it be possible to add the new [`flush_once`](https://github.com/zenstruck/foundry/pull/885) config flag to that block?

---

🍕 **[Reproducer](https://github.com/phasdev/zf-flush-once)**

I have a `Pizza` with a `OneToMany` collection of `Toppings`; i.e. each `Pizza` can have 0 or more `Toppings` and each `Topping` has a non-null relationship to the owning `Pizza`. 

My `ToppingFactory` has the following `defaults()`:

```php
return [ 'pizza' => lazy(fn () => PizzaFactory::new()) ];
```

My `PizzaFactory` has the following `defaults()`: 

```php
return [
    'toppings' => lazy(fn () => ToppingFactory::createRange(1, 5, [
        'pizza' => null
    ]))
];
```
The `toppings`' `pizza` attribute is set to `null` to avoid an infinite loop and because ZF will set that attribute when instantiating the `Pizza`.

When I run the following test:

```php
PizzaFactory::createRange(1, 5);
``` 

I get the following exception:

```
[Doctrine\DBAL\Exception\NotNullConstraintViolationException (19)]
An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 19 NOT NULL constraint failed: topping.pizza_id
```

This is expected because `toppings` are flushed when they're created and don't have a reference to a `Pizza` yet. So to work around that exception, I defer flushing:

```php
flush_after(function () {
    PizzaFactory::createRange(1, 5);
});
```

This allows ZF to set the `pizza` reference on each of its `toppings` before everything is flushed. 

Unfortunately the new `flush_once` feature now flushes `toppings` (as a group) when they're created, which breaks my deferred flush workaround.

---

I'm sure there's a better way to instantiate default `toppings`, e.g. using an `afterPersist` hook, but I wanted to post this issue in case others run into it.

Thanks for this awesome library! 😁

nikophil:
this reproduces both https://github.com/zenstruck/foundry/issues/907 and https://github.com/zenstruck/foundry/issues/911 and fixes them  (two issue, same fix) - @phasdev [did add a feature flag](https://github.com/zenstruck/foundry/pull/912) for flush once, but the bug still remains.

`flush_after()` was definitely not a good idea in `FactoryCollection::create()`: "flush once" behavior should not be a hard requirement, and something like `findOrCreate()` should actually flush.

thanks again to @phasdev, the reproducers were very helpful!

fixes #907 fixes #911

### DIFF
--- a/tests/Integration/ORM/EntityRelationship/EntityFactoryRelationshipTestCase.php
+++ b/tests/Integration/ORM/EntityRelationship/EntityFactoryRelationshipTestCase.php
@@ -34,6 +34,7 @@ use Zenstruck\Foundry\Tests\Fixture\Factories\Entity\Category\CategoryFactory;
 use Zenstruck\Foundry\Tests\Fixture\Factories\Entity\Contact\ContactFactory;
 
 use function Zenstruck\Foundry\lazy;
+use function Zenstruck\Foundry\Persistence\flush_after;
 use function Zenstruck\Foundry\Persistence\refresh;
 use function Zenstruck\Foundry\Persistence\unproxy;
 
@@ -710,6 +711,44 @@ abstract class EntityFactoryRelationshipTestCase extends KernelTestCase
             )->create();
 
         self::assertNotNull($address->getContact());
+    }
+
+    /** @test */
+    #[Test]
+    #[DataProvider('provideCascadeRelationshipsCombinations')]
+    #[UsingRelationships(Contact::class, ['category'])]
+    public function find_or_create_used_in_many_actually_create_only_one_object(): void
+    {
+        static::contactFactory()
+            ->many(2)
+            ->create(['category' => lazy(fn() => static::categoryFactory()::findOrCreate([]))]);
+
+        static::contactFactory()::assert()->count(2);
+        static::categoryFactory()::assert()->count(1);
+    }
+
+    /** @test */
+    #[Test]
+    #[DataProvider('provideCascadeRelationshipsCombinations')]
+    #[UsingRelationships(Contact::class, ['category'])]
+    public function ensure_inverse_one_to_many_is_hydrated_before_persisting_in_flush_after(): void
+    {
+        flush_after(
+            static function() {
+                $contactFactory = static::contactFactory()
+                    ->afterPersist(static function(Contact $contact) {
+                        self::assertNotNull($contact->getCategory());
+                    });
+
+                static::categoryFactory()::createMany(
+                    2,
+                    ['contacts' => lazy(static fn() => $contactFactory->many(2)->create(['category' => null]))]
+                );
+            }
+        );
+
+        static::contactFactory()::assert()->count(4);
+        static::categoryFactory()::assert()->count(2);
     }
 
     /** @return PersistentObjectFactory<Contact> */
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,6 +209,7 @@ jobs:
             doctrine/doctrine-migrations-bundle \
             doctrine/mongodb-odm-bundle \
             symfony/maker-bundle \
+            symfony/phpunit-bridge \
             symfony/framework-bundle
 
       - name: Test
--- a/src/FactoryCollection.php
+++ b/src/FactoryCollection.php
@@ -14,8 +14,6 @@ namespace Zenstruck\Foundry;
 use Zenstruck\Foundry\Persistence\PersistentObjectFactory;
 use Zenstruck\Foundry\Persistence\PersistMode;
 
-use function Zenstruck\Foundry\Persistence\flush_after;
-
 /**
  * @author Kevin Bond <kevinbond@gmail.com>
  *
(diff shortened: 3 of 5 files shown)


tests/Integration/ORM/EntityRelationship/EntityFactoryRelationshipTestCase.php
```diff
@@ -712,6 +713,44 @@ static function(Address $a): void {
...
+
+        static::contactFactory()::assert()->count(2);
+        static::categoryFactory()::assert()->count(1);
+    }
+
+    /** @test */
+    #[Test]
+    #[DataProvider('provideCascadeRelationshipsCombinations')]
+    #[UsingRelationships(Contact::class, ['category'])]
+    public function ensure_inverse_one_to_many_is_hydrated_before_persisting_in_flush_after(): void
```
nikophil, line 734: I'm not 100% sure the name of the test is accurate, this edge case is pretty specific :see_no_evil: (it reproduces #911)

tests/Integration/ORM/EntityRelationship/EntityFactoryRelationshipTestCase.php
```diff
@@ -712,6 +713,44 @@ static function(Address $a): void {
         self::assertNotNull($address->getContact());
     }
 
+    /** @test */
+    #[Test]
+    #[DataProvider('provideCascadeRelationshipsCombinations')]
+    #[UsingRelationships(Contact::class, ['category'])]
+    public function find_or_create_used_in_many_actually_create_only_one_object(): void
```
nikophil, line 720: reproduces #907

nikophil:
I don't know what's happening in this PR: seg fault in codecov, paratest which takes forever... :shrug: 

EDIT : OK I know why, that's because of data providers

smnandre:
🌹

nikophil:
@smnandre does this fixes the problem you told me about?

smnandre:
Pretty sure, i'll test friday

kbond (approved):
Awesome, thanks everyone!

@smnandre be sure to let us know if this doesn't fix your issue.

smnandre:
I will, trust me 😆

nikophil:
this was released in 2.5.3 :slightly_smiling_face:

smnandre:
Not conclusive... we still have something weird with clear and collections, but it's more and more probable recent changes in Foundry simply exposed a bug that was already here in our code

PR #912, opened 2025-05-22
Add missing flush_once feature flag

### LINKED ISSUE #911: Broken tests after upgrade to `2.5` with new `flush once` feature

ghost:
PR to address https://github.com/zenstruck/foundry/issues/911. 

It looks like the `flush_once` feature flag is missing from the block that "flushes once":

https://github.com/zenstruck/foundry/blob/f9919995669477a829fd6ae55f6effceb57faee2/src/FactoryCollection.php#L149-L153 

So it's currently not possible to disable the new behaviour by setting `flush_once: false` in the ZF config. This PR integrates the flag so the `flush_once` behaviour can be disabled.

This PR doesn't fix #911, but will let me opt-out of the new behaviour while I troubleshoot/migrate my tests.

### DIFF
--- a/src/FactoryCollection.php
+++ b/src/FactoryCollection.php
@@ -11,6 +11,7 @@
 
 namespace Zenstruck\Foundry;
 
+use Zenstruck\Foundry\Configuration;
 use Zenstruck\Foundry\Persistence\PersistentObjectFactory;
 use Zenstruck\Foundry\Persistence\PersistMode;
 
@@ -146,7 +147,7 @@ final class FactoryCollection implements \IteratorAggregate
      */
     public function create(array|callable $attributes = []): array
     {
-        if ($this->isRootFactory && $this->factory instanceof PersistentObjectFactory && $this->factory->isPersisting()) {
+        if (Configuration::instance()->flushOnce && $this->isRootFactory && $this->factory instanceof PersistentObjectFactory && $this->factory->isPersisting()) {
             return flush_after(
                 fn() => \array_map(static fn(Factory $f) => $f->create($attributes), $this->all())
             );


nikophil:
hi @phasdev 

thanks for this PR and the reproducer :heart_eyes: 

I'll merge and release this PR in 2.5.1, and I'll find a fix for this case with flush onces enabled afterwards

nikophil:
released :heavy_check_mark:

Issue #784, opened 2025-01-08
Upgrade not compatible with rector 2


nesl247:
Rector 2.0 seems to have removed SetListInterface, so it is not possible to run the set list without downgrading.

The rules also don't work unfortunately:

```Could not process "/srv/app/src/ChatBot/ChatMessage.php" file, due to:
         "System error: "The `Rector\Rector\AbstractScopeAwareRector` is removed, use `Rector\Rector\AbstractRector`
         instead, see upgrading guide https://github.com/rectorphp/rector-src/blob/main/UPGRADING.md#1-abstractscopeawarerector-is-removed-use-abstractrector-instead"
         Run Rector with "--debug" option and post the report here: https://github.com/rectorphp/rector/issues/new". On line: 20```

nikophil:
Hi @nesl247 

thanks for reporting. Yeah, I was expecting problems like this with Rector 2.

I don't know when I'll have time to work on this. Don't hesitate to provide a PR :)

nesl247:
I ended up doing it manually, so I don't have a need for it anymore unfortunately. If I find some time I'll try to tackle it, but it will probably be a while before then.

nikophil:
this has been fixed in #913

Issue #899, opened 2025-05-09
[PHP8.4] Leverage PHP 8.4 lazy objects for proxy system


nikophil:
follow up for https://github.com/zenstruck/foundry/pull/893

Here is a list of things that need to be done to remove usage of `ProxyHelper::generateLazyProxy()`

- [x] ~"undeprecate" `auto_refresh_proxies` configuration~
- [x] ~add a method to `PersistentObjectFactory` to create objects that are not auto-refreshable~
- [x] add helpers around proxies that will replace methods from `IsProxy`:  
	- [x]  `delete` 
	- [x] `_assertPersisted` 
	- [x] `_assertNotPersisted` 
	- [x] ~`_withoutAutoRefresh`~
	- [x] ~`_enableAutoRefresh`~
	- [x] ~`_disableAutoRefresh`~ 
	EDIT:  i've not implemented `_withoutAutoRefresh` / `_enableAutoRefresh` / `_disableAutoRefresh` for the same reason we don't provide an `auto_refresh_proxies` option: all the "proxy quircks" should be gone now
- [x] Migration with Rector https://github.com/zenstruck/foundry/pull/941
- [x] impact documentation https://github.com/zenstruck/foundry/pull/951
- [x] add configuration as a feature flag for this https://github.com/zenstruck/foundry/pull/950
- [x] Use PHP 8.4 proxies for usage of factories in data providers https://github.com/zenstruck/foundry/pull/910
- [x] use ghost objects for auto refresh mechanism https://github.com/zenstruck/foundry/pull/967
- [x] enable auto-refresh at factory level https://github.com/zenstruck/foundry/pull/970

I think all those need to be shipped in the same release

kbond:
> "undeprecate" auto_refresh_proxies configuration

This behaviour will be configurable? Would disabling just remove the events from the tracker? Making `refresh_all()` required? Or would it disable the system entirely?

> add a method to PersistentObjectFactory to create objects that are not auto-refreshable

Why?

nikophil:
>> "undeprecate" auto_refresh_proxies configuration
>
> This behaviour will be configurable? Would disabling just remove the events from the tracker? Making `refresh_all()` required? Or would it disable the system entirely?
>
> > add a method to PersistentObjectFactory to create objects that are not auto-refreshable
>
> Why?

[...]

PR #913, opened 2025-05-22
chore(1.x): fix rector set list for rector 2

nikophil:
fixes https://github.com/zenstruck/foundry/issues/784

I just wanted to make this rector set compatible with rector 2, because I'll use this code to create the future rector set we'll need for https://github.com/zenstruck/foundry/issues/899

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -238,7 +238,7 @@ jobs:
         shell: bash
 
       - name: Static analysis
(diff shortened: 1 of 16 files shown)


kbond (approved):
Is rector 1 still supported?

nikophil:
> Is rector 1 still supported?

no it's not, but I don't think it is a problem, do you?

By the way, Foundry 2 is finally becoming more popular than Foundry 1 :sweat_smile: 

image

kbond:
> but I don't think it is a problem, do you?

Nope, I was just curious

Issue #924, opened 2025-05-30
Uncaught TypeError: Zenstruck\Foundry\FactoryCollection::Zenstruck\Foundry\{closure}(): Argument #1 ($f) must be of type Zenstruck\Foundry\Factory null given


jdecool:
It seems the latest release [2.5.3](https://github.com/zenstruck/foundry/releases/tag/v2.5.3) introduces a bug.

When upgrading the library in one of our projects, I've this error when trying to load fixtures:

```bash
Fatal error: Uncaught TypeError: Zenstruck\Foundry\FactoryCollection::Zenstruck\Foundry\{closure}(): Argument #1 ($f) must be of type Zenstruck\Foundry\Factory, null given in /var/www/base/vendor/zenstruck/foundry/src/FactoryCollection.php:156
Stack trace:
#0 [internal function]: Zenstruck\Foundry\FactoryCollection::Zenstruck\Foundry\{closure}(NULL)
#1 /var/www/base/vendor/zenstruck/foundry/src/FactoryCollection.php(156): array_map(Object(Closure), Array)
#2 /var/www/core/src/Socle/Infrastructure/Doctrine/Fixtures/Story/AnnonceImmobiliere/AnnoncesDeBaseStory.php(84): Zenstruck\Foundry\FactoryCollection->create()
#3 /var/www/base/vendor/zenstruck/foundry/src/StoryRegistry.php(53): Core\Infrastructure\Doctrine\Fixtures\Story\AnnonceImmobiliere\AnnoncesDeBaseStory->build()
#4 /var/www/base/vendor/zenstruck/foundry/src/Story.php(112): Zenstruck\Foundry\StoryRegistry->load('Core\\Infrastruc...')
#5 /var/www/core/src/Shared/Fixtures/UserInterface/Console/ChargerStoriesDeBaseConsole.php(55): Zenstruck\Foundry\Story::load()
#6 /var/www/base/vendor/symfony/console/Command/Command.php(326): CoreShared\Fixtures\UserInterface\Console\ChargerStoriesDeBaseConsole->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#7 /var/www/base/vendor/symfony/console/Application.php(1096): Symfony\Component\Console\Command\Command->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
[...]

PR #925, opened 2025-05-30
Fix TypeError FactoryCollection::create when calling many with 0

jdecool:
Fix issue #924 

The issue is caused when calling the `FactoryCollection::many` with `0`.

This could happen when using it in a PHPUnit data provider.

### DIFF
--- a/src/FactoryCollection.php
+++ b/src/FactoryCollection.php
@@ -150,7 +150,10 @@ final class FactoryCollection implements \IteratorAggregate
             $lastFactory = \array_pop($factories);
             // @phpstan-ignore method.notFound (phpstan does not understand that we only have persistent factories here)
             $factories = \array_map(static fn(Factory $f) => $f->notRootFactory(), $factories);
-            $factories[] = $lastFactory;
+
+            if ($lastFactory !== null)  {
+                $factories[] = $lastFactory;
+            }
         }
 
         return \array_map(static fn(Factory $f) => $f->create($attributes), $factories);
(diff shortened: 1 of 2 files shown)


kbond (approved):
Awesome, thanks @jdecool!

nikophil:
yeah, thanks @jdecool 

I've released your fix in 2.5.4

I was wondering why PHPStan did not warned me about a potential null value. It actually knows that `array_pop()` can return null values: https://phpstan.org/r/4ef7d00f-c06b-4cb9-9eb1-62f36ea94732

This is because I'm ignoring a previous error, and then PHPStan, does not know anymore anything about the `$factories` var:
```php
        $factories = $this->all();

        \phpstan\dumpType($factories); // list<TFactory>

        if (Configuration::instance()->flushOnce && $this->isRootFactory && $this->factory instanceof PersistentObjectFactory && $this->factory->isPersisting()) {
            $lastFactory = \array_pop($factories);
            // @phpstan-ignore method.notFound (phpstan does not understand that we only have persistent factories here)
            $factories = \array_map(static fn(Factory $f) => $f->notRootFactory(), $factories);

            \phpstan\dumpType($factories); // list<*ERROR**>
```

Because, we're checking if `$this->factory instanceof PersistentObjectFactory` and somewhere else, we say that `$this->factory` is the generic type `TFactory`, We know we have a `FactoryCollection<PersistentObjectFactory>`, but PHPStan, is not aware of this, hence the `@phpstan-ignore` I've added :shrug:

jdecool:
Thanks @kbond and @nikophil

Issue #928, opened 2025-06-04
Unable to create collection of entities on v2.5.3+


mvhirsch:
I really love using Foundry. It makes my day wonderful. 🌞 

The issue results in an unpersisted relation of the very same object (`Episode` in my example) I just create. Prior to v2.5.3 this works; I was able to pin it down to #908 which broke it. I created a reproducer repository: https://github.com/mvhirsch/foundry-one-to-many-persist-reproducer

```
1) App\Tests\ReproducingTest::reproduce with data set #0 (App\Factory\EpisodeFactory Object (...))
Doctrine\ORM\ORMInvalidArgumentException: A new entity was found through the relationship 'App\Entity\Profile#episodes' that was not configured to cascade persist operations for entity: Quia reiciendis ipsa qui sint enim ea.. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}).

/path/to/project/mvhirsch/foundry-one-to-many-collection-persist/vendor/doctrine/orm/src/ORMInvalidArgumentException.php:103
/path/to/project/mvhirsch/foundry-one-to-many-collection-persist/vendor/doctrine/orm/src/UnitOfWork.php:3854
/path/to/project/mvhirsch/foundry-one-to-many-collection-persist/vendor/doctrine/orm/src/UnitOfWork.php:415
/path/to/project/mvhirsch/foundry-one-to-many-collection-persist/vendor/doctrine/orm/src/EntityManager.php:414
/path/to/project/mvhirsch/foundry-one-to-many-collection-persist/vendor/zenstruck/foundry/src/Persistence/PersistenceManager.php:139
/path/to/project/mvhirsch/foundry-one-to-many-collection-persist/vendor/zenstruck/foundry/src/Persistence/PersistenceManager.php:77
/path/to/project/mvhirsch/foundry-one-to-many-collection-persist/vendor/zenstruck/foundry/src/Persistence/PersistentObjectFactory.php:230
/path/to/project/mvhirsch/foundry-one-to-many-collection-persist/vendor/zenstruck/foundry/src/Persistence/PersistentProxyObjectFactory.php:41
/path/to/project/mvhirsch/foundry-one-to-many-collection-persist/vendor/zenstruck/foundry/src/FactoryCollection.php:159
[...]

PR #929, opened 2025-06-04
fix: remove useless Configuration::boted() check

nikophil:
fixes https://github.com/zenstruck/foundry/issues/928

ping @mvhirsch

:warning: I've created a 2.5.x branch, because I still need to perform some checks before releasing 2.6

### DIFF
--- a/src/Persistence/PersistentObjectFactory.php
+++ b/src/Persistence/PersistentObjectFactory.php
@@ -449,12 +449,6 @@ abstract class PersistentObjectFactory extends ObjectFactory
 
     final protected function initializeInternal(): static
     {
-        if (!Configuration::isBooted()) {
-            return $this;
-        }
-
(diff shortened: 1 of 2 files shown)


src/Persistence/PersistentObjectFactory.php
```diff
@@ -449,12 +449,6 @@ protected function normalizeObject(string $field, object $object): object
 
     final protected function initializeInternal(): static
     {
-        if (!Configuration::isBooted()) {
-            return $this;
-        }
-
-        $this->persist = $this->isPersistenceEnabled() ? PersistMode::PERSIST : PersistMode::WITHOUT_PERSISTING;
```
nikophil, line 456: this is a leftover from an old dev, now this is not needed. And the `Configuration::isBooted()` before was preventing to add the `afterInstanntiate` hook (in the case when the phpunit extension is not used)

PR #932, opened 2025-06-05
feat: add randomRangeOrCreate method

elliotbruneel:
Hello,

It would be helpful to have a `randomRangeOrCreate` method.

In some cases, we want to use randomRange without having to manually check if the required objects already exist or peristence is available.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -237,6 +237,7 @@ This command will generate a ``PostFactory`` class that looks like this:
          * @method static Post[]&Proxy[] createSequence(iterable|callable $sequence)
          * @method static Post[]|Proxy[] findBy(array $attributes)
          * @method static Post[]|Proxy[] randomRange(int $min, int $max, array $attributes = []))
+         * @method static Post[]|Proxy[] randomRangeOrCreate(int $min, int $max, array $attributes = [])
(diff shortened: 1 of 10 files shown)


docs/index.rst
nikophil, line 1: please add an example of usage under the section `Using your Factory`

tests/Integration/Persistence/GenericFactoryTestCase.php
```diff
@@ -286,6 +286,45 @@ public function random_or_create(): void
         static::factory()::repository()->assert()->count(2);
     }
 
+    /**
+     * @test
+     */
+    #[Test]
+    public function random_range_or_create(): void
```
nikophil: please, split this test in several test cases:
- can find enough objects without a criteria
- can find enough object with a criteria
- no objects matching the criteria => will create all the objects
- some objects matching the criteria, but not enough => will create some more objects

nikophil (changes requested):
Hi @elliotbruneel

some minor changes and we're good to go :ok_hand: 

thanks! 

(please rebase, the CI should be green again)

nikophil:
thanks @elliotbruneel  !

PR #933, opened 2025-06-05
minor: fix running `phpunit` w/o arguments

kbond:
I was getting a

> `./phpunit: line 78: @: unbound variable`

error when running `./phpunit` w/o args (with args it worked ok).

(This fix is an AI suggestion)

### DIFF
--- a/phpunit
+++ b/phpunit
@@ -75,7 +75,7 @@ case ${PHPUNIT_VERSION} in
     ;;
 esac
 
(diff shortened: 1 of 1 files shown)


nikophil:
that's strange I never had any problem with this...

out of curiosity which shell are you using? using a Mac ?

kbond:
> out of curiosity which shell are you using? using a Mac ?

Mac /w zsh

PR #941, opened 2025-06-15
feat: Rector rules to help migrating away from proxy

nikophil:
see https://github.com/zenstruck/foundry/issues/899#issuecomment-2915561115

This is a "best effort" stuff, I think everything cannot be removed automatically, but this should make the heavy lifting

Here is the full list of what does the "Foundry Rector set":
### RemoveWithoutAutorefreshCallRector
```php
// before 
$object = DummyPersistentProxyFactory::createOne();
$object->_withoutAutoRefresh(
    function($object) {
        $object->doSomething();
    }
);

// after
$object = DummyPersistentProxyFactory::createOne();
$object->doSomething();
```
---
### ChangeFactoryBaseClassRector
```php
// before
/**
 * @extends PersistentProxyObjectFactory<Foo>
 */
class SomeFactory extends PersistentProxyObjectFactory

//after
/**
 * @extends \Zenstruck\Foundry\Persistence\PersistentObjectFactory<Foo>
 */
class SomeFactory extends \Zenstruck\Foundry\Persistence\PersistentObjectFactory
```
---
### ChangeProxyParamTypesRector
```php
// before
/**
 * @param Proxy<Foo> $param
 */
public function someMethod(Proxy $param)
{

}

//after
/**
 * @param Foo $param
 */
public function someMethod(Foo $param)
{

}
```
---
### ChangeProxyReturnTypesRector
```php
// before
/**
 * @return Proxy<Foo>
 */
public function someMethod(): Proxy
{

}

//after
/**
 * @return Foo
 */
public function someMethod(): Foo
{

}
```
---
### RemovePhpDocProxyTypeHintRector
```php
// before
/** @var Proxy<Foo> $foo */
$foo = SomeFactory::createOne();

//after
/** @var Foo $foo */
$foo = SomeFactory::createOne();
```
---
### RemoveUnproxifyArrayMapRector
```php
// before
/**
 * @return Proxy<Foo>
 */
public function someMethod(): Proxy
{

}

//after
/**
 * @return Foo
 */
public function someMethod(): Foo
{

}
```
---
### RemovePhpDocProxyTypeHintRector
```php
// before
$objects = array_map(
    static fn (Proxy $proxy) => $proxy->_real(),
    [DummyObject::new()]
);

//after
$objects = [DummyObject::new()];
```
---
### RemoveFunctionCallRector (applies to `proxy()` and `unproxy()`)
```php
// before
$other = proxy($object);

//after
$other = $object;
```
---
### RemoveMethodCallRector (applies to `_enableAutoRefresh()`, `_disableAutoRefresh()` and `_real()`)
```php
// before
$object = $object->_save()->_enableAutoRefresh();

//after
$object = $object->_save();
```
---
### MethodCallToFuncCallWIthObjectAsFirstParameterRector (applies to `_get()`, `_set()`, `_save()`, `_refresh()`, `_delete()`, `_assertPersisted()` and `_assertNotPersisted()`)
```php
// before
$object->_set('a', 'b')->_save()->_assertPersisted('message');

//after
\Zenstruck\Foundry\Persistence\assert_persisted(\Zenstruck\Foundry\Persistence\save(\Zenstruck\Foundry\set($object, 'a', 'b')), 'message');
```


utils/rector/src/MethodCallToFuncCallWIthObjectASFirstParameter/MethodCallToFuncCallWIthObjectAsFirstParameterRector.php
```diff
@@ -0,0 +1,64 @@
...
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes() : array
+    {
+        return [MethodCall::class];
+    }
+    /**
+     * @param MethodCall $node
+     */
+    public function refactor(Node $node) : ?Node
```
TomasVotruba: I think it's this one, right?

I'll make notes as it comes to my mind:

* I'd avoid configuration and use values here instead. The fluent call has to be converted fully, to its split functoins.
* If you want to return multiple stmts = standalone lines with function calls, you need to hook into and `Expression` node. That way we're working with lines, and can return multiple lines.
* The `Expression` node must contain `$node->expr` of `MethodCall` instance, then you create functions and return them all.

First notes to kick off.

nikophil: First, thanks for your help :)

To give you some context, the [Proxy](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#object-proxy) mechanism in Foundry is clunky, and PHP 8.4 lazy objects gives us a way to drastically improve this. So I want to remove everything dealing with proxies. We have functions helper that can replace each method from `Proxy` class.

I think I'll create a dedicated rector for this case: the other fluent methods will be replaced with functions that actually return the object, so, we could mimic the fluent syntax with them.
eg: `$object->_save()->_save()` will become `save(save($object))`.

Sadly, the method `Proxy::_set()` is the more complex one to replace, and is one of the most used in the wild. If it wasn't the case, I would have let the user deal with edge cases by themselves: this Rector set is a "best effort" thing, they will have to fix stuff manually anyway... More over, not dealing with this specific method will lead to broken code after the Rector set has been run.

> If you want to return multiple stmts = standalone lines with function calls, you need to hook into and Expression node. That way we're working with lines, and can return multiple lines.

yes, that's what I finally understood :sweat_smile: but dealing with `Expression` is a lot more challenging.

if the method I want to remove is in the middle of the expression (eg: `$proxy->_save()->_set()->_save()` or `$someService->doSomething($proxy->_set()->save())`), I don't know how to deal with it. Is there some special "expression traverser" before I dive into a solution using recursion?

TomasVotruba, line 42: I think it's this one, right?

I'll make notes as it comes to my mind:

* I'd avoid configuration and use values here instead. The fluent call has to be converted fully, to its split functoins.
* If you want to return multiple stmts = standalone lines with function calls, you need to hook into and `Expression` node. That way we're working with lines, and can return multiple lines.
* The `Expression` node must contain `$node->expr` of `MethodCall` instance, then you create functions and return them all.

First notes to kick off.

TomasVotruba: Brief example:

```php
if (! $node->expr instanceof MethodCall) {
    return null;
}

$stmtsToReturn = [];

$currentMethodCall = $node->expr;
do { ... 
    // create func  call here
    $funcCall = new FuncCall(...);
    $stmtsToReturn[] = new Expression($funcCall);

    // jump to next nested method call
    $currentMethodCall = $currentMethodCall->var;	 
} while ($currentMethodCall instanceof MethodCall);



if ($stmtsToReturn === []) {
    return null;
}

return $stmtsToReturn;
```

nikophil: the code example you gave is pretty clear, thanks

How would you handle the case where you need to use a new variable, I know this pattern is widly used as well:
```diff
- SomeFactory::createOne()
-    ->_set('prop1', 'value1')
-    ->_save()
-    ->_set('prop2', 'value2')
-    ->_save()
-    ->_real();

+ $object = SomeFactory::createOne();
+ Zenstruck\Foundry\set($object, 'prop1', 'value1');
+ $object = $object->save();
+ Zenstruck\Foundry\set($object, 'prop2', 'value2');
+ $object->save()->_real();
```

is there some helper to generate unused var names?

TomasVotruba: > How would you handle the case where you need to use a new variable, 

If there is convention of naming factories by their entity, I'd simply use the prefix from `SomeFactory` → `Some` :) 

---

> if the method I want to remove is in the middle of the expression (eg: $proxy->_save()->_set()->_save() or $someService->doSomething($proxy->_set()->save())),

Removing node in the middle of the chain is mental hack. Nested calls work in inversed order. 

So:

```php
$some->first()->second()->third();
```

Is `third()`, `second()`, then `first()` in the AST tree. This might help:

https://getrector.com/ast/008d8f973cd69fa967a1aaaf5c8c04459fd3ea34


Removing a node somewhere in the middle is done like this:

```php
// here $node->var, is `MethodCall` of second()`

// to remove if, just assign it the node above (third())
$node->var  = $node;
```

nikophil: hey @TomasVotruba, sorry for the late feedback, I've been in holidays all this time :palm_tree: 

> Removing node in the middle of the chain is mental hack

yeah I just figured that out!

and despite your help on this, I still struggle with a lot of edge cases... 

I found another way to do this, and  changed the target function in order that it also returns the object passed in argument:
```diff
- function set(object $object, string $property, mixed $value): void
+ function set(object $object, string $property, mixed $value): object
{
    Hydrator::set($object, $property, $value);
+ 
+    return $object;
}
```
This way the transformation does not break the fluent syntax, and I can nest all those function calls:
```php
// before
$proxy->_set('field1', 'value1')->_set('field2', 'value')->_save();

// after
$proxy = \Zenstruck\Foundry\save(\Zenstruck\Foundry\save(\Zenstruck\Foundry\save($proxy, 'field1', 'value1'), 'field2', 'value2'));
```

much more easier to implement, and better for my mental health :laughing: 

Nevertheless, thanks for your help, I really appreciate!

kbond (approved):
🚀

PR #943, opened 2025-06-15
feat: auto refresh using PHP 8.4 ghost objects

nikophil:
relates to https://github.com/zenstruck/foundry/issues/899

I'll use this PR as a "meta" PR for all work around proxies in PHP 8.4 - this i why the branch is in `zenstruck/foundry` repo


dmitryuk:
Any plans to merge?
@nikophil can you resolve conflicts?

nikophil:
Hi @dmitryuk 

Thanks for your interest into this,  but it is still not ready

I'd like to finish the work on the rector stuff to ease the migration

See https://github.com/zenstruck/foundry/pull/941

And I have very little time for OSS these days

nikophil:
BTW, I'm curious: do you need this feature in some ways @dmitryuk ? :innocent:

dmitryuk:
Yes, to fix a lot of deprecations of doctrine/orm

nikophil:
could you elaborate, please?

you mean these deprecations? https://github.com/doctrine/orm/pull/12005/files

I think you already can enable the native lazy object stuff in Doctrine, and it will work in Foundry (as soon as you don't make your entity final/readonly)

dmitryuk:
Yes, native lazy objects are enabled and I receive deprecation below
image

nikophil:
yeah OK, it's a deprecation from `symfony/var-exporter`, not Doctrine

well, I'd really like to finish the migration path with Rector before releasing this PR - otherwise, resolving the deprecations from this PR will be a nightmare for big projects. I think it will be OK in one or two weeks.

In the meantime, I think your only solution would be to add this deprecation to the baseline

PR #946, opened 2025-07-13
doc: Use latest version of security to hash passwords

treztreiz:
Use the latest version of security command


nikophil:
thanks @treztreiz 

those failures are unrelated, I guess that doctrine just emitted a new deprecation that we need to take care of

PR #950, opened 2025-07-24
feat: auto refresh with lazy object php84 enabled by config

nikophil:


kbond (approved):

src/ZenstruckFoundryBundle.php
```diff
@@ -444,4 +455,15 @@ static function(ChildDefinition $definition, AsFixture $attribute, \ReflectionCl
             }
         );
     }
+
+    private function configureAutoRefreshWithLazyObjects(ContainerBuilder $container, bool $enableAutoRefreshWithLazyObjects): void
+    {
+        $container->setParameter('zenstruck_foundry.enable_auto_refresh_with_lazy_objects', $enableAutoRefreshWithLazyObjects);
+
+        if (!$enableAutoRefreshWithLazyObjects && \PHP_VERSION_ID >= 80400) {
```
kbond: Oh, I wonder if we should do this?

kbond (commented):
Questioning whether this should be an option that'll be available in 3.0 - just true by default

PR #952, opened 2025-07-28
refactor(maker): deprecate --with-phpdocs for PHP >=8.4

nikophil:


tests/Integration/Maker/MakeFactoryTest.php
```diff
@@ -34,6 +36,7 @@
  * @group maker
  */
 #[Group('maker')]
+#[RequiresPhp('>=8.4')]
```
nikophil, line 40: I skip those tests in other PHP versions, otherwise, we'd need to duplicate all the expected files, in order to only change the extended factory class

kbond (approved):

PR #954, opened 2025-07-28
feat: rename `foundry:load-stories` to `foundry:load-fixtures`

kbond:
Ref: https://x.com/loic_425/status/1930565067394727947

30aa0046111235df706b2fef28d3c1ed75927f50 fixes a problem where the aliases weren't being registered.

The main command is `foundry:load-fixtures` with the following aliases:
1. `foundry:load-stories` (the previous command name)
2. `foundry:load-story`
3. `foundry:load-fixture` (not really an alias but *just* works)

### DIFF
--- a/tests/Integration/Command/LoadFixturesCommandTest.php
+++ b/tests/Integration/Command/LoadFixturesCommandTest.php
@@ -30,7 +30,7 @@ use Zenstruck\Foundry\Tests\Integration\RequiresORM;
 
 use function Zenstruck\Foundry\Persistence\repository;
 
-final class LoadStoryCommandTest extends KernelTestCase
+final class LoadFixturesCommandTest extends KernelTestCase
 {
     use Factories, RequiresORM, ResetDatabase;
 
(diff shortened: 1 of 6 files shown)


tests/Integration/Command/LoadFixturesCommandTest.php
```diff
@@ -278,6 +278,16 @@ public function it_does_not_load_global_state(): void
...
+        // randomly choose the real command name or an alias
+        $commands = [
+            'foundry:load-fixtures',
+            'foundry:load-fixture',
+            'foundry:load-stories',
+            'foundry:load-story',
+        ];
+
+        return new CommandTester((new Application(self::bootKernel($options)))->find(
+            $commands[\array_rand($commands)]
```
kbond, line 290: 🙈 I didn't feel like a true data provider was needed here...

I get this isn't proper so I can change if desired. :)

nikophil: sometimes I like to use `#[TestWith]` for this kind or really simple data, but nevermind, I'm cool with this

nikophil (approved):

PR #956, opened 2025-07-30
feat: introduce method FactoryCollection::applyStateMethod()

nikophil:
this is another proposal which replaces https://github.com/zenstruck/foundry/pull/834

our factories at work are full of "state methods", and I want a way to be able to call a state method for each factory of a collection, but with different parameters

ping @smnandre :wink:


kbond (approved):
Looks great!

PR #959, opened 2025-08-05
fix: generated proxy code for methods with union or intersection return type

BackEndTea:
A method with a union or intersection return type would miss the `$this->_autoRefresh();` call, as it would be missed by the regex. A missing auto refresh causes the following error in combination with symfony /var-exporter 7.3.0 or higher.

```
Error: Typed property Symfony\Component\VarExporter\Internal\LazyObjectState::$realInstance must not be accessed before initialization
```

### DIFF
--- a/src/Persistence/ProxyGenerator.php
+++ b/src/Persistence/ProxyGenerator.php
@@ -129,11 +129,11 @@ final class ProxyGenerator
          * (?!__)                               # 4. Negative lookahead to exclude magic methods (like __serialize)
          * \w+                                  # 5. Method name
          * \s*\([^\)]*\)\s*                     # 6. Parameters inside parentheses (not captured)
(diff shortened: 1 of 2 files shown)


nikophil:
thanks @BackEndTea 

BTW all this complexity with proxies will soon be removed, see https://github.com/zenstruck/foundry/pull/943 (or at least deprecated along with PHP 8.4)

BackEndTea:
Thanks for the merge @nikophil!

And i understand this is being removed, but at least it fixes the current bug right away.

nikophil:
yeah I still expect those proxies to give me headaches for quite some time, this won't be removed from 2.x :sweat_smile: 

it will only be deprecated when using PHP 8.4.

Foundry 3 will bump PHP 8.4 and we will be able to remove this proxy mechanism, and replace it with a much more simple autorefresh mechanism which will leverage PHP 8.4 lazy objects.

nikophil:
by the way, I've released your fix

PR #962, opened 2025-08-06
fix: proxying of classes that have tenative return types

BackEndTea:
The symfony Proxy Helper will add a `ReturnTypeWillChange` attribute to methods with tenative return types.
The regex wasn't able to handle attributes on methods so it wouldn't add the `_autoRefresh`.
The 'best' example of this is jsonSerialize


nikophil:
TIL the term of "tenative return types" :sweat_smile: 

thanks @BackEndTea

BackEndTea:
Thanks for the quick merge @nikophil !

PR #964, opened 2025-08-06
bug: reproducer for not-proxy factory and uninitialized fields

nikophil:
:warning: I actually think this problem won't be resolved, because it comes from a bad design of the entities in userland and its fix would be either too complex or bad for performance

### DIFF
--- a/tests/Integration/ORM/GenericEntityFactoryTest.php
+++ b/tests/Integration/ORM/GenericEntityFactoryTest.php
@@ -11,6 +11,7 @@
 
 namespace Zenstruck\Foundry\Tests\Integration\ORM;
 
+use PHPUnit\Framework\Attributes\IgnoreDeprecations;
 use PHPUnit\Framework\Attributes\Test;
 use Zenstruck\Foundry\Tests\Fixture\Entity\EdgeCases\EntityWithUninitializedFields;
 use Zenstruck\Foundry\Tests\Fixture\Factories\Entity\EmptyConstructorFactory;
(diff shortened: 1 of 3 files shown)


tests/Integration/ORM/GenericEntityFactoryTest.php
```diff
@@ -55,6 +58,28 @@ public function can_use_factory_with_empty_constructor_without_persistence(): vo
...
+
+    /**
+     * @test
+     */
+    #[Test]
+    public function generate_entity_with_uninitialized_field_without_proxy(): void
+    {
+        $entity = persistent_factory(EntityWithUninitializedFields::class)->create();
+
+        self::assertNull($entity->getUninitializedField());
```
nikophil, line 83: :boom: 

> Error: Typed property Zenstruck\Foundry\Tests\Fixture\Entity\EdgeCases\EntityWithUninitializedFields::$uninitializedField must not be accessed before initialization

tests/Integration/ORM/GenericEntityFactoryTest.php
```diff
@@ -55,6 +58,28 @@ public function can_use_factory_with_empty_constructor_without_persistence(): vo
...
 
+    /**
+     * @test
+     */
+    #[Test]
+    public function generate_entity_with_uninitialized_field_with_proxy(): void
+    {
+        $entity = proxy_factory(EntityWithUninitializedFields::class)->create();
+
+        self::assertNull($entity->getUninitializedField());
```
nikophil, line 72: everything OK, because with the refresh, Doctrine correctly hydrates the property

nikophil:
I don't want to merge this fix (at least, not now), because it hurts performances, and it is a patch to prevent errors for bad designed entities

PR #967, opened 2025-08-25
feat: use ghost objects for auto refresh mechanism

nikophil:
Lazy ghost do not create identity problems, so they are more suited for our purpose.

I've also added a lot of tests, based on the different errors I had when testing the feature on a real life project

### DIFF
--- a/src/Object/Hydrator.php
+++ b/src/Object/Hydrator.php
@@ -141,6 +141,29 @@ final class Hydrator
         return self::accessibleProperty($object, $property)->getValue($object);
     }
 
+    /**
+     * @template T of object
+     *
+     * @param T $object
+     * @param T $other
+     */
+    public static function hydrateFromOtherObject(object $object, object $other): void
+    {
+        $classes = [$object::class, ...\array_values(\class_parents($object))];
+
+        $properties = [];
+        foreach ($classes as $class) {
+            $reflector = new \ReflectionClass($class);
+            foreach ($reflector->getProperties() as $property) {
+                $properties[$property->getName()] = $property->getName();
+            }
+        }
+
+        foreach ($properties as $property) {
+            self::set($object, $property, self::get($other, $property), catchErrors: true);
+        }
+    }
+
     private static function accessibleProperty(object $object, string $name): \ReflectionProperty
     {
         $class = new \ReflectionClass($object);
--- a/src/Persistence/PersistenceManager.php
+++ b/src/Persistence/PersistenceManager.php
@@ -16,8 +16,11 @@ use Doctrine\Persistence\ObjectManager;
 use Doctrine\Persistence\ObjectRepository;
 use Zenstruck\Foundry\Configuration;
 use Zenstruck\Foundry\Exception\PersistenceNotAvailable;
+use Zenstruck\Foundry\Object\Hydrator;
 use Zenstruck\Foundry\ORM\AbstractORMPersistenceStrategy;
 use Zenstruck\Foundry\Persistence\Exception\NoPersistenceStrategy;
+use Zenstruck\Foundry\Persistence\Exception\ObjectHasUnsavedChanges;
+use Zenstruck\Foundry\Persistence\Exception\ObjectNoLongerExist;
 use Zenstruck\Foundry\Persistence\Exception\RefreshObjectFailed;
 use Zenstruck\Foundry\Persistence\Relationship\RelationshipMetadata;
 use Zenstruck\Foundry\Persistence\ResetDatabase\ResetDatabaseManager;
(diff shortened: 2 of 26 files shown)


src/Persistence/PersistenceManager.php
```diff
@@ -140,14 +143,53 @@ public function flush(ObjectManager $om): void
...
      * @template T of object
      *
      * @param T $object
      *
      * @return T
+     *
+     * @throws RefreshObjectFailed
      */
-    public function refresh(object &$object, bool $force = false, bool $allowRefreshDeletedObject = false): object
+    public function refresh(object &$object, bool $force = false): object
```
nikophil, line 192: I've actually reset the `refresh()` method to the same state than in 2.x

src/Persistence/PersistenceManager.php
```diff
@@ -140,14 +143,53 @@ public function flush(ObjectManager $om): void
         }
     }
 
+    /**
+     * @template T of object
+     *
+     * @param T $object
+     */
+    public function autorefresh(object $object, object $clone): void
```
nikophil, line 151: I decided to split "refresh" and "autorefresh" mechanisms, because they does not involve the same problems, and the autorefresh should not manipulate the object reference (otherwise, we're exposed to identity problems)

src/Persistence/PersistenceManager.php
```diff
@@ -140,14 +143,53 @@ public function flush(ObjectManager $om): void
...
+            if ($refreshedObject = $om->find($object::class, $id)) {
+                Hydrator::hydrateFromOtherObject($object, $refreshedObject);
+
+                return;
+            }
+        }
+
+        // the object no longer exists
+        Hydrator::hydrateFromOtherObject($object, $clone);
+        $om->detach($object);
```
nikophil, line 180: if the object no longer exist, we still need to return something, otherwise, the lazy ghost will be re-initialized as an empty shell. For this purpose we're using a clone of the object

src/Persistence/PersistenceManager.php
```diff
@@ -140,14 +143,53 @@ public function flush(ObjectManager $om): void
...
+     * @template T of object
+     *
+     * @param T $object
+     */
+    public function autorefresh(object $object, object $clone): void
+    {
+        $strategy = $this->strategyFor($object::class);
+        $om = $strategy->objectManagerFor($object::class);
+
+        $id = $strategy->getIdentifierValues($clone);
```
nikophil, line 156: we need the clone here, because it still has is "identifier fields" hydrated

tests/Integration/Mongo/AutoRefreshTest.php
```diff
@@ -14,17 +14,42 @@
...
+        $objectId = $object->id;
+
+        self::getContainer()->get('services_resetter')->reset(); // @phpstan-ignore method.notFound
+        self::assertTrue((new \ReflectionClass($object))->isUninitializedLazyObject($object));
+
+        $this->updateObject($objectId);
+
+        self::assertSame('foo', $object->getProp1());
+
+        self::assertTrue($this->objectManager()->contains($object));
```
nikophil, line 50: there is exactly the same test in ORM version, but because the document manager is not cleared when `kernel.reset` is triggered, this assertions is not the same

kbond (approved):
Awesome! I'm glad using lazy ghosts solves the identity problems.

Question about auto-refreshing:

```php
$user = UserFactory::createOne();

// some code that manipulates the user

$this->assertSame('updated', $user->getUsername()); // auto-refresh is triggered

// some additional code that manipulates the user

$this->assertSame('updatedagain', $user->getUsername()); // is auto-refresh still triggered?
```

nikophil:
@kbond if by "code that manipulates the user", you mean a call with browser and/or console command and/or messenger, the answer is yes

I think I need to add a test for this

src/Object/Hydrator.php
```diff
@@ -141,6 +141,29 @@ public static function get(object $object, string $property): mixed
...
+            $reflector = new \ReflectionClass($class);
+            foreach ($reflector->getProperties() as $property) {
+                $properties[$property->getName()] = $property->getName();
+            }
+        }
+
+        foreach ($properties as $property) {
+            self::set($object, $property, self::get($other, $property), catchErrors: true);
+        }
+    }
```
nikophil, line 165: I hope this code won't break too much things...

actually if it breaks, it means that the entity/object has a design problem :sweat_smile:

src/Persistence/PersistenceManager.php
```diff
@@ -140,14 +143,53 @@ public function flush(ObjectManager $om): void
...
+
+        if ($id) {
+            try {
+                $om->refresh($object);
+
+                if ($strategy->getIdentifierValues($object)) {
+                    // no identifier values means the object no longer exists
+                    return;
+                }
+            } catch (\Throwable $e) {
```
nikophil, line 166: I really don't like to do this, but, this was the only way to have something working in all situations, without being to tight with Doctrine

nikophil:
@kbond I've added the test I was mentioning... I'm merging this PR :tada:

PR #968, opened 2025-08-25
feat(2.9): deprecate `Factories` trait and force PHPUnit extension usage

### LINKED ISSUE #965: Remove usage of `debug_backtrace()` to check presence of `Factories` trait
`debug_backtrace()` is known to harm performances.... see https://github.com/zenstruck/foundry/blob/2d28e67fb52673fac92f02e9b3d8ba61cf932a29/src/Exception/FactoriesTraitNotUsed.php#L36

nikophil:
Removes the usage of `Factories` trait in favor of the PHPUnit extension.

This will be really less error prone, since it would be impossible to omit it, and it would be added once for all 

fixes https://github.com/zenstruck/foundry/issues/965 (because we won't check anymore if the trait is used)

I had hard times (like REALLY hard times) to make the feature "use `PersistentObjectFactory::create()` method in data providers" work without the trait. The code is kinda hacky, but since we cannot access the test instance from the PHPUnit extension (even with some hacky code, this is literally impossible), I had to be creative :sweat_smile:


src/PHPUnit/KernelTestCaseHelper.php
```diff
@@ -0,0 +1,14 @@
...
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\DependencyInjection\Container;
+
+final class KernelTestCaseHelper extends KernelTestCase
+{
+    public static function getContainer(): Container
+    {
+        return parent::getContainer();
+    }
+}
```
nikophil: I'm not proud of this :see_no_evil:  :sweat_smile:

src/Test/Factories.php
```diff
@@ -27,58 +29,31 @@ trait Factories
...
      * @before
      */
-    #[Before]
+    #[Before(10)]
     public function _beforeHook(): void
     {
-        $this->_bootFoundry();
-        $this->_loadDataProvidedProxies();
-    }
+        $this->_loadDataProvidedProxies(); // todo remove
```
nikophil: This is the last missing piece before being able to drop the trait, but I have a solution

VincentLanglet (commented):
This is a great idea, we often forgot to add the Factories trait.

nikophil:
> This is a great idea, we often forgot to add the Factories trait.

thanks @VincentLanglet I also think this is where we should go.

actually, my main concern about this PR is that I had very hard time to recreate the feature where we can call `PersistentObjectFactory::create()` in data providers using the PHPUnit extension, which results in a very hacky code https://github.com/zenstruck/foundry/pull/968/files#diff-7dfab5e24ed550db77184706c3251c47f6740c6921c221dd9dcd017cd77f0b11

And I don't have a big real life project using Foundry these days, where I could have tested it

src/Persistence/ProxyGenerator.php
```diff
@@ -70,14 +71,18 @@ public static function wrapFactory(PersistentObjectFactory $factory): object
             throw new \LogicException('Native proxy generation requires PHP 8.4 or higher.');
         }
 
-        $reflector = new \ReflectionClass($factory::class());
-
-        return $reflector->newLazyProxy(static function() use ($factory) {
+        return (new \ReflectionClass($factory::class()))->newLazyGhost(static function(object $ghost) use ($factory): void {
```
nikophil, line 74: we must use "lazy ghost" objects, otherwise we'd have identity errors (see `DataProviderWithPersistentFactoryTestCase::assert_it_provided_instance_is_the_same_than_returned_by_repository()` which would not pass otherwise

src/Persistence/ProxyGenerator.php
```diff
@@ -70,14 +71,18 @@ public static function wrapFactory(PersistentObjectFactory $factory): object
...
             if (Configuration::instance()->inADataProvider() && $factory->isPersisting()) {
                 throw new \LogicException('Cannot access to a persisted object inside a data provider.');
             }
 
-            return $factory->create();
+            $createdObject = $factory->withoutPersisting()->create();
+
+            Hydrator::hydrateFromOtherObject($ghost, $createdObject);
```
nikophil, line 81: this would break with readonly properties, this can be fixed using `ReflectionProperty::setRawValueWithoutLazyInitialization()`. Will be done in another PR

src/Persistence/ProxyGenerator.php
```diff
@@ -70,14 +71,18 @@ public static function wrapFactory(PersistentObjectFactory $factory): object
...
+
+            if ($factory->isPersisting()) {
+                Configuration::instance()->persistence()->save($ghost);
+            }
```
nikophil, line 85: this means after persist callbacks are not used, this will also be fixed a further PR

nikophil:
OK then, I think I have found the solution I was looking for! I've prepared a branch `2.9.x` for this

PR #972, opened 2025-08-27
fix: misc fixes when creating objects in data provider

### LINKED ISSUE #971: Segmentation fault in dataprovider with PersistentProxyObjectFactory
Hi @nikophil, I saw you made https://github.com/zenstruck/foundry/pull/648 and I think I have an edge case.

I'm using foundry and phpunit, I have added the extension
```
<bootstrap class="Zenstruck\Foundry\PHPUnit\FoundryExtension" />
```

When using
```
DeploymentFactory::createOne()->getName();
```
in a Dataprovider of a TestCase (I didn't try a WebTestCase, my use case is a unit test ATM) I'm getting a segmentation fault.

This occurs with the following factory
```
final class DeploymentFactory extends PersistentProxyObjectFactory
{
    {
        return Deployment::class;
    }

    /**
     * @return array<string, mixed>
     */
    {
        $countryCode = self::faker()->countryCode();

        return [
            'id' => $countryCode,
            'name' => self::faker()->word(),
            'engineDomain' => 'api.'.$countryCode.'.weglot.com',
            'frontDomain' => 'app.'.$countryCode.'.weglot.com',
        ];
    }
}
```

As soon as I change the `PersistentProxyObjectFactory` into a `PersistentObjectFactory` the issue is solved.

nikophil:
fixes #971 

I also enabled the possibility to access objects in data provider even in `KernelTestCase`, if the factories are not persisting


nikophil:
@vincentLanglet I think this fixes your problem, would you mind to test the fix? Since we did not have the same error, I'm not 100% certain it fixes your problem

VincentLanglet:
> @VincentLanglet I think this fixes your problem, would you mind to test the fix? Since we did not have the same error, I'm not 100% certain it fixes your problem

This indeed solve my segmentation fault issue @nikophil.

I had to update my code which was doing `$entity1 === $entity2` into `$entity1->getId() === $entity2->getId()` in order to have a CI green, but I assume it might be a consequences of the change from `PersistentObjectFactory` to `PersistentProxyObjectFactory` (?)

nikophil:
Yes I think so, the proxy mechanism involves some identity problems 🫤

BTW, we're planning to deprecate it, and remove it in Foundry 3

VincentLanglet:
> BTW, we're planning to deprecate it, and remove it in Foundry 3

Oh, thanks for the news ; I will maybe stick with the PersistentObjectFactory then ^^'

nikophil:
The fix is released.

> I will maybe stick with the PersistentObjectFactory then

yeah that's what I'm already doing :sweat_smile: 

the main feature of those proxies being the "auto-refresh" behavior, we'll replace the whole proxy mechanism by an auto-refresh mechanism relying on PHP 8.4 lazy objects

PR #977, opened 2025-09-18
Fix readme and gitattributes

dmitryuk:
1. exclude: `phpunit-deprecation-baseline.xml`, `phpunit-rector.xml.dist`, `phpbench.json`, `docs folder` from vendor
2. Update readme for updating to 2.7 since where is no `enable_auto_refresh_with_lazy_objects` config option


UPGRADE-2.7.md
```diff
@@ -12,15 +12,7 @@ and to reduce the maintenance burden of this feature.
 > The new auto-refresh mechanism only applies for PHP 8.4,
 > if you're still not using PHP 8.4, there is nothing to do (yet!)
 
-First, you need to configure whether you want to use the new auto-refresh mechanism:
-
-```yaml
-zenstruck_foundry:
-    # from Foundry 2.7, with PHP >=8.4, not setting this configuration is deprecated
-    enable_auto_refresh_with_lazy_objects: true
-```
```
nikophil, line 21: not sure to understand why this has been removed?

we want the users to enable this option. Migrating away from the old proxy mechanism would be smoother

nikophil: > since where is no enable_auto_refresh_with_lazy_objects config option

:thinking: 

https://github.com/zenstruck/foundry/blob/2.x/src/ZenstruckFoundryBundle.php#L54

PR #980, opened 2025-09-21
fix(autorefresh): don't use clone to get the id values

### LINKED ISSUE #963: Add test to ensure WebTestCase does not trigger error
I recently had a problem with `WebTestCase`, while I was testing https://github.com/zenstruck/foundry/pull/943 on a real life project

I needed to make this commit : https://github.com/zenstruck/foundry/pull/943/commits/9425912ee45e499aad2059ff087f2f3709af13d9

otherwise I had the infamous error:
> LogicException: Booting the kernel before calling "Symfony\Bundle\FrameworkBundle\Test\WebTestCase::createClient()" is not supported, the kernel should only be booted once.

The scenario is the following:
- a test with the kernel runs
- `tearDown()` calls `ensureKernelShutdown()`
- AFTER THAT `Factories::_shutdownFoundry()` is called
- calling `Configuration::instance()` boots the kernel
- if the next test is a `WebTestCase`, it will throw

I'd like to do two things:
- now that [priority for hook methods](https://github.com/sebastianbergmann/phpunit/pull/5890) are a thing, we should add priorities to our hooks
- add a test which would prevent us if something like this happen in the future

### LINKED ISSUE #976: `clone` operation executed when `enable_auto_refresh_with_lazy_objects` is enabled, breaking custom `__clone()` methods
When `enable_auto_refresh_with_lazy_objects` is set to true, the https://github.com/zenstruck/foundry/blob/883cbb424e9a7f96e85f638ab3a0f2cc158aec95/src/Persistence/Proxy/PersistedObjectsTracker.php#L65 in the PersistedObjectsTracker::proxifyObjects() method performs a clone operation on objects before setting up lazy ghosts. 

This causes issues for entities with custom __clone() methods that modify object state (e.g., resetting IDs).

I am guessing that with lazy objects enabled, the proxy mechanism should be bypassed entirely, and no cloning should occur during the refresh process? 

*Steps to Reproduce*
- Use the version https://github.com/zenstruck/foundry/releases/tag/v2.7.0
- Enable `enable_auto_refresh_with_lazy_objects: true` in the foundry config
- Create an entity with a custom __clone() method that resets the ID
- Run tests that use foundry factories
- Tests fail because entity IDs are unexpectedly reset during refresh

### LINKED ISSUE #979: Identifiers broken between requests after updating 2.7
example code
```
$user        = UserFactory::createOne();
$id = $user->getId(); // primary identifier
$this->client->request('/any method that access user inside/' . $user->getId());

// Assert failed
self::assertEquals($id, $user->getId());
```

nikophil:
Fixing this was trickier than expected 😅

Since we cannot get the id values of the entity from the clone, we need to collect them at a place where we have access to the real object.

The first guess was to call Configuration::instance()->persistence()->getIdentifierValues($object) in PersistedObjectsTracker::proxifyObjects(). But this was messing with the container, during the kernel.reset phase. The problem is more or less the same than explained in https://github.com/zenstruck/foundry/issues/963 - BTW, I've fixed this issue, and created the test mentioned there. PHPUnit is missing an event before the tear down methods are called, so I didn't have a clean solution with this idea.

Second guess: let's collect the ids when we add the objects to the PersistedObjectsTracker! Problem: this is made in a "post instantiate" hook, so the ids are not generated at this time.

Third guess would be to make it in a "post persist" hook. But the problem here is that we perform an additional flush() when a post persist hook is used, this will have a performance impact.

The last solution I had was to collect the id values after each PersistManager::save(). This solution is not very satisfactory because now, the PersistenceManager knows about the PersistedObjectsTracker, but it is the only solution I found.

fixes https://github.com/zenstruck/foundry/issues/976
fixes https://github.com/zenstruck/foundry/issues/963
fixes https://github.com/zenstruck/foundry/issues/979

### DIFF
--- a/src/Persistence/Proxy/PersistedObjectsTracker.php
+++ b/src/Persistence/Proxy/PersistedObjectsTracker.php
@@ -23,7 +23,17 @@ final class PersistedObjectsTracker
      *
      * @var list<\WeakReference<object>>
      */
-    private static $buffer = [];
+    private static array $buffer = [];
+
+    /**
+     * @var \WeakMap<object, mixed>
+     */
+    private static \WeakMap $ids;
+
+    public function __construct()
+    {
+        self::$ids ??= new \WeakMap();
+    }
 
     public function refresh(): void
     {
@@ -34,12 +44,33 @@ final class PersistedObjectsTracker
     {
         foreach ($objects as $object) {
             self::$buffer[] = \WeakReference::create($object);
+
+            $id = Configuration::instance()->persistence()->getIdentifierValues($object);
+            if ($id) {
+                self::$ids[$object] = $id;
+            }
+        }
+    }
+
+    public static function updateIds(): void
+    {
+        foreach (self::$buffer as $reference) {
+            if (null === $object = $reference->get()) {
+                continue;
+            }
+
+            if (self::$ids->offsetExists($object)) {
+                continue;
+            }
+
+            self::$ids[$object] = Configuration::instance()->persistence()->getIdentifierValues($object);
         }
     }
 
     public static function reset(): void
     {
         self::$buffer = [];
+        self::$ids = new \WeakMap();
     }
 
     public static function countObjects(): int
@@ -51,6 +82,10 @@ final class PersistedObjectsTracker
 
     private static function proxifyObjects(): void
     {
+        if (!Configuration::isBooted()) {
+            return;
+        }
+
         self::$buffer = \array_values(
             \array_map(
                 static function(\WeakReference $weakRef) {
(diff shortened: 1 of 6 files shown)


src/Persistence/Proxy/PersistedObjectsTracker.php
```diff
@@ -63,8 +67,9 @@ static function(\WeakReference $weakRef) {
                     }
 
                     $clone = clone $object;
-                    $reflector->resetAsLazyGhost($object, function($object) use ($clone) {
-                        Configuration::instance()->persistence()->autorefresh($object, $clone);
+                    $id = Configuration::instance(checkFactoriesTrait: false)->persistence()->getIdentifierValues($object);
```
nikophil, line 70: We need to call `Configuration::instance()->persistence()->getIdentifierValues($object)` in `PersistedObjectsTracker::proxifyObjects()`. But this was messing with the container, during the `kernel.reset` phase. The problem is more or less the same than explained in [this issue](https://github.com/zenstruck/foundry/issues/963) - BTW, I've fixed this issue, and created the test mentioned there. 

Hence the new parameter `bool $checkFactoriesTrait` in `Configuratoin::instance()`

any way, this check will be removed in Foundry 3

PR #981, opened 2025-09-21
fix(autorefresh): return fresh data from RepositoryDecorator methods

### LINKED ISSUE #975: Factory::all() loads outdated entities
Hi!

I'm in the process to upgrade to the new 2.7 version with native proxies and has a weird behavior/maybe bug.
I've changed everything written in the upgrade doc, and checking my breaking test 1by1m and found this:

Minified example:

```php
public function testUpdateStatus(): void
    {
        $data = MyFactory::new()
            ->validStatus() // status == Status::VALID
            ->many(10)
            ->create();
        $idList = array_map(static fn ($item): int => $item->getId(), $data);

        static::getContainer()
            ->get(MyRepository::class)
            ->updateStatus($idList, Status::SENDING); // doctrine query that updates the status

        $rows = MyFactory::all(); // reload all from the db... or it used to be

        foreach ($rows as $row) {
            self::assertSame(Status::SENDING, $row->getStatus()); // fails and show everything is Status::VALID
        }
    }
```

if I put a refresh_all(); call between the Myfactory::all() and foreach, it works well. And the upgrade mentions it at the end to call refres (https://github.com/zenstruck/foundry/blob/2.x/UPGRADE-2.7.md#troubleshooting) But here I load everything from repo, why should I refresh right after?

Thanks for the help, awesome lib

nikophil:
fixes https://github.com/zenstruck/foundry/issues/975

The problem did exist before autorefresh with lazy objects, but we just never noticed it, and it was hidden by proxy mechanism :shrug: 

The root problem here is that `EntityRepository::findBy()` (and all other related methods like `findAll()`, etc..) do not actually return a fresh object if the object is already managed.

What's surprising is that Doctrine actually performs the query to the db, but it just don't hydrate the object, unless a hint is passed. This optimization is just here to prevent hydration. The behavior is the same [in ORM](https://github.com/doctrine/orm/blob/3.5.x/src/UnitOfWork.php#L2400) and [in ODM](https://github.com/doctrine/mongodb-odm/blob/2.13.x/lib/Doctrine/ODM/MongoDB/UnitOfWork.php#L2799) 

This is annoying, because hints can only be passed if a query builder is used. So we cannot use anymore `ObjectRepository::findBy()` and other nice helper methods

### DIFF
--- a/src/Persistence/Proxy/PersistedObjectsTracker.php
+++ b/src/Persistence/Proxy/PersistedObjectsTracker.php
@@ -21,9 +21,14 @@ final class PersistedObjectsTracker
     /**
      * This buffer of objects needs to be static to be kept between two kernel.reset events.
      *
-     * @var list<\WeakReference<object>>
+     * @var \WeakMap<object, bool>
      */
-    private static $buffer = [];
+    private static \WeakMap $buffer;
+
+    public function __construct()
+    {
+        self::$buffer ??= new \WeakMap();
+    }
 
     public function refresh(): void
     {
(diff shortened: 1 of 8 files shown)


src/Persistence/Proxy/PersistedObjectsTracker.php
```diff
@@ -21,9 +21,14 @@ final class PersistedObjectsTracker
     /**
      * This buffer of objects needs to be static to be kept between two kernel.reset events.
      *
-     * @var list<\WeakReference<object>>
+     * @var \WeakMap<object, bool>
```
nikophil, line 24: finally I'm using a weak map :smile: 

this prevents to have duplicates in the buffer

nikophil:
I made a mistake and create the PR without fork. I'm closing this one in favor of https://github.com/zenstruck/foundry/pull/983

PR #985, opened 2025-09-23
feat(2.9): deprecate `ResetDatabase` trait

nikophil:


src/ZenstruckFoundryBundle.php
```diff
@@ -37,7 +38,7 @@ final class ZenstruckFoundryBundle extends AbstractBundle implements CompilerPas
 {
     public function boot(): void
     {
-        if ($this->container) {
+        if ($this->container && (!Configuration::isBooted() || !FoundryExtension::isEnabled())) {
```
nikophil: @HypeMC, just for you to know:

now we're deprecating the usage of `ResetDatabase` trait in favor of an attribute + usage of the PHPUnit extension, I had errors `ServiceNotFoundException: The "kernel" service is synthetic, it needs to be set at boot time before it can be used.` in the database reset testsuite.

Because this currently needs to work with and without the `ResetDatabase` trait, the only solution I found was to check if the extension is enabled.

Don't hesitate if you have another idea. Anyway, this ugly stuff will be removed in Foundry 3

HypeMC: @nikophil Hello, thank you for the heads-up. I've tested this branch, and unfortunately, this change seems to bring back the original issue. I haven't had time to look into it in detail yet, but I'll try to figure it out.

EDIT: Switching to the `#[ResetDatabase]` attribute does solve the problem.

HypeMC: The reason the `EarlyBootedKernelWithTraitsTest` passes is that it's no longer the first test, so `ResetDatabaseManager::hasDatabaseBeenReset` is `true`.
It needs to be `false` for the test to fail, e.g. try running it individually:

```
./phpunit tests/Integration/ResetDatabase/EarlyBootedKernelWithTraitsTest.php
```

Perhaps the `before()` method could be improved:

```php
#[BeforeClass(10)]
public static function before(): void
{
    $ref = new \ReflectionProperty(ResetDatabaseManager::class, 'hasDatabaseBeenReset');
    $ref->setValue(false);

    self::bootKernel();
}
```

nikophil: @HypeMC I'm not sure what's the best way to fix this: we would need a marker to tell "reset database trait was used" to decide whether to check `Configuration::isBooted()` in `ZenstruckFoundryBundle::boot()` method

this becomes overly complex for such a niche bug, moreover that the problem will only occur when using the newly deprecated trait `ResetDatabase`

I think it will stay this way, unless you have a good idea

PR #989, opened 2025-09-25
fix: handle readonly when refreshing from repository decorator

### LINKED ISSUE #994: Errors with readonly properties after upgrading from `2.7.0` to `2.7.2`
Most of my entities have a readonly property `uuid` which is generated in the constructor of the entity

```php
interface UuidableInterface
{
    public Uuid $uuid {
        get;
    }
}

class User implements UuidableInterface
{
    #[Orm\Column(type: UuidType::NAME, length: 36, unique: true)]
    public readonly Uuid $uuid;
    
    public function __construct()
    {
        $this->uuid = Uuid::v4();
        // ...
    }
}
```

After upgrading from `2.7.0` to `2.7.2`, I get the following error in my tests:

```
LogicException: Attempting to change readonly property App\User\Domain\Entity\User::$uuid.
```

This seems to be trigger when calling repository methods from a factory: 

```php
UserFactory::repository()->findOneBy(['email' => 'owner@example.com']);
UserFactory::repository()->first();
// etc...
```

Could be related with #990 and #992 

Things were working normal with `2.7.0`

nikophil:
short story: fixes a bug with reaonly that https://github.com/zenstruck/foundry/pull/983/ introduced which @smnandre reported.

long story: the [original issue](https://github.com/zenstruck/foundry/issues/975) stated that a simple `MyFactory::all()` returns outdated entities since auto refresh with lazy objects is released.

Actually, `MyFactory::all()` (and all its siblings from `RepositoryDecorator`) always return outdated object if the object is already managed. That's because Doctrine does not refresh an already managed entity, [unless the hint `HINT_REFRESH` is passed](https://github.com/doctrine/orm/blob/3.5.x/src/UnitOfWork.php#L2400).

With the proxy autrefresh mechanism, the problem was hidden:
```php
$myObjects = MyFactory::createMany(10, ['status' => Status::DRAFT]);

// [somehow change the status of those objects]

foreach (MyFactory::all() as $row) {
    self::assertSame(Status::UPDATED, $row->getStatus());
}
```
because `$row->getStatus()` was performing a `refresh()` (and a SQL query for EACH ITEM)

But with the hint now, we hit a [known doctrine problem](https://github.com/doctrine/orm/issues/9505) which we was [already handling](https://github.com/zenstruck/foundry/blob/2.x/src/Persistence/PersistenceManager.php#L229) in another place.

I made the decision to perform an extra query when we hit the readonly problem. This is not ideal, but we cannot know in advance whether we should pass the hint or not. 

fixes #994

### DIFF
--- a/src/Persistence/RepositoryDecorator.php
+++ b/src/Persistence/RepositoryDecorator.php
@@ -136,7 +136,14 @@ class RepositoryDecorator implements ObjectRepository, \IteratorAggregate, \Coun
         if ($this->inMemory) {
             $results = $this->inner()->findBy($this->normalize($criteria), $orderBy, $limit, $offset);
         } else {
-            $results = Configuration::instance()->persistence()->findBy($this->class, $this->normalize($criteria), $orderBy, $limit, $offset);
+            try {
+                $results = Configuration::instance()->persistence()->findBy($this->class, $this->normalize($criteria), $orderBy, $limit, $offset);
+            } catch (\LogicException|\Error) {
+                // prevent entities/documents with readonly properties to create an error
+                // LogicException is for ORM / Error is for ODM
+                // @see https://github.com/doctrine/orm/issues/9505
+                $results = $this->inner()->findBy($this->normalize($criteria), $orderBy, $limit, $offset);
+            }
         }
 
         $objects = \array_values($results);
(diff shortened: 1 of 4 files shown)


smnandre:
Tested locally / on the project...  and the failing tests are now green ("super green") !    ⚡  🙇

PR #991, opened 2025-09-26
Fix: Add defensive check to RepositoryDecorator::random() method

sofwar:
- Add safety check to ensure findBy query returns results before accessing first element
- Prevent potential runtime errors in edge cases with race conditions
- Maintain consistent NotEnoughObjects exception handling
- No breaking changes to existing API or behavior

This enhancement improves the robustness of the random method by adding a defensive check that verifies the actual query results, not just the count, before attempting to access the first element. This prevents potential runtime errors in scenarios where count() might indicate objects exist but the actual query returns no results.


nikophil:
totally legit,

thanks @sofwar

PR #993, opened 2025-09-27
Fix edge case with Doctrine Middleware & early kernel boot

HypeMC:
This fixes an edge case that occurs when Doctrine requires the `kernel` service (or any other synthetic one) and the kernel is booted before `ResetDatabase::_resetDatabaseBeforeFirstTest()`.

Error example:

> 1) Zenstruck\Foundry\Tests\Integration\ResetDatabase\EarlyBootedKernelTest
> Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The "kernel" service is synthetic, it needs to be set at boot time before it can be used.

This happens because when [`ResetDatabaseManager` calls `self::bootKernel()`](https://github.com/zenstruck/foundry/blob/2c84cb0c2e09679562cd33e90e66559dae014d7b/src/Test/ResetDatabase.php#L36), the old [kernel is shut down](https://github.com/symfony/symfony/blob/944969c20417aa6421b555c08833540d368f3752/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L72), but a reference to its container (which no longer has the `kernel` service) remains. The new container instance is never used because of `Configuration::isBooted()`.

### DIFF
--- a/src/ZenstruckFoundryBundle.php
+++ b/src/ZenstruckFoundryBundle.php
@@ -37,7 +37,7 @@ final class ZenstruckFoundryBundle extends AbstractBundle implements CompilerPas
 {
     public function boot(): void
     {
-        if ($this->container && !Configuration::isBooted()) {
+        if ($this->container) {
             Configuration::boot($this->container->get('.zenstruck_foundry.configuration')); // @phpstan-ignore argument.type
         }
     }
--- a/tests/Fixture/ResetDatabase/DoctrineMiddleware.php
+++ b/tests/Fixture/ResetDatabase/DoctrineMiddleware.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the zenstruck/foundry package.
+ *
+ * (c) Kevin Bond <kevinbond@gmail.com>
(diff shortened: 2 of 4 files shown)


nikophil (approved):

nikophil:
thanks @HypeMC 

out of curiosity, in which "real world" scenario did you have this edge case?

nikophil:
released in [2.7.3](https://github.com/zenstruck/foundry/releases/tag/v2.7.3)

HypeMC:
> out of curiosity, in which "real world" scenario did you have this edge case?

@nikophil Hi, sorry for the late reply.

My real world example is a combination of several different features.
The first is [a library that has a Doctrine middleware](https://github.com/sofascore/purgatory-bundle) which requires Symfony's router to generate URLs when a transaction is committed.

The second is a router decorator that uses Symfony's translator to generate translated routes. The router caches certain things, so it requires access to the kernel to get the cache directory and environment.

If you have a test that boots the kernel in a `setUpBeforeClass()` method and that test runs first, you get the error. It was not visible for a long time until I started testing ParaTest. That is when I came across this issue.

PR #999, opened 2025-10-08
Fix using `refresh_all()` with `flush_after()`

HypeMC:
Currently `refresh_all()` doesn't work with `flush_after()` because the ids never get updated.

### DIFF
--- a/src/Persistence/PersistenceManager.php
+++ b/src/Persistence/PersistenceManager.php
@@ -426,6 +426,7 @@ final class PersistenceManager
                 $this->flush($om);
             }
         }
+        PersistedObjectsTracker::updateIds();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Persistence/PersistenceManager.php
```diff
@@ -426,6 +426,7 @@ private function flushAllStrategies(): void
                 $this->flush($om);
             }
         }
+        PersistedObjectsTracker::updateIds();
```
nikophil, line 429: Juste for you to know, in a further PR, we plan to remove this method, in favor of future so-called "foundry hooks"

see https://github.com/zenstruck/foundry/pull/986

I think it will work out of the box then. But we're not ready to merge it, so let's release your fix now 👍

nikophil (approved):
thank you @HypeMC 

those bug fixes are very much appreciated! 😊

src/Persistence/PersistenceManager.php
HypeMC: Great, thanks for the info.

HypeMC:
@nikophil Thank you for the reviews/merges, they are appreciated as well :smile:

PR #1000, opened 2025-10-09
docs: add a note about using `make:factory --test`

### LINKED ISSUE #786: [make:factory] Problems with `--test` flag

ttskch:
Closes #786

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -181,20 +181,20 @@ This command will generate a ``PostFactory`` class that looks like this:
 
     Using ``make:factory --test`` will generate the factory in ``tests/Factory``.
 
-    If your entity has some properties with Doctrine relationships, the factories for the related entities must be registered
-    in the service container so that the maker command can find them. To do that, add the following to your ``services.yaml``.
+.. tip::
+
+    When using ``--test`` flag, we're still dealing with ``dev`` environment. And because we want the container to know about our factories,
+    we need to declare them as services even if they are in the ``tests`` directory. To do that, add the following to your configuration:
 
     .. code-block:: yaml
 
-        # config/packages/services.yaml
+        # config/packages/zenstruck_foundry.yaml
         when@dev:
             services:
-                _defaults:
+                App\Tests\Factory\:
+                    resource: '../../tests/Factory/'
                     autowire: true
                     autoconfigure: true
-
-                App\Tests\Factory\:
-                    resource: '../tests/Factory/'
 
 .. tip::
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -192,7 +192,7 @@ This command will generate a ``PostFactory`` class that looks like this:
         when@dev:
             services:
                 App\Tests\Factory\:
-                    resource: '../../tests/Factory/'
+                    resource: '%kernel.project_dir%/tests/Factory/'
                     autowire: true
                     autoconfigure: true
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -181,6 +181,21 @@ This command will generate a ``PostFactory`` class that looks like this:
 
     Using ``make:factory --test`` will generate the factory in ``tests/Factory``.
 
+    If your entity has some properties with Doctrine relationships, the factories for the related entities must be registered
+    in the service container so that the maker command can find them. To do that, add the following to your ``services.yaml``.
+
+    .. code-block:: yaml
+
+        # config/packages/services.yaml
+        when@dev:
+            services:
+                _defaults:
+                    autowire: true
+                    autoconfigure: true
+
+                App\Tests\Factory\:
+                    resource: '../tests/Factory/'
+
 .. tip::
 
     You can globally configure which namespace the factories will be generated in:


docs/index.rst
```diff
@@ -181,6 +181,21 @@ This command will generate a ``PostFactory`` class that looks like this:
 
     Using ``make:factory --test`` will generate the factory in ``tests/Factory``.
 
+    If your entity has some properties with Doctrine relationships, the factories for the related entities must be registered
+    in the service container so that the maker command can find them. To do that, add the following to your ``services.yaml``.
```
nikophil: I don't think the problem is about "entities [which] has some properties with Doctrine relationships"

I think the problem this resolves is something like:
> When using `--test` flag, we're still dealing with `dev` environment. And because we want the container to know about our factories, we need to declare them as services even if they are in the `test` directory. To do that, add the following to your ``services.yaml``

don't hesitate to rephrase, this problem is hard to explain in a meaningful way

nikophil (commented):
this is very relevant!

I don't know if you dug into some old PRs, but this is very related to [this comment](https://github.com/zenstruck/foundry/pull/787#discussion_r1917936025). Thus, I think this fixes #786 

I'm even wondering if we should not add this to the Foundry's recipe.
Actually, I think we should do this in `config/packages/zenstruck_foundry.yaml`, but I'm not 100% convinced, WDYT?

@kbond any thoughts? We talked about this few months ago

ttskch:
@nikophil 

Oh, I thought I'd looked into the old issue/PR, but I overlooked it.

You're right. I also think that registering files under `tests/` in the service container in the dev environment is a strange workaround. (I don't think a fundamental solution will be possible anytime soon, so I wanted to at least document it to clear up any current user confusion.)

Essentially, I think the solution outlined in your comment below would be preferable.

> [Or maybe we should add a configuration node in which users can declare where they store their factories, so that we can tag them (or maybe no need to tag, at least add them in another way to the `FactoryMap`)](https://github.com/zenstruck/foundry/pull/787#discussion_r1917936025)

In my opinion, ideally, the `make:factory` command shouldn't use service container tags as a way to obtain a list of existing factories.

The basic premise is that the `--test` and the `--namespace` options are very useful and shouldn't be abolished. Given that, it's natural to assume that "existing factories" can exist in both `src/` and `tests/` (or another folder that's only loaded in other environments).

Therefore, the way `make:factory` obtains the list of existing factories should be decoupled from the service container.

In my opinion, simply hard-coding the list of folders in `zenstruck_foundry.yaml` would be sufficient. What do you think?

nikophil:
> In my opinion, simply hard-coding the list of folders in zenstruck_foundry.yaml would be sufficient. What do you think?

not sure to agree, loading classes based on their path brings some boilerplate code which I'd prefer to not add here (is this psr4 ? psr0 ? etc...)

I think the best solution (well... the _least worth_ solution :sweat_smile:) is to document this the way you've already done it.

Please could you adapt the PR with the following:
- change the wording to something like [suggested in](https://github.com/zenstruck/foundry/pull/1000/files#r2418050259)
- add it as a `.. tip::`
- I do think we should advocate to add this to `packages/zenstruck_foundry.yml` I even think we could add a tip [in the recipe](https://github.com/symfony/recipes/blob/main/zenstruck/foundry/2.7/config/packages/zenstruck_foundry.yaml), maybe as some commented code, with a link to the documentation

thanks!

(hey this PR is the 1000th! :tada:)

ttskch:
@nikophil I've updated this PR and also [submitted a PR to recipe](https://github.com/symfony/recipes/pull/1459).

Note that since we assume that the configuration will be written in `zenstruck_foundry.yaml` rather than `services.yaml`, the value of the `resource` field has been changed from `'../tests/Factory/'` to `'../../tests/Factory/'`, and the `autowire` and `autoconfigure` settings have been moved under the `App\Tests\Factory\` key instead of `_defaults`.

(Congrats on 1,000 (PR|Issue)s 🎊)

nikophil:
Can't we use `%kernel.root-dir%`?

ttskch:
@nikophil Surely! (It's `%kernel.project_dir%` though) I fixed it like that👍

- https://github.com/zenstruck/foundry/pull/1000/commits/19da7a8a3e238865d1339e0baecd75bf0bea093a
- https://github.com/symfony/recipes/pull/1459/commits/fec4535878f78a81990a9ba21323dd2c1c44b91f

nikophil (approved):

nikophil:
thanks @ttskch

PR #1006, opened 2025-10-15
Accept as story any descendants of Story, not only direct descendants

alsciende:
This small change will let users create an abstract Story class, for example, and use this abstract class for their own Stories.


nikophil:
thanks @alsciende

PR #1007, opened 2025-10-16
fix: RepositoryAssertion::exist() $criteria should allow mixed

### LINKED ISSUE #1004: RepositoryAssertions exists methods type declaration
Hello !

As I am upgrading a project that used to be on Foundry v1 to v2, I encountered an issue while running the following assertion : 

```php
$factory::assert()->notExists($id);
```

The issue is : 

```
TypeError: Zenstruck\Foundry\Persistence\ProxyRepositoryDecorator::count(): Argument #1 ($criteria) must be of type array, int given, called in /application/vendor/zenstruck/foundry/src/Persistence/RepositoryAssertions.php on line 127
```

Looking into the `RepositoryAssertions`, I see that every method that has a `$criteria` parameter is typed `array`, which seems fine to me, but `exists` and `notExists` methods allow `mixed` type, as described here : 

```php
public function exists(mixed $criteria, string $message = 'Expected {entity} to exist but it does not.'): self
{
    // ...
}

public function notExists(mixed $criteria, string $message = 'Expected {entity} to not exist but it does.'): self
{
    // ...
}
```

I am currently using Foundry v2.7.5, which is the brand new version. I can obviously work around this by switching to an array of criteria. However, shouldn't it be typed `array` ? Is there any particular reason to not do so for these methods ?

Thank you guys and have a nice day !

nikophil:
fixes https://github.com/zenstruck/foundry/issues/1004

### DIFF
--- a/src/Persistence/RepositoryAssertions.php
+++ b/src/Persistence/RepositoryAssertions.php
@@ -113,22 +113,20 @@ final class RepositoryAssertions
 
     public function exists(mixed $criteria, string $message = 'Expected {entity} to exist but it does not.'): self
     {
-        Assert::that($this->repository->count($criteria))
-            ->isGreaterThan(0, $message, [
-                'entity' => $this->repository->getClassName(),
-                'criteria' => $criteria,
-            ]);
+        Assert::that($this->repository->find($criteria))->isNotEmpty($message, [
+            'entity' => $this->repository->getClassName(),
+            'criteria' => $criteria,
+        ]);
 
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


Fabrn:
Oh so you rather allow `mixed` instead of forcing `array` ? I guess it's perfectly fine however I would have thought that you would rather make `$criteria` an `array` for each method of the `RepositoryAssertions` class

nikophil:
I do think `MyFactory::assert()->exists($id)` is a valid use case

moreover, this PR reverts a previous PR where I introduced a BC break

PR #1013, opened 2025-10-22
fix: only use PersistedObjectsTracker when auto-refresh is enabled

### LINKED ISSUE #1012: PersistedObjectsTracker::$buffer must not be accessed before initialization since v2.7.6
If running  test with a Factory create in it gives me:

```
Error: Typed static property Zenstruck\Foundry\Persistence\Proxy\PersistedObjectsTracker::$buffer must not be accessed before initialization
/var/www/html/vendor/zenstruck/foundry/src/Persistence/Proxy/PersistedObjectsTracker.php:53
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistenceManager.php:89
/var/www/html/vendor/zenstruck/foundry/src/Persistence/PersistentObjectFactory.php:266
/var/www/html/vendor/zenstruck/foundry/src/Factory.php:66
/var/www/html/tests/ApiResource/DemoTest.php:21

```

UserFactory is a default generated factory of a simple default symfony user.

DemoTest:
```PHP
<?php

declare(strict_types=1);

namespace App\Tests\ApiResource;

use ApiPlatform\Symfony\Bundle\Test\ApiTestCase;
use App\Factory\User\UserFactory;
use Zenstruck\Foundry\Test\Factories;
use Zenstruck\Foundry\Test\ResetDatabase;


final class DemoTest extends ApiTestCase
{
    use ResetDatabase;
    use Factories;

    protected static ?bool $alwaysBootKernel = false;

    public function testDemo():void{
        $user = UserFactory::createOne();
        self::assertNotNull($user);
    }
}

```

Reverting to v2.7.5 Solves the issue.

I don't have the issue when I run the whole suite, only when I run the one test or test file with phpstorm.

nikophil:
fixes https://github.com/zenstruck/foundry/issues/1012

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -140,6 +140,10 @@ final class Configuration
         self::$instance->bootedForDataProvider = true;
     }
 
+    /**
+     * /!\ Until PHPUnit 9 support is not dropped, this method MUST NOT call Configuration::instance()
+     * Otherwise, it will reboot the kernel, leading to complex bugs
+     */
     public static function shutdown(): void
     {
         PersistedObjectsTracker::reset();
--- a/src/Persistence/PersistenceManager.php
+++ b/src/Persistence/PersistenceManager.php
@@ -86,7 +86,9 @@ final class PersistenceManager
             $this->flush($om);
         }
 
-        PersistedObjectsTracker::updateIds();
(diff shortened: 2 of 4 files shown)


nikophil:
--prefer-lowest jobs are failing, I'll fix this later

PR #1028, opened 2025-11-06
Remove array params on function alwaysForce

philpichet:
The function alwaysForce expect strings, not an array of string.


nikophil:
thanks @philpichet

PR #1029, opened 2025-11-07
feat: add hooks priority

### LINKED ISSUE #1027: After instantiate hook should occur before `$em->persist()` call
this way, pre-persist listeners run after "after instantiate" hooks

nikophil:
In the meantime, this fixes https://github.com/zenstruck/foundry/issues/1027

### DIFF
--- a/src/Persistence/PersistentObjectFactory.php
+++ b/src/Persistence/PersistentObjectFactory.php
@@ -41,9 +41,11 @@ use function Zenstruck\Foundry\set;
  */
 abstract class PersistentObjectFactory extends ObjectFactory
 {
+    public const PRIORITY_SCHEDULE_FOR_INSERT = -10;
+
     private PersistMode $persist = PersistMode::PERSIST;
 
-    /** @phpstan-var list<callable(T, Parameters, static):void> */
+    /** @phpstan-var array<int, list<callable(T, Parameters, static):void>> */
     private array $afterPersist = [];
 
     /** @var list<callable(T):void> */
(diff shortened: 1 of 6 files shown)


src/Persistence/PersistentObjectFactory.php
```diff
@@ -41,9 +41,11 @@
  */
 abstract class PersistentObjectFactory extends ObjectFactory
 {
+    public const PRIORITY_SCHEDULE_FOR_INSERT = -10;
```
nikophil, line 44: I'm exposing this constant, so that it is easier to create "after instantiate" hooks which run after pre-persist listeners. 

This is very niche feature, I don't think we need to document it

src/Persistence/PersistentObjectFactory.php
```diff
@@ -526,14 +535,15 @@ static function(object $object, array $parameters, PersistentObjectFactory $fact
...
+                    foreach (array_merge(...$factoryUsed->afterPersist) as $afterPersist) {
                         $afterPersistCallbacks[] = static function() use ($object, $afterPersist, $parameters, $factoryUsed): void {
                             $afterPersist($object, $parameters, $factoryUsed);
                         };
                     }
 
                     Configuration::instance()->persistence()->scheduleForInsert($object, $afterPersistCallbacks);
-                }
+                },
+                self::PRIORITY_SCHEDULE_FOR_INSERT
```
nikophil, line 546: I'm adding a negative priority here: this "after instantiate" hook is always the first one added, thus it was the first one called. And because we are calling `$em->persist()` inside it, it was impossible to run "after insantiate" hooks before "pre persist" doctrine listeners. Now this is the default behavior.

In my current work project we have Doctrine listeners which perform some consistency checks, so it was impossible to create consistent objects using "after instantiate" hooks.

PR #1041, opened 2025-12-18
fix: call inverse relationship callback before afterInstantiate()

### LINKED ISSUE #1038: Cascade persist order
Hi, I just try to upgrade from v1.38 to v2.8, but we are facing an issue.

We have 2 entities: `Parent` and `Child`, the `Child` is cascade persisted from the `Parent` entity.

In the app, a user create a `Parent` entity with at least a `Child`, when we persist the `Parent` doctrine cascade persist the children. 

In v1.38, the `ParentFactory` allow us access the `Child` entities in `->afterInstantiate()` and `->afterPerist()` factory event, but actually we can´t.

The problem is: the `Parent` entity has a non nullable `reference` property that is defined from `->getChildren()`, but the factory do it:
 - 1. ParentFactory: ->afterInstantiate
 - 2. ParentFactory: ->afterPersist
 - 3. Child: ->afterInstantiate
 - 4. Child: ->afterPerist

So trying to persist the `Parent` with a null `reference` because we cannot process the reference without `Children` that is persisted after.

```php
class Parent
{
    #[ORM\Id, ORM\Column, ORM\GeneratedValue]
    private ?int $id = null;

    #[ORM\Column]
    private ?string $reference = null;

    #[Assert\Count(min: 1)]
    #[Assert\Valid]
    #[ORM\OneToMany(targetEntity: Child::class, mappedBy: 'parent', cascade: ['persist', 'remove'], orphanRemoval: true)]
    #[ORM\OrderBy(['id' => 'ASC'])]
    private Collection $children;

    public function __construct()
    {
        $this->children = new ArrayCollection();
    }

    public function getReference(): ?string
    {
        return $this->reference;
    }

    public function setReference(string $reference): static
    {
        $this->reference = $reference;

        return $this;
    }

/**
     * @return ReadableCollection<int, Child>
     */
    public function getChildren(): ReadableCollection
    {
        return $this->children;
    }

    public function addChild(Child $child): static
    {
        if (!$this->children->contains($child)) {
            $this->children[] = $child;
            $child->setParent($this);
        }

        return $this;
    }

    public function removeChild(Child $child): static
    {
        if ($this->children->removeElement($child)) {
            if ($child->getParent() === $this) {
                $child->setParent(null);
            }
        }

        return $this;
    }
}
```

```php
class Child
{
    #[ORM\Id, ORM\Column, ORM\GeneratedValue]
    private ?int $id = null;

    #[ORM\ManyToOne(targetEntity: Parent::class, inversedBy: 'children')]
    private ?Parent $parent = null;

   // Many other scalar properties

    public function getId(): ?int
    {
        return $this->id;
    }

    public function getParent(): ?Parent
    {
        return $this->parent;
    }

    public function setParent(?Parent $parent): static
    {
        $this->parent = $parent;

        return $this;
    }

    // Many getters/setters
}
```

Previously we can do it in the factory:
```php
final class ParentFactory extends PersistentProxyObjectFactory
{
    {
        // In v2.8, ->many() is replaced by ->range()
        // Important: withoutPersisting to avoid the Factory trying persist children, Doctrine cascade it
        return [
            'children' => ChildFactory::new()->withoutPersisting()->many(1, 2),
        ];
    }

    {
        // see https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#initialization
        return $this
            ->afterInstantiate(function (Parent $parent, array $attributes): void {
               // Here the $parent should have children inside, but no more available
               // We call a Service to calculate the reference like in Controllers
            })
        ;
    }

    {
        return Parent::class;
    }
}
```

nikophil:
fixes https://github.com/zenstruck/foundry/issues/1038

@mpiot wanna give a try before I merge this PR?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@
 /docker-compose.override.yaml
 /tests/Fixture/Maker/tmp/
 /.phpbench
+/config/reference.php
--- a/phpunit-deprecation-baseline.xml
+++ b/phpunit-deprecation-baseline.xml
@@ -18,6 +18,7 @@ See https://github.com/zenstruck/foundry/blob/2.x/UPGRADE-2.7.md to upgrade.]]><
             <issue><![CDATA[Since doctrine/mongodb-odm 2.14: Not using native lazy objects is deprecated and will be impossible in Doctrine MongoDB ODM 3.0.]]></issue>
             <issue><![CDATA[Since symfony/framework-bundle 7.3: Not setting the "property_info.with_constructor_extractor" option explicitly is deprecated because its default value will change in version 8.0.]]></issue>
             <issue><![CDATA[Since symfony/console 7.4: The "Symfony\Component\Console\Application::add()" method is deprecated and will be removed in Symfony 8.0, use "Symfony\Component\Console\Application::addCommand()" instead.]]></issue>
+            <issue><![CDATA[Since doctrine/mongodb-odm 2.3: Document short namespace aliases are deprecated, use ::class constant instead.]]></issue>
         </line>
     </file>
     <file path="vendor/doctrine/deprecations/src/Deprecation.php">
--- a/src/Persistence/PersistentObjectFactory.php
+++ b/src/Persistence/PersistentObjectFactory.php
@@ -50,7 +50,7 @@ abstract class PersistentObjectFactory extends ObjectFactory
     private array $afterPersist = [];
 
     /** @var list<callable(T):void> */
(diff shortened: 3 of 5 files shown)


mpiot:
@nikophil Thanks a lot for your PR, this one fix our error 🙏

nikophil:
@mpiot released in 2.8.3

mpiot:
@nikophil thanks so much for the feedback :)

PR #1042, opened 2025-12-19
fix: auto-refresh problem with doctrine/orm 2

### LINKED ISSUE #1040: PHP 8.4 Lazy Ghost Object initialization for Auto-Refresh doesn’t always work with Doctrine ORM 2.x
When we tried to upgrade to Foundry 2.7.3 a few weeks ago our testsuit failed due a lot of errors. These days I found some time to dig into the issue.

The issue exists since 45214f770d7dc93a49436c61dc0c2b583d7c76c8 #996. The implementation seem to rely on Doctrine ORM to trigger the initialization of the lazy ghost object while computing the changeset.

In our case this doesn’t work with some entities, because these entities have untyped properties (and we can’t change it, because they are part of vendor code).

If the entity has at least one typed property without default value, the refresh gets triggered in `computeChangeSet` via `$refProp->getValue($entity)` because `\Doctrine\Persistence\Reflection\TypedNoDefaultReflectionProperty` is used.

Otherwise, the refresh isn’t triggered because `\Doctrine\Persistence\Reflection\RuntimeReflectionProperty` is used, which is designed to avoid lazy loading, by accessing the property by casting the object to an array and hence not triggering the initialization of the lazy ghost object.

This leads to doctrine computing the changeset based on an uninitialized („empty“) object, which in turn leads to an update statement which tries to set all attributes to  `null`.

You can find a reproducer at https://github.com/xelaris/foundry-no-lazy-ghost-initialization-reproducer.

I’m not quite sure where and how this might be solved. Maybe Foundry could explicitly trigger the initialization of lazy ghosts in `PersistedObjectsTracker` before calling `ObjectManager::flush()`, at least when dealing with ORM 2.x.

The issue doesn’t seem to be reproducible with ORM 3.4 (with `enable_native_lazy_objects` being set to `true`). The initialization/auto-refresh isn’t triggered there either, but Doctrine doesn’t try to update the entity.`\Doctrine\ORM\Mapping\PropertyAccessors\RawValuePropertyAccessor` would be used there. This might need some more investigation. For the time being upgrading to 3.4 wouldn’t be an option for us though unfortunately, due to dependencies.

Btw… there is a catch when using a debugger: When stepping through the code, and/or setting a breakpoint at the „wrong“ place (e.g. inside `\Doctrine\ORM\UnitOfWork::computeChangeSet`) the debugger itself may trigger the initialization of the lazy ghost object and the issue isn’t reproducible in that case.

nikophil:
fixes https://github.com/zenstruck/foundry/issues/1040

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "dama/doctrine-test-bundle": "^8.0",
         "doctrine/collections": "^1.7|^2.0",
         "doctrine/common": "^3.2.2",
-        "doctrine/doctrine-bundle": "^2.10|^3.0|3.1.x-dev",
+        "doctrine/doctrine-bundle": "^2.10|^3.0",
         "doctrine/doctrine-migrations-bundle": "^2.2|^3.0",
         "doctrine/mongodb-odm": "^2.4",
         "doctrine/mongodb-odm-bundle": "^4.6|^5.0",
@@ -43,7 +43,7 @@
         "symfony/event-dispatcher": "^6.4|^7.0|^8.0",
         "symfony/framework-bundle": "^6.4|^7.0|^8.0",
         "symfony/maker-bundle": "^1.55",
-        "symfony/phpunit-bridge": "^6.4|^7.0|^8.0",
+        "symfony/phpunit-bridge": "^6.4.26|^7.0|^8.0",
         "symfony/routing": "^6.4|^7.0|^8.0",
         "symfony/runtime": "^6.4|^7.0|^8.0",
         "symfony/translation-contracts": "^3.4",
@@ -84,8 +84,8 @@
     },
     "conflict": {
         "doctrine/persistence": "<2.0",
-        "symfony/event-dispatcher": "<6.4",
-        "symfony/framework-bundle": "<6.4"
+        "doctrine/cache": "<1.12.1",
+        "symfony/polyfill-php80": "<1.16"
     },
     "extra": {
         "bamarni-bin": {
--- a/phpunit
+++ b/phpunit
@@ -85,6 +85,10 @@ case ${PHPUNIT_VERSION} in
       PHPUNIT_EXEC="${PHPUNIT_EXEC} --extension "${FOUNDRY_EXTENSION}""
     fi
 
+    if [ "${WITH_LOWEST_DEPENDENCIES:-0}" = "1" ]; then
+      PHPUNIT_EXEC="${PHPUNIT_EXEC} --do-not-fail-on-deprecation --do-not-fail-on-phpunit-deprecation"
+    fi
+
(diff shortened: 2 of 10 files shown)


src/Persistence/PersistedObjectsTracker.php
```diff
@@ -52,7 +54,14 @@ public function add(object ...$objects): void
     {
         foreach ($objects as $object) {
             if (self::$trackedObjects->offsetExists($object) && self::$trackedObjects[$object]) {
-                self::proxifyObject($object, self::$trackedObjects[$object]);
+                if (DoctrineOrmVersionGuesser::isOrmV3()) {
+                    self::resetObjectAsLazyGhost($object, self::$trackedObjects[$object]);
+                } else {
+                    try {
+                        refresh($object);
```
nikophil, line 61: we're still forced to refresh objects that are coming from the repository but which are already managed. See https://github.com/zenstruck/foundry/issues/975

this is just the way to do it with `doctrine/orm` 2 😅

phpunit
```diff
@@ -85,6 +85,10 @@ case ${PHPUNIT_VERSION} in
       PHPUNIT_EXEC="${PHPUNIT_EXEC} --extension "${FOUNDRY_EXTENSION}""
     fi
 
+    if [ "${WITH_LOWEST_DEPENDENCIES:-0}" = "1" ]; then
+      PHPUNIT_EXEC="${PHPUNIT_EXEC} --do-not-fail-on-deprecation --do-not-fail-on-phpunit-deprecation"
+    fi
```
nikophil, line 90: this saved me a lot of headaches 😅

composer.json
```diff
@@ -84,8 +84,8 @@
     },
     "conflict": {
         "doctrine/persistence": "<2.0",
-        "symfony/event-dispatcher": "<6.4",
-        "symfony/framework-bundle": "<6.4"
+        "doctrine/cache": "<1.12.1",
+        "symfony/polyfill-php80": "<1.16"
```
nikophil, line 88: this was actually needed in order to actually install orm 2 when using `--prefer-lowest`

composer.json
```diff
@@ -43,7 +43,7 @@
         "symfony/event-dispatcher": "^6.4|^7.0|^8.0",
         "symfony/framework-bundle": "^6.4|^7.0|^8.0",
         "symfony/maker-bundle": "^1.55",
-        "symfony/phpunit-bridge": "^6.4|^7.0|^8.0",
+        "symfony/phpunit-bridge": "^6.4.26|^7.0|^8.0",
```
nikophil, line 46: previous version actually messes with `$this->expectDeprecation()` with phpunit versions > 9  🤷

Issue #1034, opened 2025-11-25
Booting the `Configuration` from the bundle when using Behat and FriendsOfBehat/SymfonyExtension


mpdude:
This is not really a bug report or a feature request. But since it took me some time to investigate and understand what's going on, I thought I might drop the information here, hoping that it might be useful for others, a future me – or maybe it brings insights and helps to draw conclusions or see parallels with other issues. Let's go.

I am using Behat for acceptance testing and Foundry to create objects (mostly Doctrine Entities) from my Behat `Context` classes. Basically, these context classes contain the fragments of code that are wired together by the tests specification written in Gherkin.

To integrate with Symfony, I use the https://github.com/FriendsOfBehat/SymfonyExtension/ Behat extension. This extension does two things:

1. When Behat starts, it creates and boots an instance of the Symfony kernel and prepares the DIC. This allows to have Behat Context classes that can be created from or wired with services by the Symfony container.
2. When making requests to the Symfony application in functional (or "end-to-end") testing, this extension takes care of _creating a second `Kernel` instance_, booting and resetting it between each scenario. This second Kernel is what is running the application (the system under test). 

This is also described in https://github.com/FriendsOfBehat/SymfonyExtension/blob/master/DOCUMENTATION.md#isolated-kernel-and-container-for-the-mink-driver. That way, you can make sure that the application's kernel is clean in a way that it has not been affected by the test setup phase. In particular, I care about the entity manager being in a clean state, not holding anything in its Identity Map and loading everything from the database.

The ~problem~ challenge is that `ZenstruckFoundryBundle::boot()` is called multiple times – the first time when the "outer" Kernel and container (the one Behat is running on/within) is created and booted. And then once before each scenario when the `SymfonyExtension` resets/reboots the "inner" (application) Kernel.

This leads to `Configuration::boot()` being called with the `.zenstruck_foundry.configuration` from the "outer" container the very first time, and then with new instances of that before every scenario starts. This means that Foundry in most cases is actually using the _application's_ entity manager instance to persist entities.

I have created the following Behat context class to work around that:

```php
<?php

namespace App\Behat;

use Behat\Behat\Context\Context;
use Behat\Hook\BeforeScenario;
use Symfony\Component\DependencyInjection\Attribute\Autowire;
use Zenstruck\Foundry\Configuration;

class FoundryContext implements Context
{
    public function __construct(
        #[Autowire(service: '.zenstruck_foundry.configuration')]
        private readonly Configuration $configuration
[...]

PR #1049, opened 2026-01-08
refactor: explicitly boot Foundry in ResetDatabase

nikophil:
related to https://github.com/zenstruck/foundry/issues/1034

### DIFF
--- a/src/Persistence/ResetDatabase/ResetDatabaseManager.php
+++ b/src/Persistence/ResetDatabase/ResetDatabaseManager.php
@@ -112,6 +112,6 @@ final class ResetDatabaseManager
 
     private static function canSkipSchemaReset(): bool
     {
-        return PersistenceManager::isOrmOnly() && self::isDAMADoctrineTestBundleEnabled();
+        return self::isDAMADoctrineTestBundleEnabled() && PersistenceManager::isOrmOnly();
     }
 }
--- a/src/Test/ResetDatabase.php
+++ b/src/Test/ResetDatabase.php
@@ -14,6 +14,8 @@ namespace Zenstruck\Foundry\Test;
 use PHPUnit\Framework\Attributes\Before;
 use PHPUnit\Framework\Attributes\BeforeClass;
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+use Symfony\Component\HttpKernel\KernelInterface;
+use Zenstruck\Foundry\Configuration;
 use Zenstruck\Foundry\Persistence\ResetDatabase\ResetDatabaseManager;
 
 /**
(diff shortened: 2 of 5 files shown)


nikophil:
@HypeMC would this do the trick?

HypeMC:
@nikophil Thank you, I'll give it a try right away.

src/Persistence/ResetDatabase/ResetDatabaseManager.php
```diff
@@ -77,12 +77,13 @@ public static function resetBeforeFirstTest(callable $createKernel, callable $sh
      */
     public static function resetBeforeEachTest(callable $createKernel, callable $shutdownKernel): void
     {
+        $kernel = $createKernel();
+
```
HypeMC: @nikophil Unfortunately, this change sees to break the following case:

```php
class SomeTest extends WebTestCase
{
    use Factories;
    use ResetDatabase;

    public function testMyCode(): void
    {
        $client = self::createClient();

        // ...
    }
```

> LogicException: Booting the kernel before calling "Symfony\Bundle\FrameworkBundle\Test\WebTestCase::createClient()" is not supported, the kernel should only be booted once.

nikophil: Damn... I'll add this use case to our test suite right now! A test using `WebTestCase` already exists, but it does not use `ResetDatabase` trait

nikophil: @HypeMC mind to test this new version, please?

I added a test to ensure `ResetDatabase` does not mess with `self::createClient()`

HypeMC: On it.

HypeMC: @nikophil Sadly, this reintroduces the original issue:

> Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The "kernel" service is synthetic, it needs to be set at boot time before it can be used.

nikophil: but the test you introduced in https://github.com/zenstruck/foundry/pull/993 is still green, we'll have to figure out why, I guess

HypeMC: I'll look into it tonight, and try to figure it out.

nikophil: super nice, thanks. Sorry to come back on this stuff but this would unlock an easy way to integrate Foundry with Behat, which is one of the next things I'd like to see in foundry

HypeMC: @nikophil Apologies for the late response, I couldn't find the time yesterday. Long story short, one test failed because it didn't use `Factories`. Once that was added, everything passed. So this fix is actually good. Thank you for investigating this and sorry for the false alert.

HypeMC (approved):

src/Persistence/ResetDatabase/ResetDatabaseManager.php
nikophil: ok thanks, this is very nice.

This kind of problem won't exist anymore once we have [deprecated `Factories` trait](https://github.com/zenstruck/foundry/pull/968) :smile:

PR #1050, opened 2026-01-11
feature: Behat extension

### LINKED ISSUE #235: Behat Integration

nikophil:
Here's a new attempt to introduce Behat support info Foundry :blush: 

Here are some remarks: 
- ~according to [this comment](https://github.com/zenstruck/foundry/pull/378), I've built this PR in a way where `src/Behat` could be a subtree (using [splitsh](https://github.com/splitsh/lite)), with its own repo and composer package. I'm still torn on this, because it involves a lot more complexity, but I feel like it's the way to go (maybe this would pave the way for other packages such as `zenstruck/foundry-doctrine-orm`, `zenstruck/foundry-doctrine-odm`, ...)~ 
EDIT: Due to the complexity of this approach, we've decided that we won't do this now, moreover it does not feel necessary for a package which is mostly a test/dev package.
- this current PR is the minimum code needed to have Foundry working with Behat out of the box. In further PRs, I'd like to add the following features:
    - add a `@withStory [story]` tag, that would behave like [`#[WithStory]` attribute](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#withstory-attribute) in PHPUnit.
    - add a way to reset the database (maybe with a tag as well, it could also be configured globally once for all in the extension configuration). As well as in PHPUnit, this mechanism should support a reset mode (schema or migrate) and [Foundry's global states](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#global-state)  
    - the reset database should be compatible with [dama/docintre-test-bundle](https://github.com/dmaicher/doctrine-test-bundle)

ping @mpdude @mellowgrab @kevin-schmitt @roboticflamingo don't hesistate to open some other issues with ideas about how  Foundry could be integrated in Behat

fixes https://github.com/zenstruck/foundry/issues/235 https://github.com/zenstruck/foundry/issues/1034


mpdude:
Great to see you're giving this a shot!

I've been carrying around this idea with me for quite some time now. Still, there's one piece of the puzzle that I could not yet work out in a generic way and that I will try to outline in the following. I find it very useful and convenient, and have successfully (IMHO) applied it in a few projects. I think it would be great if we could support it directly from a Foundry Behat extension in a generic way... let's go!

To start with Behat scenarios, I have something like:

```gherkin
Feature: Blog post detail page shows post categories

    Scenario: Blog post with one category
		Given a blog post category C has been created
		And the blog post category C has the name "random thoughts"
		And a blog post P has been created
		And the blog post P is categorized as C
		When I go to the blog post P detail page
		Then the "#categories" element should contain "random thoughts"
```

So, basically what's going on here is that I have a special `BuilderContext` that knows how to create Foundry Factories, and which Factory to use for what noun.

blog post category => CategoryFactory
blog post => BlogPostFactory

The names `C`, `P` etc. are used to store those factory instances in a simple array which is tracking state for that context.

Then, phrases like `blog post category C has the name "random thoughts"` take the appropriate factory instance from that array, do the right thing `->with(...)` it, and put the new value back into the array.

Then, there is a scenario hook that is triggered when the first `When` step is reached. It will look at all objects in the BuilderContext array and turn factories into object instances, making sure everything is flushed to the database.

It requires the use of Foundry `Lazy` in a few places to defer accessing the real objects a bit... the `Lazy` callback will basically look into the array as well.

Unfortunately, I can't directly share real-life code as an example, but I hope you get the idea. In those cases, all this is written out as stupid boilerplate code.

What I am hoping for is to make it generic with the help of Foundry and/or its Behat extension.

For example, if the `BlogPostFactory` could be marked with an attribute like `#[FoundryContextNoun('blog post')]`, or the `CategoryFactory` had an attribute `#[FoundryContextWith('has the name "(?P<name>\w+)"')]`, maybe we could find a way to wire it all up automagically?

src/Test/Behat/BehatListener.php
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            ScenarioTested::BEFORE => 'bootFoundry',
+            ExampleTested::BEFORE => 'bootFoundry',
+            ScenarioTested::AFTER => 'shutdownFoundry',
+            ExampleTested::AFTER => 'shutdownFoundry',
+        ];
```
nikophil, line 25: I think a listener is better suited for this, it would give a better control on the priorities, more over if we want to make Foundry for Behat nicely integrate with dama [which is using a listener too](https://github.com/dmaicher/doctrine-test-bundle/blob/master/src/Behat/BehatListener.php)

nikophil:
@mpdude this becomes interesting!

> Great to see you're giving this a shot!

yes I got now some personal interest to make Foundry working with Behat: we're using Behat a lot in my new workplace, so expect some work around this soon!

I was indeed planing to ask for any good ideas about a possible API for this Behat support (I've currently never used Foundry with Behat yet :sweat_smile:). I kinda like your idea, I'll open an issue with your proposal to discuss around it. I'd like this PR to only provide the skeleton for the the Behat extension, all other features will have their own PR.

mpdude:
I'm happy this finally was the trigger to write down the idea.

nikophil:
closed in favor of https://github.com/zenstruck/foundry/pull/1083

PR #1060, opened 2026-01-19
fix(autorefresh): prevent weird autorefresh recursion

nikophil:
@tandev mind to test the fix please?

I'm pretty sure it fixes your problem, but let's test before merge :blush:


tandev:
Yes, together with the other fix, all tests green :) Thank you!

nikophil:
@tandev the two fixes are released in https://github.com/zenstruck/foundry/releases/tag/v2.8.5

PR #1061, opened 2026-01-19
fix(faker): change how FOUNDRY_FAKER_SEED is read

### LINKED ISSUE #1055: faker seed is not applied after kernel shutdown
hi,

image a symfony kernel testcase like this:

```php
class FakerTest extends ApplicationWebTestCase {
    public function test(): void {
//        $this->client->request('get','/');
        static::assertSame('dolore', faker()->word());
    }
    public function test2(): void {
        static::assertSame('distinctio', faker()->word());
    }

    public function test3(): void {
        // 924691
        static::assertSame('dolore', faker()->word());
    }
    public function test4(): void {
        // 924691
        static::assertSame('distinctio', faker()->word());
    }
}
```

(the kerneltestcase has `use Factories;`, nothing to fancy in the shutdown, kernel gets rebootet with every test)

running this via

```shell
$ FOUNDRY_FAKER_SEED=924691 phpunit --testsuite Application --filter Faker
[...]
..FF                                                                4 / 4 (100%)

Faker seed: 924691

Time: 00:00.071, Memory: 28.00 MB

There were 2 failures:

1) FakerTest::test3
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'dolore'
+'odit'

FakerTest.php:17

2) FakerTest::test4
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'distinctio'
+'vitae'
```

i get what i expect, the "errors" stay the same, because the seed is actually used. if i now run the request `$this->client->request('get','/');` , those values will not stay the same anymore,

i would expect to be stable, but with every run of the tests there will be different outcomes 🤔 

```
Current configuration for extension with alias "zenstruck_foundry"
==================================================================

zenstruck_foundry:
    persistence:
        flush_once: true
    auto_refresh_proxies: null
    faker:
        locale: null
        seed: null
        service: null
    instantiator:
        use_constructor: true
        allow_extra_attributes: false
        always_force_properties: false
        service: null
    global_state: {  }
    orm:
        auto_persist: true
        reset:
            connections:
                - default
            entity_managers:
                - default
            mode: !php/enum Zenstruck\Foundry\ORM\ResetDatabase\ResetDatabaseMode::SCHEMA
            migrations:
                configurations: {  }
    mongo:
        auto_persist: true
        reset:
            document_managers:
                - default
    make_factory:
        default_namespace: Factory
        add_hints: true
    make_story:
        default_namespace: Story
```

this is symfony8, php8.5, latest foundry version

any suggestions where to look for to fix this?

nikophil:
fixes https://github.com/zenstruck/foundry/issues/1055

@verfriemelt-dot-org could you test this branch please?

I don't understand the expectations in your reproducer, though:
```php
    public function test(): void {
        static::assertSame('dolore', faker()->word());
    }
    public function test2(): void {
        static::assertSame('distinctio', faker()->word());
    }
```

The faker seed needs to be reset between each test, so we can safely run all tests as well as each test independently and expect the same result within each test:
```php
class FakerTest extends WebTestCase
 {
    use Factories;

    public function test1(): void
    {
        static::assertSame('dolore', faker()->word());
        static::assertSame('distinctio', faker()->word());
    }

    public function test2(): void
    {
        static::assertSame('dolore', faker()->word());
        static::assertSame('distinctio', faker()->word());
    }
```

image


config/services.php
```diff
@@ -42,7 +42,8 @@
             service('.zenstruck_foundry.story_registry'),
             service('.zenstruck_foundry.persistence_manager')->nullOnInvalid(),
             param('zenstruck_foundry.persistence.flush_once'),
-            '%env(default:zenstruck_foundry.faker.seed:int:FOUNDRY_FAKER_SEED)%',
```
nikophil, line 45: this `default` env var resolver was a mistake, let's split it in two

verfriemelt-dot-org:
this solves the issue in our project :) thank you.

the reproducer was intended to have "stable" faker values for each test. this was not the case.
with this PR checked out, everything works as expected!

nikophil:
super nice, I think the faker seed is handled much better now! 

I hope this won't create BC breaks in other people's testsuites :sweat_smile: 

I'll release this very soon

PR #1063, opened 2026-01-20
fix: always register "inverseRelationshipCallbacks" even if Foundry is not booted

### LINKED ISSUE #1062: Unable to generate OneToMany collection using dataproviders in v2.8.3
Hi,

first things first: I love working with Foundry. It's beautiful ❤️ 

## Prerequisites

- zenstruck/foundry: v2.8.2
- php: 8.1.34
- doctrine/orm: 2.20.9

## Entity and relations

```php
class Episode
{
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\Column(name: 'id', type: Types::INTEGER)]
    private ?int $id = null;

    /**
     * @var Collection<int, Segment>
     */
    #[ORM\OneToMany(mappedBy: 'episode', targetEntity: Segment::class, cascade: ['persist', 'remove'], orphanRemoval: true)]
    #[ORM\JoinColumn(onDelete: 'CASCADE')]
    #[ORM\OrderBy(['position' => 'ASC'])]
    private Collection $segments;

    /**
     * @return Collection<int, Segment>
     */
    public function getSegments(): Collection
    {
        return $this->segments;
    }

   // getter/setters included, but removed for simpleness
}

class Segment
{
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\Column(name: 'id', type: Types::INTEGER)]
    private ?int $id = null;

    #[ORM\ManyToOne(targetEntity: Episode::class, cascade: ['persist'], inversedBy: 'segments')]
    #[ORM\JoinColumn(name: 'episode_id', nullable: false, onDelete: 'CASCADE')]
    private ?Episode $episode = null;

    // getter/setters included, but removed for simpleness
}
```

## Error description

Running my tests on v2.8.2 works fine. However updating to v2.8.3 fails, due to not generating related entities.

Pls note this is not the exact test I'm running (e.g. I'm trusting you/foundry to create a relation I defined in my arrange-part of testing code). **This is a minimal reproducer.** I noticed, this fails in a dataprovider, but does not fail using in tests directly. However, I'm using multiple tests with multiple dataproviders, so this should work IMO.

```php
class MyTestFile extends KernelTestCase
{
    use ResetDatabase;
    use Factories;

    #[Test]
    #[DataProvider('provider')]
    public function example(string $expectation, EpisodeFactory $factory): void
    {
        $episode = $factory->create();
        self::assertNotEmpty($episode->getSegments()); // since v2.8.3 this is an empty array, expected 2 segment entities here
    }

    /**
     * @return iterable<non-empty-string, array{non-empty-string, EpisodeFactory}>
     */
    public static function provider(): iterable
    {
        $factory = EpisodeFactory::new([
            // properties
            'segments' => SegmentFactory::new(['state' => TranscodingState::DISTRIBUTED])->sequence([['isLastSegment' => false], ['isLastSegment' => true]]),
        ])->withoutPersisting();

        yield 'protected' => ['foobar', $factory->with([/* properties */])];
    }
```

Running this test works totally fine in v2.8.2. After upgrading to v2.8.3+ (tested with 2.8.4, too) it does not work.

nikophil:
fixes https://github.com/zenstruck/foundry/issues/1062

ping @mvhirsch 

actually I've asked Claude to fix this bug, and even with almost no context, it managed to understand where the bug was coming from. I think I would have struggled more than ~him~ it :sweat_smile: :scream:

### DIFF
--- a/src/Persistence/PersistentObjectFactory.php
+++ b/src/Persistence/PersistentObjectFactory.php
@@ -547,31 +547,35 @@ abstract class PersistentObjectFactory extends ObjectFactory
                 },
                 self::PRIORITY_SCHEDULE_FOR_INSERT
             )
+
+            // Always register the callback that executes inverse relationship callbacks.
+            // This is needed even when Foundry is not booted (e.g., when factories are created in data providers
+            // without the PHPUnit extension), because the callbacks are populated later during create().
+            ->afterInstantiate(
+                static function(object $object, array $parameters, self $factoryUsed): void {
+                    $tempAfterInstantiateCallbacks = $factoryUsed->inverseRelationshipCallbacks;
+                    $factoryUsed->inverseRelationshipCallbacks = [];
+                    foreach ($tempAfterInstantiateCallbacks as $tempAfterInstantiateCallback) {
+                        $tempAfterInstantiateCallback($object);
+                    }
+                },
+                priority: 1000
+            )
         ;
 
         if (!Configuration::isBooted() || !Configuration::instance()->hasEventDispatcher()) {
             return $factory;
         }
 
-        return $factory->afterInstantiate(
-            static function(object $object, array $parameters, self $factoryUsed): void {
-                $tempAfterInstantiateCallbacks = $factoryUsed->inverseRelationshipCallbacks;
-                $factoryUsed->inverseRelationshipCallbacks = [];
-                foreach ($tempAfterInstantiateCallbacks as $tempAfterInstantiateCallback) {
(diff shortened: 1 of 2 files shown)


mvhirsch:
Nice one. So I give credits to Claude(? :sweat_smile:) and a big thank you to you! :heart:

nikophil:
> So I give credits to Claude

I paid for it, so I deserve all the credit :laughing:

PR #1064, opened 2026-01-21
minor: some house keeping

nikophil:
do some cleaning before merging https://github.com/zenstruck/foundry/pull/968

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -21,7 +21,7 @@ concurrency:
 
 jobs:
   tests:
-    name: P:${{ matrix.php }}, S:${{ matrix.symfony }}, D:${{ matrix.database }}, PU:${{ matrix.phpunit }}${{ matrix.deps == 'lowest' && ' (lowest)' || '' }}${{ matrix.use-phpunit-extension == 1 && ' (phpunit extension)' || '' }}${{ (matrix.use-php-84-lazy-objects != 1 || matrix.php != '8.4') && ' (legacy proxy)' || '' }}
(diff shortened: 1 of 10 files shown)


.github/workflows/phpunit.yml
```diff
@@ -30,11 +30,12 @@ jobs:
         symfony: [ 6.4.*, 7.4.*, 8.0.* ]
         database: [ mysql|mongo ]
         phpunit: [ 11, 12 ]
-        use-php-84-lazy-objects: [ 1 ]
 
         # default values:
         # deps: [ highest ]
-        # use-phpunit-extension: [ 0 ]
+        # use-phpunit-extension: [ 1 ]
```
nikophil, line 37: let's use phpunit extension by default 🎉

.github/workflows/phpunit.yml
```diff
@@ -30,11 +30,12 @@ jobs:
...
+        # use-php-84-lazy-objects: [ 1 ]
+        # use-dama: [ 1 ]
```
nikophil: it appears that we need to have a permutation without dama in the main test suite

.github/workflows/phpunit.yml
```diff
@@ -30,11 +30,12 @@ jobs:
         symfony: [ 6.4.*, 7.4.*, 8.0.* ]
         database: [ mysql|mongo ]
         phpunit: [ 11, 12 ]
-        use-php-84-lazy-objects: [ 1 ]
+        use-dama: [ 1 ]
```
nikophil, line 33: it appears that we need to have a permutation without dama in the main test suite

PR #1072, opened 2026-01-29
minor: use test environment

nikophil:
Our testsuite was using `dev` environment, which becomes problematic with the future Behat extension

### DIFF
--- a/src/Mongo/MongoPersistenceStrategy.php
+++ b/src/Mongo/MongoPersistenceStrategy.php
@@ -54,13 +54,7 @@ final class MongoPersistenceStrategy extends PersistenceStrategy
 
     public function managedNamespaces(): array
     {
-        $namespaces = [];
-
-        foreach ($this->objectManagers() as $objectManager) {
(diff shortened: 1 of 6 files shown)


src/Mongo/MongoPersistenceStrategy.php
```diff
@@ -54,13 +54,7 @@ public function truncate(string $class): void
 
     public function managedNamespaces(): array
     {
-        $namespaces = [];
-
-        foreach ($this->objectManagers() as $objectManager) {
-            $namespaces[] = $objectManager->getConfiguration()->getDocumentNamespaces();
```
nikophil, line 60: since `doctrine/mongodb-odm-bundle` this create an error: `getDocumentNamespaces()` returns `null` but is typhinted to return an array :shrug: 

anyway, this method does not have any effect anymore in doctrine ODM :shrug: 

this means [this feature](https://github.com/zenstruck/foundry/pull/411) does not work with Mongo since a while. Not sure that this is really problematic. I'll fix it if someone notices :sweat_smile:

PR #1073, opened 2026-01-30
minor: add priorities for PHPUnit hooks

### LINKED ISSUE #1071: Bug Report: Kernel Reboot in v2.8.5+ Breaks Container State in WebTestCase
| Q                                                                                                                                  | A                       |
|------------------------------------------------------------------------------------------------------------------------------------|-------------------------|
| Foundry's version?                                                                                                                 | 2.8.5, 2.8.6                |
| PHP version                                                                                                                        | 8.5.2   |
| PHPUnit version                                                                                                                    | 12.5.8        |
| `symfony/framework-bundle` version?                                                                                                | 8.0.4      |
| Do you use [dama/doctrine-test-bundle](https://github.com/dmaicher/doctrine-test-bundle)?                                          |  no                |
| Which reset database strategy do you use?                                                                                          | migrate |
| Do you use Foundry's [PHPUnit extension](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#phpunit-extension)? |  no                |

## Summary

The kernel reboot introduced in v2.8.5 (PR #1059) breaks tests that depend on container state being preserved between test setup and HTTP request execution in `WebTestCase` scenarios. Specifically, cached translations and security context are lost after the kernel reboot.

## Versions

- **Working:** zenstruck/foundry v2.8.4
- **Broken:** zenstruck/foundry v2.8.5, v2.8.6
- **Symfony:** 8.0.4
- **PHPUnit:** 12.5.8
- **PHP:** 8.5.2

## Root Cause

PR #1059 ("minor: reboot kernel in database resetter instead of shutdown") changed `DamaDatabaseResetter` to reboot the kernel instead of shutting it down. This reboot clears the service container, losing:
- Cached data in tag-aware cache services
- Security context/tokens
- Any other stateful services

## Steps to Reproduce

1. Create a test using `WebTestCase` with `ResetDatabase` trait
2. In test setup, create entities with custom data that gets cached in the container
3. Invalidate cache tags to ensure fresh data
4. Log in a user via `$client->loginUser($user, 'firewall')`
5. Make an HTTP request that relies on:
   - The cached data from step 2
   - The security context from step 4
6. Assert that the custom cached data is used

## Expected Behavior

The test should pass. Cached data and security context should be available during the HTTP request, just as they were in v2.8.4.

## Actual Behavior

The test fails. The kernel reboot between test setup and HTTP request execution clears:
- The cache entries (even though they were invalidated and should be fresh)
- The security context

This causes the application to fall back to default behavior instead of using the test-specific data.

## Example Test Case

```php
final class InventoryExportControllerTest extends WebTestCase
{
    use ResetDatabase;
    use Factories;

    public function testManagerCanDownloadInventoryExport(): void
    {
        // 1. Create user with custom translation settings
        $consignee = ConsigneeFactory::createOne([
            'settings' => json_encode([
                'translations' => [
                    'en' => [
                        'content_admin.label.stock_id' => 'Custom Stock ID',
                    ],
                ],
            ]),
        ]);
        
        $user = UserFactory::createOne([
            'consignee' => $consignee,
            'roles' => ['ROLE_MANAGER'],
        ]);

        // 2. Flush and invalidate cache
        $entityManager->flush();
        $cache->invalidateTags([(string) $consignee->getId()]);

        // 3. Login user
        $this->client->loginUser($user, 'stock');

        // 4. Make HTTP request
        $this->client->request('GET', '/en/stock/inventory/export');

        // 5. Assert custom translation is used
        $content = $this->client->getResponse()->getContent();
        
        // ✅ Passes in v2.8.4
        // ❌ Fails in v2.8.5+ - uses default translation instead
        self::assertStringContainsString('Custom Stock ID', $content);
    }
}
```

## Application Context

We use a custom translator decorator that reads per-user translation overrides from the database and caches them.

```php
#[AsDecorator(decorates: TranslatorInterface::class)]
final readonly class ConsigneeTranslator implements TranslatorInterface, TranslatorBagInterface, LocaleAwareInterface
{
    public function __construct(
        #[AutowireDecorated]
        private TranslatorInterface&TranslatorBagInterface&LocaleAwareInterface $translator,
        private Security $security,
        private TagAwareCacheInterface $cache,
    ) {}

    public function trans(string $id, array $parameters = [], ?string $domain = null, ?string $locale = null): string
    {
        $user = $this->security->getUser();
        // ... check cache for custom translation
        // ... return custom or fallback to default
    }
}
```

The kernel reboot causes:
1. The cache to be cleared
2. The security context to be lost (no user available)

## Workaround

Lock to v2.8.4:
```json
{
    "require-dev": {
        "zenstruck/foundry": "2.8.4"
    }
}
```

## Questions

1. Is the kernel reboot necessary for the database reset functionality?
2. Can we preserve cached state and security context during the reboot?
3. Should Foundry provide a mechanism to preserve specific services/state during resets?

## Related

- PR #1059: "mnior: reboot kernel in database resetter instead of shutdown"
- PR #1070: "fix: handle reset DB mechanism along with 'before hooks' methods" (open, may be related)

nikophil:
fixes https://github.com/zenstruck/foundry/issues/1071
fixes a little bit of https://github.com/zenstruck/foundry/issues/963 :)

anyway, such problems will be definitely fixed with v2.9 and drop of the `Factories` trait

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -11,6 +11,7 @@ $file = __DIR__.'/.php-cs-fixer.temp.php';
 $csFixerConfig = require $file;
 $csFixerConfig->setFinder(
     $csFixerConfig->getFinder()
+        ->notName('WebTestCaseWithBothTraitsInWrongOrderTest.php')
         ->in(__DIR__.'/utils')
         ->in(__DIR__.'/config')
 );
--- a/bin/tools/php-cs-fixer/composer.lock
+++ b/bin/tools/php-cs-fixer/composer.lock
@@ -403,16 +403,16 @@
         },
         {
             "name": "friendsofphp/php-cs-fixer",
-            "version": "v3.89.1",
+            "version": "v3.93.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/PHP-CS-Fixer/PHP-CS-Fixer.git",
-                "reference": "f34967da2866ace090a2b447de1f357356474573"
+                "reference": "b3546ab487c0762c39f308dc1ec0ea2c461fc21a"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/PHP-CS-Fixer/PHP-CS-Fixer/zipball/f34967da2866ace090a2b447de1f357356474573",
-                "reference": "f34967da2866ace090a2b447de1f357356474573",
+                "url": "https://api.github.com/repos/PHP-CS-Fixer/PHP-CS-Fixer/zipball/b3546ab487c0762c39f308dc1ec0ea2c461fc21a",
+                "reference": "b3546ab487c0762c39f308dc1ec0ea2c461fc21a",
                 "shasum": ""
             },
             "require": {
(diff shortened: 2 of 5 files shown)


.php-cs-fixer.dist.php
```diff
@@ -11,6 +11,7 @@
 $csFixerConfig = require $file;
 $csFixerConfig->setFinder(
     $csFixerConfig->getFinder()
+        ->notName('WebTestCaseWithBothTraitsInWrongOrderTest.php')
```
nikophil, line 14: I know cs fixer released a `@cs-fixer-ignore` annotation, but I couldn't make it work  :shrug:

PR #1085, opened 2026-02-06
feat: Behat extension

### LINKED ISSUE #235: Behat Integration

### LINKED ISSUE #1051: Behat extension ideas
Here are the features I think a Behat extension for Foundry should have.

@aegypius @mpdude @mellowgrab @kevin-schmitt @roboticflamingo any thoughts about this?

Any help about the API and the naming of the `Given`, `When`, `Then` definitions would be appreciated

# 1. Create objects with PyTable

## Create one object
```gherkin
Given a post A is created with properties
    | name            | state     | published_at |
    | my awesome post | published | <now()>      |
```

> [!WARNING]
> Maybe a Gherkin definition like `a post A is created with properties` is too generic and would create conflicts with contexts from userland?

> [!NOTE]
> We need to find a way to resolve the object name to a factory class:
> "post" => `PostFactory`, "post category" => `PostCategoryFactory`
>
> Solutions could be:
> - Automagically resolve the factory name from the class's shortname, based on camelCase. If a conflict is detected, we should throw an exception or fallback on another solution
> - introduce an attribute `#[FactoryName('post')]`
> - possibility to configure this in `behat.yml`:
>   ```yml
>   extensions:
>     Zenstruck\Foundry\Test\Behat\FoundryExtension:
>       factories:
>         App\Factory\PostFactory: post
>         App\Factory\PostCategoryFactory: post category
>   ```

## Create multiple objects
```gherkin
Given posts are created with properties
    | _ref | name                  | state     | published_at |
    | A    | my awesome post A     | published | <now()>      |
    | B    | my awesome post B     | draft     | <now()>      |
```

> [!NOTE]
> We need a way to handle pluralized names:
> - use an inflector
> - Handle it in the attribute: `#[FactoryName(shortName: 'post', plural: 'posts')]`
> - Or in `behat.yml`
>   ```yml
>   extensions:
>       Zenstruck\Foundry\Test\Behat\FoundryExtension:
>         factories:
>           App\Factory\PostFactory: {shortName: post, plural: posts}
>   ```

## Can reference another object when creating an object
```gherkin
Given a post category A is created with properties
    | name                | 
    | my awesome category | 

Given a post A is created with properties
    | name            | category |
    | my awesome post | A        |
```

# 2. Create objects with natural language

This is the more challenging feature

## Create an object with zero or one property

```gherkin
Given a post A is created
```
```gherkin
Given a post A is created with name "awesome post"
```

> [!WARNING]
> This definition `a post A is created` is even more generic than the previous ones.
> I think that we should find something more specific.
> Maybe `a fixture post named A is created`? `a post A is created by Foundry`?

## Create an object and reuse it in a later step

```gherkin
Given a post A is created
Given the post A has property "name" "awesome post"
Given this post has property "state" "published"
```
 
> [!NOTE]
> This would mean that the objects are not created right away at the first line
> So we need to store the factories somewhere, and create the objects before the first `When`

## Reference an object in a later step

```gherkin
Given a post category A is created
And a post A is created
And the post A has category "post category B"
```

> [!NOTE]
> We should detect that `Post::$category` is an object, and not a string
> and then, we need to use a factory

> [!WARNING]
> In order not to create multiple categories in this example, we will need to leverage `lazy()` or `memoize()` helpers

## Be able to call factory state methods

Given this factory:
```php
class PostFactory extenbds PersistentObjectFactory
{
    public function publishedAWeekAgo(): self
    {
        return $this->with([
            'published' => true,
            'published_at' => now('-1 week'),
        ]);
    }
}
```

We should be able to reference to the "state method" in a later step:

```gherkin
Given a post A is created
Given the post A was published a week ago
```

> [!WARNING]
> I don't like this syntax, we must find something better

> [!NOTE]
> We'd also need a way to pass one or more arguments to the state method

> [!NOTE]
> The state method could be resolved automagically, 
> but we should also add a way to configure it, with an attribute `#[FactoryStateMethod('published a week ago')]`

# 3. Create objects within a Story

Given this story:
```php
class PostStory extends Story
{
    public function build(): void
    {
        $this->addState('my blog post', PostFactory::createOne());
    }
}
```

```gherkin
@withStory postStory
When I do something...
```

> [!NOTE]
> Once again, we need to resolve the story from this short name
> This could be done automagically, from the class's short name
> or we could leverage `#[AsFixture()]` attribute

# 4. Provide a way to access the created object in `When` and `Then` steps

Foundry could expose some sort of "objects registry", where we can store the created objects.

```gherkin
Given a post A is created
```

```php
use Zenstruck\Foundry\Test\Behat\ObjectsRegistry;

class FoundryContext
{
    public function __construct(
        private ObjectsRegistry $objectsRegistry
    ) {}
    
    #[Given('a :object is created')]
    public function createObject(string $objectName): void
    {
        $object = // somehow create the object with Foundry
        
        $this->objectsRegistry->store($objectName, $object);
    }
}
```

This object registry could be injected in userland contexts, and be used to easily "act" and "assert" on the created objects.

> [!NOTE]
> Objects created within a story should also be accessible from the registry

> [!NOTE]
> Of course, the object registry should handle name conflicts

> [!NOTE]
> The object registry should be reset between scenarios (or features? or on demand?)

# 5. Provide a way to make simple assertions on the created objects

```gherkin
Given a post A is created with properties
    | name            | state     | published_at |
    | my awesome post | published | <now()>      |

# this would use a userland context
When I unpublish the post A

# From Foundry extension
Then the post A property "state" is "unpublished"
Then the post A properties are now
    | state       | unpublished_at |
    | unpublished | <now()>        |
```

# 6. Reset the database

We should provide a way to reset the database between scenarios (or between features? or on demand?) the same way as 
`ResetDatabase` trait does in PHPUnit.

The database reset could be configured to run automatically, this would be configured by the extension configuration:
```yaml
extensions:
  Zenstruck\Foundry\Test\Behat\FoundryExtension:
    reset_database: before_scenario # or before_feature
```
(or maybe this should be done in the bundle's configuration?)

And for maximum flexibility, it could also be triggered manually, with a tag 
```gherkin
Feature something
    @resetDB
    Scenario do something with fresh DB
    Given / When / Then

    Scenario another scenario using the same DB
    Given / When / Then

    @resetDB
    Scenario do something with fresh DB
    Given / When / Then
```

The reset database mechanism should work with and without [dmaicher/doctrine-test-bundle](https://github.com/dmaicher/doctrine-test-bundle)

> [!WARNING]
> This will be problematic if we want the database to be reset at another pace than before every scenario, because dama
> [rollbacks the transaction after each scenario](https://github.com/dmaicher/doctrine-test-bundle/blob/master/src/Behat/BehatListener.php)
> and this is not configurable.

nikophil:
This is a "meta PR" for the whole feature of the Behat extension + context

@aegypius @loic425 @mpdude the extension is now 90% ready! It would be super nice if you could give your thoughts about it :pray: 

don't bother to review the whole massive PR, but [the docs](https://github.com/zenstruck/foundry/pull/1085/files#diff-8d068e8797e88947c320f79e856c3e16a72b730124a8f9d7031e2c4680dfa534) and [the example features](https://github.com/zenstruck/foundry/pull/1085/files#diff-4be3e32eba1357f5e5319e82c32f20c5dd94c163894a598d859b8422902b2dcb) will help you to grasp what it does.

We've finally decided to ship an external package `zenstruck/foundry-behat`, mainly because Behat does not support SF 8.0 and it would have been a PITA to maintain the other way.

Currently Foundry won't ship any way to create objects in "natural language" nor a solution to use so-called "state methods" out of the box, I gave it a try and the complexity increased drastically, but this could be a nice improvement for the future.

Another future evolution would be to handle `-ToMany` relations which are currently not handled

fixes https://github.com/zenstruck/foundry/issues/1051
fixes https://github.com/zenstruck/foundry/issues/235


src/Persistence/PersistenceManager.php
```diff
@@ -30,8 +30,9 @@
  * @author Kevin Bond <kevinbond@gmail.com>
  *
  * @internal
+ * @final
```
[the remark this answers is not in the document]
nikophil: I've done this because I now mock this class in some tests

src/Story/Event/StateAddedToStory.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+namespace Zenstruck\Foundry\Story\Event;
+
+/**
+ * @internal
+ *
+ * @template T of object
+ */
+final class StateAddedToStory
```
nikophil, line 19: I use this event to give a name to the objects created in story, so that we can access them the Behat features

src/Test/Behat/symlink-vendor.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
nikophil, line 1: this script is used to symlink `./src` into `./src/Test/Behat/vendor/zenstruck/foundry`

this improves DX a lot while working on the Behat extension

src/Test/Behat/composer.json
```diff
@@ -0,0 +1,74 @@
...
+    }
+  },
+  "require": {
+    "php": "^8.1",
+    "behat/behat": "^3.22",
+    "friends-of-behat/symfony-extension": "^2.0",
+    "symfony/polyfill-php84": "^1.33",
+    "symfony/polyfill-php85": "^1.33",
+    "symfony/string": "^6.4|^7.0",
+    "zenstruck/foundry": "dev-behat-extension"
```
nikophil, line 41: :warning:  🚨 this needs to be changed at some point

src/ZenstruckFoundryBundle.php
```diff
@@ -274,6 +275,10 @@ public function build(ContainerBuilder $container): void
         $container->addCompilerPass($this);
         $container->addCompilerPass(new InMemoryCompilerPass());
         $container->addCompilerPass(new AsFixtureStoryCompilerPass());
+
+        if (\class_exists(BehatServicesCompilerPass::class)) {
+            $container->addCompilerPass(new BehatServicesCompilerPass()); // @phpstan-ignore argument.type
+        }
```
nikophil, line 287: this is the simpler way I found to plug the extension into our bundle. And it is the only place where the namespace `Zenstruck\Foundry\Test\Behat` is used in all Foundry's core repo

src/Test/Behat/src/FoundryContext.php
```diff
@@ -0,0 +1,115 @@
...
+    public function createObject(string $factoryShortName, ?string $objectName = null): void
+    {
+        parent::createObject($factoryShortName, $objectName);
+    }
+
+    #[\Override]
+    #[Given('there is a(n) :factoryShortName with')]
+    #[Given('there is a(n) :factoryShortName :objectName with')]
+    #[Given('there is a(n) :factoryShortName called :objectName with')]
+    #[Given('there is a(n) :factoryShortName named :objectName with')]
```
jdeniau:
```suggestion
    #[Given('there is a(n) :factoryShortName with:')]
    #[Given('there is a(n) :factoryShortName :objectName with:')]
    #[Given('there is a(n) :factoryShortName called :objectName with:')]
    #[Given('there is a(n) :factoryShortName named :objectName with:')]
```

I think that with behat, it's common to finish a line with `:` when it's followed by a TableNode. 

Examples in [behat codebase](https://github.com/Behat/Behat/blob/02dffb3780cafdc6f35be1201b661d4451b63594/features/format_options.feature#L82) or in [soyuka context](https://github.com/soyuka/contexts/blob/e98b5e5a48ad5e05d0102f8a201ea86d528bcf2f/src/Context/TableContext.php#L12) or even in the [cucumber gherkin reference guide](https://cucumber.io/docs/gherkin/reference#data-tables).

src/Test/Behat/src/FoundryContext.php
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * @phpstan-import-type Parameters from Factory
+ *
+ * @internal
+ * @final
+ */
+class FoundryContext extends AbstractFoundryContext implements Context
+{
+    #[\Override]
+    #[Given('there is a(n) :factoryShortName')]
```
jdeniau: Won't those strings be too generic ? It may override, or be overridden by other contexts ? 
Shouldn't it contain a reference to foundry (or at least "fixture") ?

nikophil: this gave me some headaches :sweat_smile: 

it may not be really problematic to introduce the word "fixture" or "foundry" in the creation steps, but it would be very counterintuitive to have them in the assertion steps.

At the end, my decision has been to:
- keep the steps like this
- split the contexts between `FoundryCreationContext` / `FoundryAssertionContext` / `FoundryPlaceholderContext` - users could decide to take some and not the other
- provide an easy way to override those steps

About the override, my first thought was to use `TranslatableContext` (which are documented no where :roll_eyes:), this works well but PHPStorm nor VS Code understand it, which makes it a no-go :shrug: 

anyway, this extension will be released (very soon! :crossed_fingers: :innocent:) under a `0.x` tag, so there will still be room for breaking stuff.

PR #1091, opened 2026-02-08
feat: Foundry v3 🎉

nikophil:
_I will wait for the Behat extension to be released before releasing 3.x_

:star_struck: 
image

sorry for the massive PR but it was really easier for me. I made Claude work on this and ~he~ it did a very good job for 90% of this boring stuff


nikophil:
actually I will split this is multiple PRs :sweat_smile:

PR #1109, opened 2026-03-24
feat: add `withoutDoctrineEvents()` to suppress listeners during factory create

### LINKED ISSUE #957: [RFC] allow factories to disable doctrine events
I think this could be very nice addition, there are valid use cases for this

someone tried to do it in Foundry 1, but the PR was never finished

https://github.com/zenstruck/foundry/pull/216

seb-jean:
## Motivation

When Foundry factories create entities during fixture loading, Doctrine event listeners attached to those entities fire as usual. This can cause unwanted side effects in tests — for example, sending emails, dispatching messages, or calling external services on every persist.

Related #216.

## What this PR adds

A `withoutDoctrineEvents()` method on `PersistentObjectFactory` that temporarily removes Doctrine event listeners from the `EventManager` around the `flush()` call, then restores them immediately after.

## API

```php
// Disable ALL Doctrine event listeners during create()
PostFactory::new()->withoutDoctrineEvents()->create([...]);

// Disable specific listeners only
PostFactory::new()
    ->withoutDoctrineEvents(MyEventListener::class)
    ->create([...]);
```

I have 2 failing checks, but both are pre-existing issues unrelated to this PR.

fixes #957

### DIFF
--- a/src/Persistence/PersistenceManager.php
+++ b/src/Persistence/PersistenceManager.php
@@ -11,6 +11,8 @@
 
 namespace Zenstruck\Foundry\Persistence;
 
+use Doctrine\Common\EventManager;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\ObjectManager;
 use Doctrine\Persistence\ObjectRepository;
@@ -40,6 +42,24 @@ final class PersistenceManager
     private array $afterPersistCallbacks = [];
 
     /**
+     * Event listener classes to keep disabled until the next real flush, accumulated across
+     * multiple scheduleForInsert() calls inside a flush_after() context.
+     * Keyed by spl_object_id of the ObjectManager.
+     *
+     * @var array<int, list<class-string>|null>
+     */
+    private array $pendingEventClassesToDisable = [];
+
+    /**
+     * Entity-listener overrides to apply during the next real flush, accumulated across
+     * multiple scheduleForInsert() calls inside a flush_after() context.
+     * Keyed by spl_object_id of the ObjectManager.
+     *
+     * @var array<int, list<array{entityClass: class-string, disabledClasses: list<class-string>|null}>>
+     */
+    private array $pendingEntityListenerOverrides = [];
+
+    /**
      * @param iterable<PersistenceStrategy> $strategies
      */
     public function __construct(
@@ -66,19 +86,34 @@ final class PersistenceManager
     /**
      * @template T of object
      *
-     * @param T $object
+     * @param T                       $object
+     * @param list<class-string>|null $disabledDoctrineEventClasses null = no events disabled, [] = all disabled, [Foo::class] = specific classes disabled
      *
      * @return T
      */
-    public function save(object $object): object
+    public function save(object $object, ?array $disabledDoctrineEventClasses = null): object
     {
         if ($object instanceof Proxy) {
             return $object->_save();
         }
 
         $om = $this->strategyFor($object::class)->objectManagerFor($object::class);
+
+        // Disable for prePersist (fires during persist())
+        // Note: disableEntityListeners must be called first, as it calls getClassMetadata() which
+        // triggers the loadClassMetadata event to register #[AsEntityListener] listeners. If we
+        // disabled global events first, that event would fire without the EntityListenerRegistry
+        // handler, leaving entityListeners permanently empty in the metadata cache.
+        $entityListenerBackup = $this->disableEntityListeners($om, $object::class, $disabledDoctrineEventClasses);
+        $removedListeners = $this->disableDoctrineEvents($om, $disabledDoctrineEventClasses);
+
         $om->persist($object);
-        $this->flush($om);
+
+        $this->restoreDoctrineEvents($om, $removedListeners);
+        $this->restoreEntityListeners($om, $object::class, $entityListenerBackup);
+
+        // Disable for postPersist / preUpdate (fire during flush())
+        $this->flush($om, $disabledDoctrineEventClasses);
 
         $shouldFlush = $this->callPostPersistCallbacks();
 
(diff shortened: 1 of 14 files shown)


src/Persistence/PersistentObjectFactory.php
```diff
@@ -306,6 +323,16 @@ final public function withoutAutorefresh(): static
...
 
+    /**
+     * @return list<class-string>|null
+     *
+     * @internal
+     */
+    public function getDisabledDoctrineEventClasses(): ?array
+    {
+        return $this->disabledDoctrineEventClasses;
+    }
```
nikophil: this getter is not needed

src/Persistence/PersistentObjectFactory.php
```diff
@@ -541,7 +572,7 @@ static function(object $object, array $parameters, PersistentObjectFactory $fact
                         };
                     }
 
-                    Configuration::instance()->persistence()->scheduleForInsert($object, $afterPersistCallbacks);
+                    Configuration::instance()->persistence()->scheduleForInsert($object, $afterPersistCallbacks, $factoryUsed->getDisabledDoctrineEventClasses());
```
nikophil:
```suggestion
                    Configuration::instance()->persistence()->scheduleForInsert($object, $afterPersistCallbacks, $factoryUsed->disabledDoctrineEventClasses);
```

src/Persistence/PersistenceManager.php
```diff
@@ -469,6 +480,56 @@ private function callPostPersistCallbacks(): bool
...
     }
 
+    /**
+     * Temporarily removes Doctrine event listeners from the EventManager.
+     *
+     * @param list<class-string>|null $disabledClasses null = nothing removed, [] = all removed, [Foo::class] = specific classes removed
+     *
+     * @return array<string, list<object>> map of eventName => removed listeners, for later restoration
+     */
+    private function disableDoctrineEvents(ObjectManager $om, ?array $disabledClasses): array
```
nikophil:
```suggestion
    private function disableDoctrineEvents(ObjectManager $om, ?array $eventClassesToDisabled): array
```
?

tests/Fixture/DoctrineEvents/DoctrineEventsSubscriber.php
```diff
@@ -0,0 +1,34 @@
...
+
+use Doctrine\Bundle\DoctrineBundle\Attribute\AsDoctrineListener;
+use Doctrine\ORM\Event\PrePersistEventArgs;
+use Doctrine\ORM\Events;
+use Zenstruck\Foundry\Tests\Fixture\Entity\EntityForDoctrineEvents;
+
+#[AsDoctrineListener(event: Events::prePersist)]
+final class DoctrineEventsSubscriber
+{
+    public function prePersist(PrePersistEventArgs $eventArgs): void
```
nikophil, line 26: maybe add a `preUpdate()` event as well? as we don't disable "pre/post persist" listeners at the same place then the others

src/Persistence/PersistenceManager.php
```diff
@@ -66,20 +67,26 @@ public function enablePersisting(): void
...
         if ($object instanceof Proxy) {
             return $object->_save();
         }
 
         $om = $this->strategyFor($object::class)->objectManagerFor($object::class);
+
+        $removedListeners = $this->disableDoctrineEvents($om, $disabledDoctrineEventClasses);
+
         $om->persist($object);
         $this->flush($om);
```
nikophil: in order to make this feature work with `flush_after()`, I guess you'll need to disable the doctrine events in `PersistenceManager::flush()`

nikophil (commented):
hi @seb-jean 

thanks to work on this! I have the need for this feature from time to time

Could you please add some more test cases:
- one using `#[ORM\EntityListeners]` on the entity
- one using `#[AsEntityListener]` on a listener
- one using [`flush_after()`](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#delay-flush) (I think you'll have to change some code to make it work)

Here are some questions, maybe for the future:
- is it possible to make it work with `#[ORM\HasLifecycleCallbacks]`? 
- do you think it's possible to disable all listeners for a given event for an entity? meaning something like `->disableDoctrineEvents('prePersist')`

src/Persistence/PersistentObjectFactory.php
seb-jean: I fixed

src/Persistence/PersistentObjectFactory.php
seb-jean: I fixed

src/Persistence/PersistenceManager.php
seb-jean: I fixed

tests/Fixture/DoctrineEvents/DoctrineEventsSubscriber.php
seb-jean: I fixed

src/Persistence/PersistenceManager.php
seb-jean: I fixed

nikophil (approved):
thanks @seb-jean 

the code to make this work with `flush_after()` is really too complex, moreover it is problematic because such thing won't work:
```php
flush_after(
    function() {
        PostFactory::new()->withoutDoctrineEvents()->create();
        PostFactory::new()->create();
    }
);
```
both objects will be created without the events.

All in all, I prefer that we rollback this, and throw if `withoutDoctrineEvents()` is used within `flush_after()`

I'm going to make another PR afterwards before releasing

VincentLanglet:
Hi @nikophil could you make a release with this ? Thanks

nikophil:
@VincentLanglet yeha, I'll do it very soon

nikophil:
@VincentLanglet done

VincentLanglet:
Thanks

PR #1110, opened 2026-03-26
fix: factory template formatting

enekochan:
Because of how php tag works the output of `make:factory` generates some "not so pretty" PHP code. It lacks some carriage returns that place `Override` attributes where they shouldn't be and has more spaces than needed in some places.

Generated code:

```php
<?php

namespace App\Factory;

use App\Entity\User;
use App\Repository\UserRepository;
use Doctrine\ORM\EntityRepository;
use Zenstruck\Foundry\Persistence\PersistentObjectFactory;
use Zenstruck\Foundry\Persistence\Proxy;
use Zenstruck\Foundry\Persistence\ProxyRepositoryDecorator;

/**
 * @extends PersistentObjectFactory<User>
 */
final class UserFactory extends PersistentObjectFactory{
    /**
     * @see https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#factories-as-services
     *
     * @todo inject services if required
     */
    public function __construct()
    {
    }

    #[\Override]    public static function class(): string
    {
        return User::class;
    }

        /**
     * @see https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#model-factories
     *
     * @todo add your default values here
     */
    #[\Override]    protected function defaults(): array|callable    {
        return [
            'name' => self::faker()->text(10),
        ];
    }

        /**
     * @see https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#initialization
     */
    #[\Override]    protected function initialize(): static
    {
        return $this
            // ->afterInstantiate(function(User $user): void {})
        ;
    }
}

```

### DIFF
--- a/skeleton/Factory.tpl.php
+++ b/skeleton/Factory.tpl.php
@@ -25,7 +25,7 @@ if (count($makeFactoryData->getMethodsPHPDoc())) {
 }
 ?>
  */
-final class <?php echo $class_name; ?> extends <?php echo $makeFactoryData->getFactoryClassShortName(); ?>
+final class <?php echo $class_name; ?> extends <?php echo $makeFactoryData->getFactoryClassShortName(); ?><?= "\n" ?>
 {
 <?php if ($makeFactoryData->shouldAddHints()): ?>    /**
      * @see https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#factories-as-services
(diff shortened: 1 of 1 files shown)


nikophil:
Hi!

this seems weird, because we're testing the output the generated classes
see the test class https://github.com/zenstruck/foundry/blob/2.x/tests/Integration/Maker/MakeFactoryTest.php
and one of the expected output https://github.com/zenstruck/foundry/blob/2.x/tests/Fixture/Maker/expected/can_create_factory.php

which maker bundle version are you running? or maybe it could come from the OS you're using? we're only running our CI on ubuntu

enekochan:
I'm using Maker Bundle 1.62.1, in macOS but with a Docker environment with Alpine, PHP 8.4 and Symfony 6.4. Anyway I tried an example in onlinephp and got the same output:

https://onlinephp.io/c/0bb924

Only if I add an extra carriage return after `<?php endif ?>` or use the proposed `<?= "\n" ?>` the output is as spected. Probably some option in php.ini changes the way the output is generated?

skeleton/Factory.tpl.php
```diff
@@ -34,21 +34,22 @@ final class <?php echo $class_name; ?> extends <?php echo $makeFactoryData->getF
      */
     public function __construct()
     {
+        parent::__construct();
```
nikophil: I think if you remove this line, the tests will pass

nikophil:
> Only if I add an extra carriage return after <?php endif ?> or use the proposed <?= "\n" ?> the output is as spected. Probably some option in php.ini changes the way the output is generated?

I don't think, we don't use any custom php config in the CI, and tests are passing with the current code. They also pass on my machine, with ubuntu + basic php installation

anyway, when the CI will be green on this PR (see my other comment), no problem to merge it

nikophil:
hmmm I think the test do pass because maker bundle runs csfixer on the created files :sweat_smile:  

I don't know why it does not apply in your stack: the maker bundle is actually shipped with an embedded version of cs fixer, so it should be used even if it is not directly installed in the project.

anyway, this makes your changes even more legitimate :+1:

nikophil:
thanks @enekochan

PR #1130, opened 2026-07-09
Fix withoutDoctrineEvents with nested entities

### LINKED ISSUE #1129: [BUG] withoutDoctrineEvents doesn't work with relation.
| Foundry's version?                                             | latest                                     |
| PHP version                                                         | 8.5             |
| PHPUnit version                                                 | 13                    |
| `symfony/framework-bundle` version?          | 8.1                      |
| Do you use dama/doctrine-test-bundle?       | yes                               |
| Which reset database strategy do you use? | none |
| Do you use Foundry's PHPUnit extension?   | yes                               |

Hi @seb-jean @nikophil 

Let's say I have a listener on User
```
#[AsEntityListener(event: Events::postPersist, method: 'postPersistUser', entity: User::class)]
```

The UserFactory
```
    {
        return $this
            ->withoutDoctrineEvents(MyListener::class);
    }
```
works fine to disable the Listener when I do `UserFactory::createOne()`.

But seems like it doesn't work for a related factory UserRoleFactory
```
    {
        return [
            'user' => UserFactory::new(),
            'role' => 'admin',
        ];
    }
```
even with `->withoutDoctrineEvents(MyListener::class)` in the initialize.

It does ignore the
```
#[AsEntityListener(event: Events::postPersist, method: 'postPersistUserRole', entity: UserRole::class)]
```
part, but still run the postPersistUser method.

VincentLanglet:
Closes https://github.com/zenstruck/foundry/issues/1129

Reproducer with failing tests https://github.com/zenstruck/foundry/pull/1131

It was hard to reproduce my issue, seems like it was related to postPersist event.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@
 /tests/Fixture/Maker/tmp/
 /.phpbench
 /config/reference.php
+/.claude/settings.local.json
--- a/src/ORM/AbstractORMPersistenceStrategy.php
+++ b/src/ORM/AbstractORMPersistenceStrategy.php
@@ -26,6 +26,14 @@ use Zenstruck\Foundry\Persistence\PersistenceStrategy;
  */
 abstract class AbstractORMPersistenceStrategy extends PersistenceStrategy
 {
+    private int $withoutDoctrineEventsDepth = 0;
+
+    /** @var array<string, list<object>> */
+    private array $pendingGlobalListenerRestoration = [];
+
+    /** @var array<class-string, array<string, list<array{class: class-string, method: string}>>> */
+    private array $pendingEntityListenerRestoration = [];
+
     final public function contains(object $object): bool
     {
         $em = $this->objectManagerFor($object::class);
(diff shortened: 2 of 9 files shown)


nikophil:
Hi @VincentLanglet 

this whole feature of disabling the doctrine events appeared to be much more complex than I thought :sweat_smile: 

I've refactored some stuff in https://github.com/zenstruck/foundry/pull/1132, in order to make the feature less brittle. The refacto fixes your problem, and I've reused your test case there.
 
thank you for your work, though.

VincentLanglet:
Hi @nikophil ; I can confirm your fix works on my projet. Thanks ! 
A patch release would be great when you'll have time :)

nikophil:
@VincentLanglet yeah, I just wanted to merge [another fix](https://github.com/zenstruck/foundry/pull/1133) before releasing

https://github.com/zenstruck/foundry/releases/tag/v2.10.2 is released!

PR #1134, opened 2026-07-19
fix: defer `$em->persist()` until the object graph is wired + natural accessor directionality

### LINKED ISSUE #1100: [BUG] Cascade persist order

### LINKED ISSUE #1115: [BUG] Orphan removal is triggered (adder not called)
| Foundry's version?                                             | 2.9.2                                   |
| PHP version                                                         | 8.5                                       |
| PHPUnit version                                                 | 13                                        |
| `symfony/framework-bundle` version?          | 7.4                                       |
| Do you use dama/doctrine-test-bundle?       | yes                                       |
| Which reset database strategy do you use? |  migrate                              |
| Do you use Foundry's PHPUnit extension?   | yes                                       |

In case of relations with `oprhanRemoval: true`, if the adder is not called, the children are missing.

After the call of the Factory, the object is full, with all children, but in the database children are gone.
Removing orphanRemoval properties make them appear in database too. 

So I mean, because of the adder is not called, the parent have an empty Collection, then Doctrine remove them.

```php
// This class is a child (its parent have a Collection of it)
class SubParent {
    #[ORM\OneToMany(targetEntity: Child::class, mappedBy: 'children', cascade: ['persist', 'remove']], oprhanRemoval: true)]
    private Collection $children;

    /**
     * @return ReadableCollection<int, Child>
     */
    public function getChildren(): ReadableCollection
    {
        return $this->children;
    }

    public function addChild(Child $child): static
    {
        if (!$this->children->contains($child)) {
            $this->children->add($child);
            $child->setParent($this);
        }

        return $this;
    }

    public function removeChild(Child $child): static
    {
        if ($this->children->removeElement($child) && $child->getParent() === $this) {
            $child->setParent(null);
        }

        return $this;
    }
}

```

nikophil:
Fixes zenstruck/foundry#1100
Fixes zenstruck/foundry#1115
Supersedes zenstruck/foundry#1104 (its fixtures/tests ideas are reused, thanks @mpiot for the feedback)

`em->persist()` was called synchronously while the object graph was still being built:

- a parent's `PrePersist` listeners always observed an **empty** one-to-many collection, whatever the creation direction (#1100);
- combined with `orphanRemoval: true` and listeners triggering extra flushes, children rows could even be **deleted from the database** (#1115);
- nested factories (`ChildFactory::createOne(['parent' => ParentFactory::new()])`) triggered their own mid-graph flush.

ping @mpiot any chance that you test this PR please?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -665,6 +665,24 @@ they were added, or by priority (higher priority hooks are executed first).
 
     Hook priority were added in Foundry 2.8.
 
+``beforeInstantiate`` and ``afterInstantiate`` hooks run **before** the object graph is handed over to the
+object manager: state set by Doctrine lifecycle listeners (e.g. a field computed in a ``PrePersist``
+listener) is not yet available, even for objects created by a nested factory. To observe such state,
+register an ``afterInstantiate`` hook with a priority lower than
+``PersistentObjectFactory::PRIORITY_SCHEDULE_FOR_INSERT``: it runs once the whole object graph has been
+persisted (but not flushed yet):
+
+::
+
+    use Zenstruck\Foundry\Persistence\PersistentObjectFactory;
+
+    PostFactory::new()
+        ->afterInstantiate(function(Post $post): void {
+            // the "pre persist" events of the whole object graph have been triggered:
+            // any state they computed is visible here
+        }, priority: PersistentObjectFactory::PRIORITY_SCHEDULE_FOR_INSERT - 1)
+    ;
+
 You can also add hooks directly in your factory class:
 
 ::
--- a/src/FactoryCollection.php
+++ b/src/FactoryCollection.php
@@ -218,6 +218,24 @@ final class FactoryCollection implements \IteratorAggregate
     }
 
     /**
+     * @param callable(TFactory):TFactory $callback
+     *
+     * @return self<T, TFactory>
+     *
+     * @internal
+     */
+    public function map(callable $callback): self
+    {
+        $factories = $this->all();
+
+        if ([] === $factories) {
+            return $this;
+        }
+
+        return new self($this->factory, static fn() => \array_map($callback, $factories));
+    }
+
+    /**
      * @internal
      */
     public function reuse(object ...$objects): static
@@ -226,15 +244,7 @@ final class FactoryCollection implements \IteratorAggregate
             return $this;
         }
 
-        $factories = $this->all();
-
-        return new self(
-            $this->factory,
-            static fn() => \array_map(
-                static fn(Factory $f) => $f instanceof ObjectFactory ? $f->reuse(...$objects) : $f,
-                $factories,
-            )
-        );
+        return $this->map(static fn(Factory $f) => $f instanceof ObjectFactory ? $f->reuse(...$objects) : $f);
     }
 
     /**
--- a/src/Object/Hydrator.php
+++ b/src/Object/Hydrator.php
@@ -51,13 +51,13 @@ final class Hydrator
                 continue;
             }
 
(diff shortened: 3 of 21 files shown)


mpiot:
@nikophil I've tried your branch on the project: prePersist events work as expected (actually onFlush to bypass, but refactor them on prePersist (just as they were) and it works) (#1100 )

I don't remember exactly the case where (#1115 ) fired in our codebase, but no error actually (I think it was somewhere in our EventListener to).

Juste a minor failure (can be a regression for some users), but easily solved in our case (and appear as normal):

```php
final class ProductionOrderItemFactory extends PersistentObjectFactory
{
    public function  withProduct(Product $product): array
    {
        return [
            'productionOrderLine' => ProductionOrderLineFactory::new([
                'productionOrder' => ProductionOrderFactory::new([
                    'product' => $product,
                ]),
            ]),
        ]);
    }

    {
        return $this
            ->afterInstantiate(static function (ProductionOrderItem $productionOrderItem): void {
                $productionOrder = $productionOrderItem->getProductionOrderLine()->getProductionOrder();
                $productionOrderItem
                    ->setProduct($productionOrder->getProduct())
                     //->setProcess($productionOrder()->getProcess()) <= Define automatically on prePersist, but not yet defined: is NULL
                    ->setProcess($productionOrder->getAcoustic()->getProcess()) // Workaround, not wait the PO prePersist call, because no cascade. In he app, we create it in the right order
                ;
            })
        ;
    }
```

nikophil:
thank you for you feedback. Yes Indeed this is a small BC break

this could even be mitigated by using a priority on your `afterInstantiate()`:
```php
    {
        return $this
            ->afterInstantiate(static function (ProductionOrderItem $productionOrderItem): void {
                $productionOrder = $productionOrderItem->getProductionOrderLine()->getProductionOrder();
                $productionOrderItem
                    ->setProduct($productionOrder->getProduct())
                    ->setProcess($productionOrder()->getProcess())
                ;
            }, PersistentObjectFactory::PRIORITY_SCHEDULE_FOR_INSERT - 1)
        ;
    }
```

This priority is now documented (it exists since https://github.com/zenstruck/foundry/pull/1029 but was not documented).

I think this small BC break is tolerable (quite edgy and Foundry is just a dev/test package) - I'll document it in the release note

PR #1138, opened 2026-07-29
feat: load multiple groups/stories with foundry:load-fixtures

podoko:
## Why

When setting up an environment, one might need to load multiple stories or groups of stories. It might not be possible to create a "meta-story" with all the stories needed for a specific test case and calling the command multiple times with `--append` might lead to duplicates if two of the stories depend on a common third one. 

## What 

I figured out `LoadFixturesCommand` alreay had all the elements needed to load multiple stories so I made the "name" argument an `array<string>`

The same system that prevents a story to be called multiple times in a process allows us to load multiple groups without having the stories present in multiple groups being called multiple times.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1696,6 +1696,19 @@ You can also load stories by group, by using the ``groups`` option:
 
 ``bin/console foundry:load-fixtures all`` will load both stories ``CategoryStory`` and ``PostStory``.
 
+.. versionadded:: 2.12
+
+    The ability to load several stories or groups at once was added in 2.12.
+
+Several names can be passed at once, and story names and group names can be mixed:
+
+.. code-block:: terminal
+
+    $ php bin/console foundry:load-fixtures category post
+    $ php bin/console foundry:load-fixtures all other-group category
+
+The database is only reset once, and a story belonging to several of the given groups is only loaded once.
+
 .. tip::
 
     It is possible to call a story inside another story, by using ``OtherStory::load();``. Because the stories are only
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1696,10 +1696,6 @@ You can also load stories by group, by using the ``groups`` option:
 
 ``bin/console foundry:load-fixtures all`` will load both stories ``CategoryStory`` and ``PostStory``.
 
-.. versionadded:: 2.12
-
-    The ability to load several stories or groups at once was added in 2.12.
-
 Several names can be passed at once, and story names and group names can be mixed:
 
 .. code-block:: terminal
@@ -1708,6 +1704,10 @@ Several names can be passed at once, and story names and group names can be mixe
     $ php bin/console foundry:load-fixtures all other-group category
 
 The database is only reset once, and a story belonging to several of the given groups is only loaded once.
+
+.. versionadded:: 2.12
+
+    The ability to load several stories or groups at once was added in 2.12.
 
 .. tip::
 
--- a/src/Command/LoadFixturesCommand.php
+++ b/src/Command/LoadFixturesCommand.php
@@ -65,10 +65,16 @@ final class LoadFixturesCommand extends Command
             $this->resetDatabase();
         }
 
-        /** @var array<string> $fixtureNamesOrGroups */
+        /** @var array<string>|string|null $fixtureNamesOrGroups */
         $fixtureNamesOrGroups = $input->getArgument('name');
 
-        if ([] === $fixtureNamesOrGroups) {
+        if (\is_string($fixtureNamesOrGroups)) {
+            trigger_deprecation('zenstruck/foundry', '2.12', 'Passing a string as the "name" argument of the "foundry:load-fixtures" command is deprecated and will throw an error in Foundry 3: pass an array of names instead.');
+
+            $fixtureNamesOrGroups = [$fixtureNamesOrGroups];
+        }
+
+        if (!$fixtureNamesOrGroups) {
             $fixtureNamesOrGroups = [$this->getNameWhenNotProvided($io)];
         }
 
--- a/src/Command/LoadFixturesCommand.php
+++ b/src/Command/LoadFixturesCommand.php
@@ -80,10 +80,13 @@ final class LoadFixturesCommand extends Command
 
         $resolvedStories = [];
         foreach (\array_unique($fixtureNamesOrGroups) as $fixtureNameOrGroup) {
-            $resolvedStories[] = [$fixtureNameOrGroup, $this->fixtureStoryResolver->resolve($fixtureNameOrGroup)];
+            $resolvedStories[$fixtureNameOrGroup] = $this->fixtureStoryResolver->resolve($fixtureNameOrGroup);
         }
 
-        foreach ($resolvedStories as [$fixtureNameOrGroup, $stories]) {
+        $loadedStoryClasses = [];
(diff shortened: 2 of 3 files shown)


src/Command/LoadFixturesCommand.php
```diff
@@ -65,21 +65,37 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                $io->comment("Loading stories group \"{$fixtureNameOrGroup}\"...");
+            }
+
+            foreach ($stories as $name => $storyClass) {
+                $storyClass::load();
 
-            if ($io->isVerbose()) {
-                $io->info("Story \"{$storyClass}\" loaded (name: {$name}).");
+                if ($io->isVerbose()) {
+                    $io->info("Story \"{$storyClass}\" loaded (name: {$name}).");
```
podoko, line 107: Here, if two groups include the same story, you might have two notifications that it was loaded. I wondered if we should keep track of the loaded stories in the command in order to avoid it but it felt like a duplicate of the system that prevents a story to be loaded twice. 

I can add it though if you think we should

docs/index.rst
```diff
@@ -1696,6 +1696,19 @@ You can also load stories by group, by using the ``groups`` option:
 
 ``bin/console foundry:load-fixtures all`` will load both stories ``CategoryStory`` and ``PostStory``.
 
+.. versionadded:: 2.12
+
+    The ability to load several stories or groups at once was added in 2.12.
```
nikophil: usually we put `.. versionadded` admonition after the feature

docs/index.rst
```diff
@@ -1696,6 +1696,19 @@ You can also load stories by group, by using the ``groups`` option:
...
+
+Several names can be passed at once, and story names and group names can be mixed:
+
+.. code-block:: terminal
+
+    $ php bin/console foundry:load-fixtures category post
+    $ php bin/console foundry:load-fixtures all other-group category
+
+The database is only reset once, and a story belonging to several of the given groups is only loaded once.
```
nikophil, line 1706: 👍

src/Command/LoadFixturesCommand.php
```diff
@@ -65,21 +65,37 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $this->resetDatabase();
         }
 
-        $fixtureNameOrGroup = $input->getArgument('name') ?? $this->getNameWhenNotProvided($io);
+        /** @var array<string>|string|null $fixtureNamesOrGroups */
+        $fixtureNamesOrGroups = $input->getArgument('name');
 
-        $stories = $this->fixtureStoryResolver->resolve($fixtureNameOrGroup);
+        if (\is_string($fixtureNamesOrGroups)) {
+            trigger_deprecation('zenstruck/foundry', '2.12', 'Passing a string as the "name" argument of the "foundry:load-fixtures" command is deprecated and will throw an error in Foundry 3: pass an array of names instead.');
```
nikophil, line 72: I'm not sure this deprecation is needed: 
```bash
# the command used to be used like this:
➜ bin/console foundry:load-fixtures some_fixture

# now we can do:
➜ bin/console foundry:load-fixtures fixture1 fixture2

# but we still can do:
➜ bin/console foundry:load-fixtures some_fixture
```

so it does not seem there is a BC break here, WDYT?

`$input->getArgument('name');` will always return an array even if only one name is passed

I would even say that we can safely rename the argument to `names`

src/Command/LoadFixturesCommand.php
```diff
@@ -65,21 +65,37 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-        foreach ($stories as $name => $storyClass) {
-            $storyClass::load();
+        if (!$fixtureNamesOrGroups) {
+            $fixtureNamesOrGroups = [$this->getNameWhenNotProvided($io)];
+        }
+
+        $resolvedStories = [];
+        foreach (\array_unique($fixtureNamesOrGroups) as $fixtureNameOrGroup) {
+            $resolvedStories[] = [$fixtureNameOrGroup, $this->fixtureStoryResolver->resolve($fixtureNameOrGroup)];
+        }
```
nikophil, line 84:
```suggestion
        foreach (\array_unique($fixtureNamesOrGroups) as $fixtureNameOrGroup) {
            $resolvedStories[$fixtureNameOrGroup] = $this->fixtureStoryResolver->resolve($fixtureNameOrGroup);
        }
```
seems easier to grasp

src/Command/LoadFixturesCommand.php
nikophil: I think that we should display a warning (`$io->warning()`)

something like: "story {$storyClass} (name: {$name}) already loaded. skipping..."

tests/Integration/Command/LoadFixturesCommandTest.php
```diff
@@ -55,7 +56,7 @@ public function it_throws_if_story_does_not_exist(): void
         $this->expectException(FixtureStoryNotFound::class);
         $this->expectExceptionMessage('Fixture story with name or group "invalid-name" not found');
 
-        $this->commandTester(['environment' => 'stories_as_fixtures'])->execute(['name' => 'invalid-name', '--append' => true]);
+        $this->commandTester(['environment' => 'stories_as_fixtures'])->execute(['name' => ['invalid-name'], '--append' => true]);
```
nikophil, line 59: maybe those don't need to be changed?

nikophil (changes requested):
super nice, thanks!

here are a few minor comments.

Don't bother with "Backward compatibility check" which is failing, I have some troubles with it

src/Command/LoadFixturesCommand.php
podoko: I didn't rename the argument to `names` and made it still accept `string|null` because my use case is actually to call the command programmatically similarly to how it is done in the tests. It would break my code. I believe (hope) this is not so uncommon, that's why I treated it as a BC

tests/Integration/Command/LoadFixturesCommandTest.php
podoko: Tried it, `::execute()` doesn't transform the string argument to an array : `$input->getArgument('names')` ends up being a string

src/Command/LoadFixturesCommand.php
nikophil: Out of curiosity, what is your use case to call this programmatically?

Ok then, let's treat it as a BC break

tests/Integration/Command/LoadFixturesCommandTest.php
nikophil: Yeah let's close this

src/Command/LoadFixturesCommand.php
podoko: When setting up a test environment, I have a bigger command that loads fixtures from multiple sources. The foundry ones are some of them. Also it allows me to load them in a single transaction (which is an other issue I saw in the repo).

nikophil: Ok thanks I naively thought evrybody uses command only in cli 😅

Don't hesitate to open an issue about your transaction problem

src/Command/LoadFixturesCommand.php
podoko: Helo, it is done line 97 to 101

nikophil (approved):

nikophil:
thank you @podoko

PR #1149, opened 2026-08-03
fix: hydrate snapshots with deepclone_hydrate() (requires symfony/polyfill-deepclone ^1.42)

### LINKED ISSUE #1148: [BUG] 2.11.1: `Cannot access property starting with "\0"` — `hydrateFromSnapshot()` passes `(array)`-mangled keys to `deepclone_hydrate()`
## Summary

`Hydrator::hydrateFromSnapshot()` (introduced in 2.11.1 by #1146) captures object state with an
`(array) $object` cast and feeds the result straight to `deepclone_hydrate()`. Those two APIs use
**different key conventions**, so hydration fails with a fatal `Error` as soon as every property in
the snapshot belongs to the object's own class.

This is a regression in 2.11.1 — 2.11.0 is unaffected.

It surfaces whenever an autorefreshed object is **deleted** (or its row disappears through a
rollback) and the tracked lazy ghost is then initialized: `autorefresh()` takes the
"object no longer exists" branch and calls `hydrateFromSnapshot()`.

Every functional test that deletes an entity failed — 6 in total, across the two test suites that
reached the test phase — on both PHP 8.4 and PHP 8.5.

## Environment

| Package | Version |
|---|---|
| `zenstruck/foundry` | **2.11.1** (2.11.0 OK) |
| `doctrine/orm` | 3.6.7 |
| `symfony/var-exporter` | v8.1.1 |
| `symfony/polyfill-deepclone` | v1.40.0 |
| PHP | 8.4.22 locally, 8.4.17 and 8.5 in CI (both fail) |

```yaml
zenstruck_foundry:
    persistence:
        flush_once: true
```

## Stack trace

```
Error: Cannot access property starting with "\0"

vendor/symfony/polyfill-deepclone/DeepClone.php:2083
vendor/symfony/polyfill-deepclone/DeepClone.php:481
vendor/symfony/polyfill-deepclone/bootstrap81.php:34
vendor/zenstruck/foundry/src/Object/Hydrator.php:219                   <- deepclone_hydrate(...)
vendor/zenstruck/foundry/src/Persistence/PersistenceManager.php:215    <- Hydrator::hydrateFromSnapshot(...)
vendor/zenstruck/foundry/src/Persistence/PersistedObjectsTracker.php:119
```

The two remaining frames are application code: any getter on the deleted entity (which triggers the
lazy ghost initialization), called from the test asserting the deletion.

## Minimal reproduction — no Foundry, no lazy ghost, no database

```php
final class Solo
{
    private int $x = 1;
}

deepclone_hydrate(new Solo(), (array) new Solo(), \DEEPCLONE_HYDRATE_PRESERVE_REFS);
// Error: Cannot access property starting with "\0"
```

Same result through `Symfony\Component\VarExporter\Hydrator::hydrate()` (it delegates to
`deepclone_hydrate()` on this version), so the `function_exists('deepclone_hydrate')` branch in
`hydrateFromSnapshot()` is **not** a workaround — both branches end in the same call.

## Root cause

`(array) $object` produces PHP's **serialization** mangling for non-public properties:

```php
final class Leaf extends Base
{
    private int $leafPrivate = 1;
    protected string $leafProtected = 'l';
    public string $leafPublic = 'p';
}

// (array) $leaf  =>
//   "\0Base\0basePrivate"   (private inherited from a parent)
//   "\0*\0baseProtected"    (protected)
//   "\0Leaf\0leafPrivate"   (private on the object's own class)
//   "\0*\0leafProtected"    (protected)
//   "leafPublic"            (public)
```

But `deepclone_hydrate()` expects **VarExporter's** convention, which `VarExporter\Hydrator::hydrate()`
builds explicitly:

```php
$mangledVars[$isOwnScope ? $name : "\0$scope\0$name"] = &$value;
```

That is: a **plain name** for the object's own scope, and `"\0Scope\0name"` **only** for a private
property inherited from another class. `"\0*\0name"` is never a valid input either.

### Why it looks intermittent

`DeepClone.php` has two paths:

```php
if (!$needsGroup) {
    // Fast path: every key resolves to $class. Hand $vars straight
    // to the $class hydrator — no grouping, no re-keying.
```

The fast path performs **no demangling**, so it ends up doing `$object->{"\0Solo\0x"} = ...` →
fatal `Error` at `DeepClone.php:2083`. The full-parse path *does* demangle, which is why the bug
only shows when **no** foreign-scope key is present:

| Snapshot contents | Result |
|---|---|
| private + protected, all declared on the object's own class | ❌ `Error` |
| a private inherited from a parent class is also present | ✅ works (full-parse path) |
| public properties only | ✅ works (no mangled key) |

So a class with no inherited private property fails, while one that has such a property silently
works — which is likely why this slipped through CI.

## Suggested fix

Translate the cast's mangling to VarExporter's convention before hydrating. Verified on all the
cases above, including inherited privates and modified public properties:

```php
public static function hydrateFromSnapshot(object $object, array $snapshot): void
{
    $vars = [];

    foreach ($snapshot as $key => $value) {
        if (!\is_string($key) || '' === $key || "\0" !== $key[0]) {
            $vars[$key] = $value;
            continue;
        }

        $parts = \explode("\0", $key, 3);

        if (3 !== \count($parts)) {
            $vars[$key] = $value; // e.g. the "\0" SPL state key
            continue;
        }

        [, $scope, $name] = $parts;

        // plain name for the object's own scope (and for protected),
        // mangled only for privates inherited from another class
        $vars['*' === $scope || $scope === $object::class ? $name : "\0{$scope}\0{$name}"] = $value;
    }

    // ... then hydrate with $vars instead of $snapshot
}
```

An alternative would be to capture the snapshot per declaring class via reflection and use
`VarExporter\Hydrator::hydrate($object, [], $scopedVars)`, whose `$scopedVars` form is the
documented API and handles the own-scope/foreign-scope distinction itself.

## Side note

Arguably `symfony/polyfill-deepclone` is inconsistent here — its full-parse path accepts
serialization-mangled keys while its fast path does not. But since `VarExporter\Hydrator` documents
and builds the plain-name-for-own-scope convention, the mismatch looks like it belongs on the
Foundry side. Happy to open an issue there instead if you prefer.

kabylixx:
Fixes #1148.

`hydrateFromSnapshot()` feeds an `(array) $object` cast straight to `deepclone_hydrate()`, but when every mangled key (`"\0Class\0name"`, `"\0*\0name"`) resolves to the object's own scope, the hydrator took its un-grouped fast path and used the mangled name verbatim, failing with `Error: Cannot access property starting with "\0"`. In practice this hits any autorefreshed object that gets deleted: `autorefresh()` then takes the "object no longer exists" branch and restores the snapshot.

The root cause was in the polyfill, and is fixed upstream in symfony/polyfill#643 (released in v1.42.0), so this PR:

- requires `symfony/polyfill-deepclone: ^1.42` — the minimal version carrying the fix;
- always hydrates snapshots with `deepclone_hydrate()`: the polyfill guarantees the function exists, so the deprecated `VarExporter\Hydrator` fallback is dropped;
- keeps the tests: 2 of the 4 unit cases fail with the pre-1.42 polyfill (own private — on an anonymous class —, own protected), the other 2 guard the paths that already worked (public-only, inherited properties including a private shadowed across scopes), plus a functional autorefresh-after-delete test on ORM and Mongo.

The Behat CI also needed a change: it installed the released `zenstruck/foundry` from Packagist and symlinked the checkout's `src/` over it, so a new dependency of the main package was never installed. It now installs the checkout itself as a composer path repository, resolving foundry's dependencies transitively from the branch's `composer.json` (composer refuses to install a path package inside its own source tree, hence the copy of the checkout outside of it).

https://claude.ai/code/session_01YKn786eyMPtirge4Jejumm

### DIFF
--- a/src/Object/Hydrator.php
+++ b/src/Object/Hydrator.php
@@ -15,7 +15,6 @@ use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException;
 use Symfony\Component\PropertyAccess\PropertyAccessor;
-use Symfony\Component\VarExporter\Hydrator as VarExporterHydrator;
 use Zenstruck\Foundry\Factory;
 use Zenstruck\Foundry\ForceValue;
 
@@ -213,13 +212,9 @@ final class Hydrator
      */
     public static function hydrateFromSnapshot(object $object, array $snapshot): void
     {
-        if (\function_exists('deepclone_hydrate')) {
-            // VarExporter's Hydrator is deprecated since symfony/var-exporter 8.1
-            // the constant is resolved dynamically: it only exists along with the function
-            deepclone_hydrate($object, $snapshot, \constant('DEEPCLONE_HYDRATE_PRESERVE_REFS'));
-        } else {
-            VarExporterHydrator::hydrate($object, $snapshot);
-        }
+        // symfony/polyfill-deepclone >= 1.42 is required: it fixes the hydration
+        // of mangled keys pointing to the object's own scope (symfony/polyfill#643)
+        deepclone_hydrate($object, $snapshot, \DEEPCLONE_HYDRATE_PRESERVE_REFS);
     }
 
     private static function accessibleProperty(object $object, string $name): \ReflectionProperty
--- a/src/Object/Hydrator.php
+++ b/src/Object/Hydrator.php
@@ -213,8 +213,6 @@ final class Hydrator
      */
     public static function hydrateFromSnapshot(object $object, array $snapshot): void
     {
-        $snapshot = self::normalizeMangledKeys($object, $snapshot);
-
         if (\function_exists('deepclone_hydrate')) {
             // VarExporter's Hydrator is deprecated since symfony/var-exporter 8.1
             // the constant is resolved dynamically: it only exists along with the function
@@ -222,48 +220,6 @@ final class Hydrator
         } else {
             VarExporterHydrator::hydrate($object, $snapshot);
         }
-    }
-
-    /**
-     * An `(array) $object` cast mangles non-public property names ("\0Class\0name" for a private
-     * one, "\0*\0name" for a protected one), but the hydrators expect the convention built by
-     * VarExporter\Hydrator: a plain name for anything writable in the object's own scope, and
-     * "\0Scope\0name" only for a private property declared by another class.
-     *
-     * Without this translation, a snapshot whose keys all resolve to the object's own scope takes
-     * the hydrator's un-grouped fast path and the mangled name is used verbatim, which fails with
-     * `Error: Cannot access property starting with "\0"`.
-     *
-     * @param array<string, mixed> $snapshot
-     *
-     * @return array<string, mixed>
-     */
-    private static function normalizeMangledKeys(object $object, array $snapshot): array
-    {
-        $vars = [];
-
-        foreach ($snapshot as $key => $value) {
-            $parts = \explode("\0", $key);
-
-            // a mangled key is "\0", the scope, "\0" and the property name, so it explodes into at
-            // least 3 parts. Anything shorter is left untouched: "\0" alone carries the state of
-            // SplObjectStorage/ArrayObject/ArrayIterator.
-            if (\count($parts) < 3 || '' !== $parts[0]) {
-                $vars[$key] = $value;
-
-                continue;
-            }
-
-            // the property name never contains a "\0" but an anonymous class name does, so the name
-            // is the last part and everything in between belongs to the scope
-            $name = \array_pop($parts);
-            \array_shift($parts);
-            $scope = \implode("\0", $parts);
-
-            $vars['*' === $scope || $scope === $object::class ? $name : "\0{$scope}\0{$name}"] = $value;
-        }
-
-        return $vars;
     }
 
     private static function accessibleProperty(object $object, string $name): \ReflectionProperty
--- a/src/Object/Hydrator.php
+++ b/src/Object/Hydrator.php
@@ -213,6 +213,8 @@ final class Hydrator
      */
     public static function hydrateFromSnapshot(object $object, array $snapshot): void
     {
+        $snapshot = self::normalizeMangledKeys($object, $snapshot);
+
         if (\function_exists('deepclone_hydrate')) {
             // VarExporter's Hydrator is deprecated since symfony/var-exporter 8.1
             // the constant is resolved dynamically: it only exists along with the function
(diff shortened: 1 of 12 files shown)


src/Object/Hydrator.php
```diff
@@ -213,6 +213,8 @@ public static function hydrateFromOtherObject(object $object, object $other): vo
      */
     public static function hydrateFromSnapshot(object $object, array $snapshot): void
     {
+        $snapshot = self::normalizeMangledKeys($object, $snapshot);
+
         if (\function_exists('deepclone_hydrate')) {
```
nikophil, line 216: I'm wondering if we should not require `symfony/polyfill-deepclone` and always use `deepclone_hydrate()` and use `deepclone_to_array()` instead of casting to array in `\Zenstruck\Foundry\Persistence\PersistedObjectsTracker::resetObjectAsLazyGhost()`, 

does it remove the problem of "mangled keys"? since we're using the same tool to export and clone, this should be OK?

WDYT?

tests/Unit/Object/HydratorTest.php
```diff
@@ -222,4 +223,101 @@ public function getFoo(): string
 
         $this->assertSame('foo', $object->getFoo());
     }
+
+    /**
+     * @test
+     */
+    #[Test]
+    public function can_hydrate_from_snapshot_with_public_property_only(): void
```
nikophil, line 231: any chance that you also reproduce the problem in a functional test please? in `\Zenstruck\Foundry\Tests\Integration\Persistence\AutoRefreshTestCase`

nikophil (commented):
thanks for the fix, I'm just wondering if there is not a solution out of the box using deepclone (and its polyfill) directly

nikophil:
hey @kabylixx I think the problem is in symfony/polyfill-deepclone

I'll propose a fix there

src/Object/Hydrator.php
nikophil: Hmmm sorry I was wrong here, this would not resolve anything `deepclone_to_array()` is not made for that

kabylixx:
Makes sense — and I agree the polyfill is inconsistent with itself here: its grouping path resolves `"\0Class\0name"` correctly, only the un-grouped fast path uses the key verbatim.

Two things that might matter for a fix landing there:

- foundry doesn't require the polyfill directly, it comes through `symfony/var-exporter`, which allows `^1.40` — so a fix released as 1.41 wouldn't reach users who already have 1.40 resolved. Would you add a direct requirement, or a `conflict` on the affected versions?
- `bootstrap81.php` returns early when `ext-deepclone` is loaded, so on PHP 8.5 with the extension the native implementation runs instead of the polyfill. Do you know whether it has the same fast-path behaviour? A polyfill-only fix wouldn't cover that case.

Happy to reduce this PR to just the tests if you'd rather fix it upstream — they fail against the current polyfill and would guard against a regression wherever the fix ends up living. Just say the word.

nikophil:
Hi!

what I'm planning to do is:
- propose a fix on `symfony/polyfill-deepclone` (https://github.com/symfony/polyfill/pull/643)
- create a fix release in Foundry with:
  - require `symfony/polyfill-deepclone` with the minimal version vith the fix
  - change the code in `\Zenstruck\Foundry\Object\Hydrator::hydrateFromSnapshot()` to always use `deepclone_hydrate()` (apparently, the problem does not exist in `ext/deepclone` - says Claude :robot: )
  - keep the tests you added in this PR

does that seem OK to you? I'd like to prevent monkey patching deepclone in Foundry :sweat_smile:

kabylixx:
Sounds good to me — and I'd rather not monkey-patch deepclone from Foundry either.

I've dropped `normalizeMangledKeys()` from this PR and kept only the tests, so it's ready to merge once the polyfill is tagged.

nikophil:
hey @kabylixx tests are now passing, I'll merge/release this soon

nikophil:
fix released in https://github.com/zenstruck/foundry/releases/tag/v2.12.1

thanks for your help!

PR #1159, opened 2026-08-25
docs: disambiguate factory "attributes" from native PHP attributes

Amoifr:
Fixes #1125

Foundry's "attributes" (the property-value arrays passed to `defaults()`, `createOne()`, ...) now collide with native PHP `#[...]` attributes in readers' minds. Both improvements suggested in the issue, with @garak's blessing:

- a terminology note at the top of the docs' "Attributes" section;
- the docblocks of the methods and functions taking `$attributes` now say what the array represents.

No API or behavior change.

PR #1161, opened 2026-08-26
feat: add Factory::memoize() and FactoryCollection::memoize()

### LINKED ISSUE #1139: Introduce `Factory::memoize()` and `FactoryCollection::memoize()`
so we can do:
```php
class TaskFactory extends PersistentObjectFactory
{
    // ...

    {
        $owner = UserFactory::new()->memoize();

        return [
            'project' => ProjectFactory::new(['users' => [$owner]]),
            'owner'   => $owner,
        ];
    }
}
```

`memoize()` helper is still needed to memoize things that are not related to entity/factory, so we will keep it.

Maybe add a deprecation if a factory/factory collection is returned by the callback, because it is error prone:
```php
// will emit a deprecation: "passing a factory to memoize() is deprecated, please use Factory::memoize() instead, or actually create the object in the callback"
memoize(fn() => UserFactory::new())
```

Amoifr:
Fixes #1139.

`memoize()` takes a callback, so sharing an object between attributes means remembering to create it inside that callback. Passing the factory instead is the natural thing to write, and it silently does the wrong thing: the LazyValue memoizes the factory, and `normalizeParameter()` calls `create()` on it at every use, so each attribute ends up with its own object.

`Factory::memoize()` and `FactoryCollection::memoize()` remove both the callback and the trap, and the old spelling is deprecated rather than left to surprise people.

Built on the `LazyValue::memoizeFromFactory()` named constructor you suggested. It reads well to me so I did not go looking for another design, but I have no attachment to it.

### DIFF
--- a/src/LazyValue.php
+++ b/src/LazyValue.php
@@ -23,7 +23,7 @@ final class LazyValue
     /**
      * @param callable():mixed $factory
      */
-    private function __construct(callable $factory, private bool $memoize = false)
+    private function __construct(callable $factory, private bool $memoize = false, private bool $fromFactory = false)
     {
         $this->factory = $factory(...);
     }
@@ -41,6 +41,14 @@ final class LazyValue
 
         if ($value instanceof self) {
             $value = ($value)();
+        }
+
+        if ($this->fromFactory) {
+            // the factory is created here so that the object it produces is what gets memoized:
+            // memoizing the factory itself would create a new object on each use of the value
+            $value = $value->create();
+        } elseif ($this->memoize && ($value instanceof Factory || $value instanceof FactoryCollection)) {
+            trigger_deprecation('zenstruck/foundry', '2.13', 'Passing a factory to memoize() is deprecated: it memoizes the factory and not the object it creates, so a new object is created on each use. Use Factory::memoize() instead, or create the object in the callback.');
         }
 
         if (\is_array($value)) {
@@ -67,7 +75,17 @@ final class LazyValue
      */
     public static function memoize(callable $factory): self
     {
-        return new self($factory, true);
+        return new self($factory, memoize: true);
+    }
+
+    /**
+     * @internal
+     *
+     * @param callable():(Factory<mixed>|FactoryCollection<mixed, Factory<mixed>>) $factory
+     */
+    public static function memoizeFromFactory(callable $factory): self
+    {
+        return new self($factory, memoize: true, fromFactory: true);
     }
 
     /**
--- a/src/LazyValue.php
+++ b/src/LazyValue.php
@@ -43,12 +43,11 @@ final class LazyValue
             $value = ($value)();
         }
 
-        if ($this->fromFactory) {
-            // the factory is created here so that the object it produces is what gets memoized:
-            // memoizing the factory itself would create a new object on each use of the value
-            $value = $value->create();
-        } elseif ($this->memoize && ($value instanceof Factory || $value instanceof FactoryCollection)) {
-            trigger_deprecation('zenstruck/foundry', '2.13', 'Passing a factory to memoize() is deprecated: it memoizes the factory and not the object it creates, so a new object is created on each use. Use Factory::memoize() instead, or create the object in the callback.');
+        // memoizeFromFactory() creates the object in its own callback, so what is memoized here is
+        // the object; a user callback returning a factory memoizes the factory instead, which
+        // creates a new object on each use of the value
+        if (!$this->fromFactory && $this->memoize && ($value instanceof Factory || $value instanceof FactoryCollection)) {
+            trigger_deprecation('zenstruck/foundry', '2.13', 'Passing a factory to memoize() is deprecated and will throw an error in Foundry 3: it memoizes the factory and not the object it creates, so a new object is created on each use. Use Factory::new()->memoize() instead, or create the object in the callback.');
         }
 
         if (\is_array($value)) {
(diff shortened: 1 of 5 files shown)


src/LazyValue.php
```diff
@@ -67,7 +75,17 @@ public static function new(callable $factory): self
...
+    }
+
+    /**
+     * @internal
+     *
+     * @param callable():(Factory<mixed>|FactoryCollection<mixed, Factory<mixed>>) $factory
+     */
+    public static function memoizeFromFactory(callable $factory): self
+    {
+        return new self($factory, memoize: true, fromFactory: true);
```
nikophil: I think we can do better:

```suggestion
    public static function memoizeFromFactory(Factory|FactoryCollection $factory): self
    {
        return new self(static fn() => $factory->create(), memoize: true, fromFactory: true);
```

src/LazyValue.php
```diff
@@ -43,6 +43,14 @@ public function __invoke(): mixed
             $value = ($value)();
         }
 
+        if ($this->fromFactory) {
+            // the factory is created here so that the object it produces is what gets memoized:
+            // memoizing the factory itself would create a new object on each use of the value
+            $value = $value->create();
+        } elseif ($this->memoize && ($value instanceof Factory || $value instanceof FactoryCollection)) {
+            trigger_deprecation('zenstruck/foundry', '2.13', 'Passing a factory to memoize() is deprecated: it memoizes the factory and not the object it creates, so a new object is created on each use. Use Factory::memoize() instead, or create the object in the callback.');
```
nikophil:
```suggestion
            trigger_deprecation('zenstruck/foundry', '2.13', 'Passing a factory to memoize() is deprecated and will throw an error in Foundry 3: it memoizes the factory and not the object it creates, so a new object is created on each use. Use Factory::new()->memoize() instead, or create the object in the callback.');
```

tests/Unit/LazyValueTest.php
```diff
@@ -75,4 +81,49 @@ public function can_handle_array_with_lazy_values(): void
...
+    /**
+     * @test
+     *
+     * @group legacy
+     */
+    #[Test]
+    #[IgnoreDeprecations]
+    public function memoizing_a_factory_creates_a_new_object_on_each_use_and_is_deprecated(): void
+    {
+        $this->expectUserDeprecationMessageMatches('/Passing a factory to memoize\(\) is deprecated/');
```
nikophil, line 126: Foundry is tested with PHPUnit 9 to 13, and this method does not exist in PHPUnit 9

I think you can just ignore this test in PHPUnit 9: add `#[RequiresPhpunit('>=10.0.0')]`

nikophil (changes requested):

src/LazyValue.php
Amoifr: Applied. Since the closure now creates, `__invoke()` no longer has anything to create, so `$fromFactory` is down to one job: keeping the deprecation below from firing on this path, where the value is an object rather than a factory. Kept it for that, say the word if you would rather see the check written differently.

src/LazyValue.php
Amoifr: Applied verbatim.

tests/Unit/LazyValueTest.php
Amoifr: Good catch, thanks. I went with `>=11.0.0` rather than `>=10.0.0`: the run on this PR shows the method is missing on 10 too.

```
P:8.2, S:7.4.*, D:mysql, PU:10
Error: Call to undefined method LazyValueTest::expectUserDeprecationMessageMatches()
```

It was added in PHPUnit 11, which is also the bound `FakerAdapterTest` uses. Added the `@requires PHPUnit >=11.0.0` annotation next to the attribute, like the other tests do, so PHPUnit 9 honours it as well.

Amoifr:
The remaining red `Code Coverage` is a dependency mismatch in that job rather than something in this PR. The suite itself is green there (2053 tests, no failures); it exits non-zero on two deprecations:

`symfony/framework-bundle` **v8.0.15** calls `RegisterListenersPass::setHotPathEvents()` (`FrameworkBundle.php:143`), which `symfony/event-dispatcher` **v8.1.5** deprecates. The job locked that pair despite `SYMFONY_REQUIRE: 8.1.x`, while the last green run on `2.x` had framework-bundle v8.1.5 with the same event-dispatcher.

They are triggered by `LoadFixturesCommandTest`, which this PR does not touch, and an untouched `2.x` reproduces them locally with the same dependency mix. A re-run should sort it out.

Amoifr:
@nikophil whenever you have a moment: your three points landed in 97a97ea on August 27th, and each thread has a reply.

The one check still red is `Code Coverage`, and it does not come from this PR: 2053 tests, 0 failures, the non-zero exit comes from two `symfony/event-dispatcher 8.1` deprecations (`setHotPathEvents()` and `setNoPreloadEvents()`). The same job fails on `2.x` itself, where the count was up to four on September 3rd, so it is quietly growing on its own.

The PR is only waiting on a re-review. Thanks again for the quick first pass!

nikophil:
hi @Amoifr sorry for the delay, I'll review all your PRs today, thanks for your help!

nikophil:
BTW you can rebase your PR, the deprecation is gone

Amoifr:
Rebased, thanks. I did the same for #1159 and #1164 since you are going through them today, so all three are on current `2.x` and the Code Coverage job should be green now that the upstream deprecation is gone.

No rush on the review, and thanks for taking the time.

nikophil (approved):

nikophil:
thanks @Amoifr

PR #1163, opened 2026-08-31
Document valid string values for orm.reset.mode

### LINKED ISSUE #1162: [BUG]
| Foundry's version?                                             | 2.11.3                                     |
| PHP version                                                         | 8.4             |
| PHPUnit version                                                 | 13                    |
| `symfony/framework-bundle` version?          | 8.1.4                      |
| Do you use dama/doctrine-test-bundle?       | no                               |
| Which reset database strategy do you use? | schema / migrate / none |
| Do you use Foundry's PHPUnit extension?   | yes / no                               |

Trying new `symfony lsp:check` results in this error:

```
.:config/packages/zenstruck_foundry.yaml:9:23: warning [config.invalid_type] Expected enum for "zenstruck_foundry.orm.reset.mode".
```

my config:

```yaml
when@dev: &dev
    zenstruck_foundry:
        instantiator:
            use_constructor:  true
        orm:
            reset:
                connections: [ default ]
                entity_managers: [ default ]
                mode: schema
        persistence:
            # Flush only once per call of `PersistentObjectFactory::create()`
            flush_once: true

    # If you use the `make:factory --test` command, you may need to uncomment the following.
    # See https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html#generate
    #services:
    #    App\Tests\Factory\:
    #        resource: '%kernel.project_dir%/tests/Factory/'
    #        autowire: true
    #        autoconfigure: true
when@test: *dev
```

it doesn't like `mode: schema` but this is your documented value.

derrabus:
Fixes #1162

The configuration key `orm.reset.mode` accepts a case of the `ResetDatabaseMode`, but also one of its backing string values. The value is normalized here:

https://github.com/zenstruck/foundry/blob/65b9f6029a7d82bdcc59aeeb784c6630356d5591/src/ZenstruckFoundryBundle.php#L155

However, we currently tell the Config component that only the actual enum cases are valid which causes tooling like PHPStan or the new `symfony lsp:check` command to flag the backing string values as invalid.

This PR fixes this issue by merging the enum cases and their backing values into a single array of valid values.

### DIFF
--- a/src/ZenstruckFoundryBundle.php
+++ b/src/ZenstruckFoundryBundle.php
@@ -154,7 +154,10 @@ final class ZenstruckFoundryBundle extends AbstractBundle implements CompilerPas
                                         ->ifString()
                                         ->then(static fn(string $mode): ?ResetDatabaseMode => ResetDatabaseMode::tryFrom($mode))
                                     ->end()
-                                    ->values(ResetDatabaseMode::cases())
+                                    ->values([
+                                        ...ResetDatabaseMode::cases(),
+                                        ...array_column(ResetDatabaseMode::cases(), 'value'),
+                                    ])
                                 ->end()
                                 ->arrayNode('migrations')
                                     ->addDefaultsIfNotSet()


nikophil (approved):

nikophil:
Thanks @derrabus

Issue #883, opened 2025-04-20
Insert a lot of data - batch processing


seb-jean:
Hi,

I'd like to insert a lot of data into my `User` table, which contains many columns, but it's taking a long time. So I used another entity called `Message`, which has a few columns. Unfortunately, importing 1 million rows is also taking a long time. It's been going on for over an hour.
I'm wondering if it's possible to insert a lot of data with Foundy.
Would a [batch processing](https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/batch-processing.html#bulk-inserts) solution offered by Doctrine be worth implementing in Foundry?
Thanks :).

```php
<?php

namespace App\DataFixtures;

use App\Entity\User;
use App\Factory\MessageFactory;
use Doctrine\Bundle\FixturesBundle\Fixture;
use Doctrine\Persistence\ObjectManager;
use function Zenstruck\Foundry\Persistence\repository;

use function Zenstruck\Foundry\Persistence\flush_after;

final class AppFixtures extends Fixture
{
    public function load(ObjectManager $manager): void
    {
        flush_after(function () {
            $user1 = repository(User::class)->find(805);
            $user2 = repository(User::class)->find(804);

            MessageFactory::createMany(1000000, attributes: fn ($i) => [
                'sender' => $user1,
                'receiver' => $user2,
            ]);
        });
    }
}
```

nikophil:
Hi!

yes, this could actually be possible I think. 

Maybe we should introduce a new method that don't return anything, so that all the objects won't be loaded in memory

Then we should transform `FactoryCollection::all()` so that it returns a generator. But it would a BC break, so we have to find a solution for this, maybe introduce a new private method.

Finally, the batch handling should take place in the `PersistenceManager` 

@kbond any thoughts?

kbond:
Yeah, not a lot of effort has gone into making Foundry performant. That being said, we have been improving it lately. Because of the `LazyObjectManager`, clearing the EM shouldn't be an issue. I don't know that time would be improved much but certainly memory usage.

nikophil:
I think that currently there is no way to do this, because of memory exploding, but yeah, Foundry would not be the best to create tons of data. 

On the other hand, I don't think this would be to hard to do, so I'm not opposed to it

seb-jean:
What would be the best library for creating tons of data?

nikophil:
don't know...

maybe custom stuf? depends on your needs.

[...]

PR #1170, opened 2026-09-19
feat: add bulk insert API for mass data loading

seb-jean:
I've been experimenting for a few days with adding the ability to insert 2,000,000 users into a table, along with their respective relations. enum fields, password, etc.


nikophil:
hi @seb-jean 

please tell me more about what you want to achieve: what's the purpose to insert millions of items with Foundry? do you need this feature to seed a dev DB or are you using it inside your tests? using `foundry:load-fixtures` command?

Are you able to use `--no-debug` flag? In your opinion, what is an acceptable duration to create 1M lines? would you need to still handle relationships? Foundry's hooks?

nikophil:
BTW, I looked a little bit to you experimentation, there is almost no chance that this kind of thing would be merged in Foundry: this is too much complexity, coupled with PgSQL...

I have tried some stuff, which reuses almost 100% of the features Foundry provides using batch processing and the result is still kinda slow, but memory is stable and speed is linear, it would produce 1M `Message` like in [your first example](https://github.com/zenstruck/foundry/issues/883) in 19 minutes. It also produces 1M `Contact` objects in Foundry's fixtures (see https://github.com/zenstruck/foundry/blob/2.x/tests/Fixture/Factories/Entity/Contact/ContactFactory.php) in 29 minutes. I think this is still better than nothing, but not sure that's the kind of performances you're looking for

seb-jean:
hi @nikophil ,

I want to insert millions of items using Foundry into a development database to verify that the code handles such large datasets correctly.

For instance, I need millions of records to test features like cursor-based pagination or PostGIS functionality.

I am using the `symfony console foundry:load-fixtures main` command. I tried running it with `--no-debug`, but it didn't speed things up.

An acceptable timeframe would be 30 minutes to create 1 million rows.

Relationships and hooks also need to be handled; specifically, I am using `afterPersist` and `afterInstantiate`.

nikophil:
@seb-jean ok nice, this is somehow the way I was understanding your need

I'll soon provide a draft PR so that you can test

PR #951, opened 2025-07-27
docs: create upgrade guide to 2.7 and document auto-refresh

nikophil:

PR #1171, opened 2026-09-21
Fix the namespace of the persistence helper functions in the docs

Amoifr:
The "Persistence related helpers" list spells the namespace as `Zenstruck\Foundry\Persistent\`, while the functions live in `Zenstruck\Foundry\Persistence\`. Copying a line straight out of the docs gives a fatal error:

```php
use function Zenstruck\Foundry\Persistent\save; // Error: Call to undefined function
```

All twelve entries of the list are affected, and all twelve functions exist under the corrected namespace. The code blocks elsewhere in the page already use `Persistence`, so only that list is out of step.

It came in with the list itself in #951, which is why `2.3.x` is untouched and `3.x` carries it too.

DOCtor-RST stays green on the page.


nikophil:
thanks @Amoifr

